PR #9, opened 2018-10-02
Fix numerical bugs in PD and Simplex distributions link, invlink and tests

### LINKED ISSUE #7: Dirichlet transform is not numerically stable for high dimensions
Running the following code 

```julia
using Turing

@model mdemo(d, N) = begin
    Θ = Vector(undef, N)
   for n=1:N
      Θ[n] ~ d
   end
end

# This leads to many numerical issues in gradients
dim  = 5 # This works fine
dim2 = 50 # This breaks
d = Dirichlet(ones(dim2) / dim2)

Turing.setadbackend(:forward_diff)
chain = sample(mdemo(d, 1), HMC(5000, 0.1, 5)) 

Turing.setadbackend(:reverse_diff)
chain = sample(mdemo(d, 1), HMC(5000, 0.1, 5)) 
```

produces

```julia
julia> chain = sample(mdemo(d, 1), HMC(5000, 0.1, 5)) 
[ Info:  Assume - `Θ` is a parameter
[HMC] Sampling...  0%  ETA: 3:23:51
  ϵ:         0.1
  α:         0.9992054011165935
  pre_cond:  [1.0, 1.0, 1.0, 1.0, 1.0, 1.0,...┌ Warning: Numerical error has been found in gradients.
└ @ Turing ~/.julia/dev/Turing/src/core/ad.jl:114
┌ Warning: grad = Real[-Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, NaN]
└ @ Turing ~/.julia/dev/Turing/src/core/ad.jl:115
┌ Warning: Numerical error has been found in gradients.
└ @ Turing ~/.julia/dev/Turing/src/core/ad.jl:114
┌ Warning: grad = Real[-Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, NaN]
└ @ Turing ~/.julia/dev/Turing/src/core/ad.jl:115
┌ Warning: Numerical error has been found in gradients.
└ @ Turing ~/.julia/dev/Turing/src/core/ad.jl:114
┌ Warning: grad = Real[-Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, -Inf, NaN]
└ @ Turing ~/.julia/dev/Turing/src/core/ad.jl:115
```

mohamed82008:
Closes #7. This PR fixes a few issues with the `link` and `invlink` functions as well as the tests for simplex and positive definite distributions. Most notably, it makes sure that `logit` only receives numbers from `eps(T)` to `1 - eps(T)`, while imposing some correctness constraint, i.e. `z = 1` when `x[k] = 0` and `sum(x[1:k-1]) = 1`. The mathematical justification for this is given in #7. 

Another important point is the treatment of `y[end]` in `link` of the Simplex distribution. `y[end]` should be 0 if `sum(x) == 1`, but there is no need for the `link` function to impose that constraint. `y[end] = 0` is implied when `x` is in the support of the distribution. Imposing the constraint in the function makes the function's Jacobian rank deficient since `y[end]` is now a constant. So to make the Jacobian full rank, the value of `y[end]` was computed from `x`. Conversely, the value of `x[end]` was computed from `y`.

The last important change in this PR is in the Jacobian of the Cholesky factor. In the Cholesky decomposition algorithm, the matrix is symmetric. The top triangle of numbers are used to compute the Cholesky factor which is written in the lower triangle of the output matrix. The strict top triangle of the Cholesky factor is all zeros, constant. And the strict bottom triangle of the input matrix does not contribute to the Cholesky factor since the matrix is symmetric so they get ignored. This means that the effective Jacobian of the Cholesky factor is `J[lowerinds, upperinds]`, whose `(i,j)` element gives the partial derivative of the `i`th lower triangular element of the Cholesky factor with respect to the `j`th upper triangular element of the input matrix. This change was made in the test for the Jacobian of the Wishart distribution.

All turing tests pass except 3 broken tests which probably need the same treatment as the Cholesky's Jacobian above, didn't check them in details though. I will go back to TuringBenchmarks now, sorry for digressing!

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -143,7 +143,11 @@ end
 
 const SimplexDistribution = Union{Dirichlet}
 
-function link(d::SimplexDistribution, x::AbstractVector{T}, ::Type{Val{proj}} = Val{true}) where {T<:Real, proj}
+function link(
+    d::SimplexDistribution, 
+    x::AbstractVector{T}, 
+    ::Type{Val{proj}} = Val{true}
+) where {T<:Real, proj}
     y, K = similar(x), length(x)
 
     ϵ = eps(T)
@@ -168,7 +172,11 @@ function link(d::SimplexDistribution, x::AbstractVector{T}, ::Type{Val{proj}} = 
 end
 
 # Vectorised implementation of the above.
-function link(d::SimplexDistribution, X::AbstractMatrix{T}, ::Type{Val{proj}} = Val{true}) where {T<:Real, proj}
+function link(
+    d::SimplexDistribution, 
+    X::AbstractMatrix{T}, 
+    ::Type{Val{proj}} = Val{true}
+) where {T<:Real, proj}
     Y, K, N = similar(X), size(X, 1), size(X, 2)
 
     ϵ = eps(T)
@@ -192,7 +200,11 @@ function link(d::SimplexDistribution, X::AbstractMatrix{T}, ::Type{Val{proj}} = 
     return Y
 end
 
-function invlink(d::SimplexDistribution, y::AbstractVector{T}, ::Type{Val{proj}} = Val{true}) where {T<:Real, proj}
+function invlink(
+    d::SimplexDistribution, 
+    y::AbstractVector{T}, 
+    ::Type{Val{proj}} = Val{true}
+) where {T<:Real, proj}
     x, K = similar(y), length(y)
 
     ϵ = eps(T)
@@ -214,7 +226,11 @@ function invlink(d::SimplexDistribution, y::AbstractVector{T}, ::Type{Val{proj}}
 end
 
 # Vectorised implementation of the above.
-function invlink(d::SimplexDistribution, Y::AbstractMatrix{T}, ::Type{Val{proj}} = Val{true}) where {T<:Real, proj}
+function invlink(
+    d::SimplexDistribution, 
+    Y::AbstractMatrix{T}, 
+    ::Type{Val{proj}} = Val{true}
+) where {T<:Real, proj}
     X, K, N = similar(Y), size(Y, 1), size(Y, 2)
 
     ϵ = eps(T)
@@ -304,7 +320,11 @@ function invlink(d::PDMatDistribution, Y::AbstractMatrix{T}) where {T<:Real}
     return LowerTriangular(X) * LowerTriangular(X)'
 end
 
-function logpdf_with_trans(d::PDMatDistribution, X::AbstractMatrix{<:Real}, transform::Bool)
+function logpdf_with_trans(
+    d::PDMatDistribution, 
+    X::AbstractMatrix{<:Real}, 
+    transform::Bool
+)
     lp = logpdf(d, X)
     if transform && isfinite(lp)
         U = cholesky(X).U
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -49,8 +49,6 @@ export  TransformDistribution,
 # a ≦ x ≦ b #
 #############
 
-proj_invlink(d, y) = invlink(d, y)
-
 const TransformDistribution{T<:ContinuousUnivariateDistribution} = Union{T, Truncated{T}}
 
 function link(d::TransformDistribution, x::Real)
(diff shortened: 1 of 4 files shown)


yebai:
I'm happy with the PR. Ready to merge once @willtebbutt is happy!

mohamed82008:
These new tests break. I will need to look further into them, maybe tomorrow.

mohamed82008:
I believe this is ready for a second review @willtebbutt

mohamed82008:
Please don't merge this, it seems to be causing problems in TuringBenchmaks.

willtebbutt:
@mohamed82008 any movement on this? Does it still cause issues for TuringBenchmarks.jl?

mohamed82008:
TuringBenchmarks is still in a little of bit of flux as I am trying to replace mlab with an in-house approach for saving benchmark results to the repo. When it settles down a little, I will give this a second shot.

mohamed82008:
I reduced the error causing code to:
```julia
using Turing

@model nbmodel(K, alpha, beta) = begin
  theta ~ Dirichlet(alpha)
  phi = TArray{Any}(K)
  for k = 1:K
    phi[k] ~ Dirichlet(beta)
  end
  phi
end

K = 4
alpha = [1, 1, 1, 1]
beta = [0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1, 0.1]
chn = sample(nbmodel(K, alpha, beta), HMC(100, 0.01, 5))
```

mohamed82008:
Further reduced to:
```julia
using Bijectors

dist = Dirichlet{Float64}([1.0, 1.0, 1.0, 1.0])
invlink(dist, [1.84197, 0.0897041, 2.09031, 1.74824])

#=
4-element Array{Float64,1}:
  0.6777296584697047 
  0.1139407310887609 
  0.18540455939051306
 -1.7253149489489785 
=#
```
giving a negative number.

mohamed82008:
Ok so the reason for this failure is very interesting. One change made in this PR is to make sure `link` and `invlink` for `Dirichlet` don't enforce the constraint of `x` summing to 1. This leads to a full rank Jacobian when using autodiff. Of course, because changes  in `x` must sum up to 0, the changes in `y` will also happen in a reduced subspace of dimension `K - 1`, where `K = length(x)`. Conversely, this also means that not every `y` can be mapped back to a valid `x`, the `y` has to be in the valid reduced subspace. But since, the `Turing.initdist` in https://github.com/TuringLang/Turing.jl/blob/5caa97a733cac885aa8959c3a692ed784df6cbec/src/utilities/robustinit.jl#L16 falls back to the `invlink` of a `y` sampled from a uniform distribution, this `y` is most likely going to be outside the valid subspace. So mapping it back to `x` using the blind `invlink` will give an invalid `x`. This causes the above negative number which is an extremely invalid case, but it could have also just been an `x` where `sum(x) != 1`. 

The fix for this would be to have a `projected_invlink` which projects the output `x` on the subspace satisfying `sum(x) == 1` for `Dirichlet` and falls back on `invlink` for other distributions, and then have `initdist` call `projected_invlink` instead of `inv_link`. If you approve, I will go ahead and try that fix.

mohamed82008:
Or maybe it is more correct to make `invlink` projecting by default and live with the fact that the Jacobian will be rank deficient?

mohamed82008:
Ok so I went with solution 1 just to see if more errors will show up, and the above HMC example errors with another Flux related error:
```julia
ERROR: MethodError: no method matching eps(::Type{Flux.Tracker.TrackedReal{Float64}})
```

mohamed82008:
Ok, so it seems that it is a deeply embedded assumption that `invlink` is supposed to be projecting by default, will fix that.

mohamed82008:
@willtebbutt this is ready for another review.

src/Bijectors.jl
```diff
@@ -141,71 +143,95 @@ end
 
 const SimplexDistribution = Union{Dirichlet}
 
-function link(d::SimplexDistribution, x::AbstractVector{T}) where T<:Real
+function link(d::SimplexDistribution, x::AbstractVector{T}, ::Type{Val{proj}} = Val{true}) where {T<:Real, proj}
```
willtebbutt: There are a few locations where the function headers exceed 92 characters. Please resolve this by splitting them up in accordance with the Turing style guide

mohamed82008: Sure. I will fix this and merge.

Issue #12, opened 2019-01-03
Numerical issue with random walk for Dirichlet


xukai92:
I've noticed a numeric issue when using this package with random walk algorithm. The code below is the minimal example to produce the bug:

```julia
using Distributions, Bijectors
dim = 10
dist = Dirichlet(dim, 2.0)
stepsize = 1e10
# Below gives DomainError with high probability
[logpdf_with_trans(dist, invlink(dist, link(dist, rand(dist)) .+ randn(dim) .* stepsize), true) for _ in 1:1_000]
```

I don't know if a stepsize of 10 is too high or not but it seems to be possible to have this kind of values during adaptation.

yebai:
`stepsize` of `10` is quite big actually, did you look at adapted `epsilon` in HMC?

xukai92:
Just checked - this happens when the step size is tuned to something like `5.75`.

yebai:
I think NUTS adaption needs some further improvement. Before that happens, perhaps we can heuristically upper bound step size, e.g. `< 2`?

xukai92:
> I think NUTS adaption needs some further improvement. Before that happens, perhaps we can heuristically upper bound step size, e.g. < 2?

2 still seems to be too high. 1 works.

> Dirichlet distribution is a very nasty one for transforms, I suspect it'll be hard to make it stable in all cases. If it occasionally throws off the edges, we can try to reject a Hamiltonian simulation.

Then we need to do a catch-exception which is slow. Alternatively we can make `logpdf_with_trans` return `-Inf` by checking terms inside `log` before evaluation?

yebai:
> Then we need to do a catch-exception which is slow. Alternatively, we can make logpdf_with_trans return -Inf by checking terms inside log before evaluation?

Sounds good to me.

mohamed82008:
We can just do `log(max(0, x))` everywhere there is `log(x)`. The output may then be `-Inf` or `NaN` if we divide 2 `Inf`s by each other for some reason. So outside we can check the output with `isnan` and `isinf`.

yebai:
This sounds good to me. Perhaps give it a try and see whether it can fix #621?

xukai92:
@yebai I will take a try today

xukai92:
@mohamed82008 @yebai great it works

PR #13, opened 2019-01-11
implement mohamed's solution (#12#issuecomment-451416440)

xukai92:
Resolving https://github.com/TuringLang/Bijectors.jl/issues/12

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -6,14 +6,14 @@ using StatsFuns
 using LinearAlgebra
 using MappedArrays
 
-export  TransformDistribution, 
+export  TransformDistribution,
         RealDistribution,
         PositiveDistribution,
         UnitDistribution,
         SimplexDistribution,
         PDMatDistribution,
-        link, 
-        invlink, 
+        link,
+        invlink,
         logpdf_with_trans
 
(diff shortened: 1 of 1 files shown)


xukai92:
My editor removed some empty spaces automatically. But I guess it's fine.

src/Bijectors.jl
```diff
@@ -276,11 +276,11 @@ function logpdf_with_trans(
 
         sum_tmp = zero(eltype(x))
         z = x[1]
-        lp += log(z + ϵ) + log(one(T) - z + ϵ)
+        lp += log(max(0, z + ϵ)) + log(max(0, one(T) - z + ϵ))
```
yebai: perhaps we can overload `log(x)=log(max(0, x))`? This would fix all uses of `log`'s, e.g. those in univariate distributions with lower / upper bounds.

mohamed82008: May be but that would give a stack overflow. `_log(x) = log(max(0, x))` would do, replacing all uses of `log` with `_log`.

xukai92: Or we'd name it something like `logsafe` so that it's more informative?

yebai:
Closed in favour of https://github.com/TuringLang/Bijectors.jl/pull/14

PR #15, opened 2019-01-19
Fix Float32 support

mohamed82008:
This is a bug fix. I noticed some test failure happening in Turing on 32 bit machines on AppVeyor with NUTS. I am not sure if this is the cause or not, but using `eps(Float64)` is definitely a bug here for 32 bit machines.


mohamed82008:
Never mind. I misunderstood how floating point calculations work. This is not needed.

PR #19, opened 2019-02-23
Low hanging performance

### LINKED ISSUE #18: `@debug` causing type instability and performance hit
It seems the `@debug` statements when expanded are causing a type instability and performance hit. ref: #17

mohamed82008:
Fix #18

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -16,8 +16,10 @@ export  TransformDistribution,
         invlink,
         logpdf_with_trans
 
-_eps(::Type{T}) where {T} = eps(T)
-_eps(::Type{Real}) = eps(Float64)
+const DEBUG = Bool(parse(Int, get(ENV, "DEBUG_BIJECTORS", "0")))
+
+@inline _eps(::Type{T}) where {T} = eps(T)
+@inline _eps(::Type{Real}) = eps(Float64)
 function __init__()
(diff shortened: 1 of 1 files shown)


willtebbutt:
This generally looks good, but I wonder whether inlining everything is a good idea? Do we actually see a significant performance gain?

mohamed82008:
Inlining medium sized functions which are called very often can trigger more compiler optimizations, especially when using views whose allocations can be eliminated by inlining the function. Small functions are usually inlined automatically by the compiler anyways. My goal here is to get to a point where the model function in Turing is as good as a function written by hand. This requires inlining all the `assume` and `observe` functions beside all the Bijectors functions used there. This will remove all the function call overhead but more importantly it should give the compiler more optimization options, e.g. eliminating intermediate variables. I will benchmark it properly tomorrow to be sure it actually helps. My intuition says this is more future proof though, since the smarter the compiler gets, the more it would be able to optimize the inlined functions simultaneously considering their interactions, not just individual function bodies. This will also play well with things like `StaticArray` data or parameters which work best when everything is inlined. `StaticArrays.jl` inlines the hell out of almost every function.

mohamed82008:
I ran a few experiments and couldn't prove any speedup to some of the inlines using the Dirichlet distribution, and they are a bit too many to test them all now. So I will remove them for now. It seems that the compiler's heuristic for inlining functions is hard to beat. This could be a nice benchmarking exercise some other time.

PR #22, opened 2019-03-03
Fix Turing issue 700

mohamed82008:
This PR fixes https://github.com/TuringLang/Turing.jl/issues/700 by working around https://github.com/JuliaDiff/ForwardDiff.jl/issues/396. This workaround is not necessary if the source of the bug is fixed upstream. If not, we can merge this.


yebai:
Thanks, @mohamed82008!

Issue #6, opened 2018-09-28
New interface


yebai:
`logpdf(t::Transform, l::Distribution, x)`

willtebbutt:
I was thinking something along the lines of
```julia
struct Transformed{D<:Distribution, T} <: Distribution
    d::D
    t::T
end
logpdf(::Transformed{D, T}, ::Real) = ...
rand(::Transformed{D, T}) = ...
```
(this would need to be specialised for Univariate, Multivariate etc)

torfjelde:
Hey! I made an attempt at implementing automatic differentiable variational inference (ADVI) for Turing, which requires transforming the distribution to a distribution with real support, i.e. can use `Bijectors.jl` for this!

While doing this I was thinking a bit about the interface and implemented the following as a "wrapper" for `Bijectors.jl`.

```julia
using Distributions, Bijectors
import Distributions: logpdf, rand

# Issue: this is not enforcing `VariateForm` of `D` and `Transformed{D}` to be the same
struct Transformed{D} <: Distribution{VariateForm, Continuous} where D <: Distribution
    d::D
end

# size
Base.size(td::Transformed) = size(td.d)
Base.length(td::Transformed) = length(td.d)

# logp
logpdf(d::Transformed{D} where D <: Distribution{Univariate}, x::T where T <: Real) = logpdf_with_trans(d.d, x, true)
logpdf(d::Transformed{D} where D <: Distribution{Multivariate}, x::AbstractMatrix{T} where T <: Real) = logpdf_with_trans(d.d, x, true)

# rand
rand(td::Transformed, n::Int64) = link(td.d, rand(td.d, n))
rand(d::Transformed{D} where D <: Distribution{Univariate}) = first(rand(d, 1))
rand(d::Transformed{D} where D <: Distribution{Multivariate}) = rand(d, 1)
```

**Important:** vectorized `logpdf` does not work, and results in the following error

```
ERROR: MethodError: no method matching iterate(::Transformed{MvNormal{Float64,PDMats.PDMat{Float64,Array{Float64,2}},Array{Float64,1}}})
```

If we could fix the above issue, maybe this would be a nice "ad-hoc" implementation. Comparing to the above suggestion which to me seems like it would require a more thorough rewrite of `Bijectors.jl` (though I might be thoroughly wrong!).

I haven't investigated this to a satisfactory degree, but I was encouraged on `#turing` to share anyways to initiate a discussion.

cpfiffer:
[...]

PR #27, opened 2019-07-08
New interface

torfjelde:
~~EDIT: I missed the latest suggestion from @kai in issue #6: https://github.com/TuringLang/Bijectors.jl/issues/6#issuecomment-509245069. Seems like better approach!~~
EDIT 2: I've now tried incorporating the suggested interface.

# Overview
This is a proposal for a new interface related to discussion in issue #6. It introduces
- `Bijector` and `Inversed{<: Bijector} <: Bijector` which represents an invertible transform and its inverse, respectively, with the following methods:
  - `transform(b::Bijector, x)` and `transform(ib::Inversed{<: Bijector}, y)`: transforms from original variable space to transformed space, and vice versa.
  - `inv(b::Bijector, y)`: simply returns the `Inversed{<: Bijector}`.
  - `logabsdetjac(b::Bijector, x)` and `logabsdetjac(b::Inversed{<: Bijector}, y)`: computes the `logabsdet` of the Jacobian of the transformation and its inverse. Inverse defaults to `-logabsdetjac(ib.orig, transform(ib, y))`
  - [OPTIONAL] `jacobian(b::Bijector, x)`: this is currently only used in default impl of `logabsdetjac` for `ADBijector`, thus only requires impl if `logabsdetjac` is not implemented (further than default impl).
- `Composed{<: Bijector}` which allow composing bijectors
  - `compose(bs...)`: returns the `Composed` object of the bijectors `bs`
  - `∘(b1::Bijector, b2::Bijector)`: results in a composition of the two bijectors (first `b2`, then `b1`)
- `UnivariateTransformed <: UnivariateDistribution{Continuous}` and `MultivariateTransformed <: MultivariateDistribution{Continuous}` both which implement the correspoding Distributions.jl interfaces:
  - `logpdf(td, y)`
  - `rand(td)`
  - `logpdf_with_jac(td, y)`: returns both `logpdf(td, y)` and `logabsdetjac(td, y)`, saving repetitive computation in cases where both these quantities are wanted. I believe this is often enough to warrent such a method.
- `transformed(d::Distribution, b::Bijector)`: returns a transformed distribution using `b` as the bijector.
  - Default `transformed(d::Distribution) = transformed(d, DistributionBijector(d))` is provided, where `DistributionBijector` is a proxy for the `link` and `invlink` already in Bijectors.jl (`logabsdetjac` is provided by AD)

For the user to implement their own `UserBijector`, the easiest approach is then to make `UserBijector{AD} <: ADBijector{AD}` and then they only _need_ to implement `transform(b::UserBijector, x)` and `transform(ib::Inversed{UserBijector}, y)`.

# Examples
## Composition
```julia
d = Beta()
b = DistributionBijector(d)
b⁻¹ = inv(b)

id = Identity()

x = rand(d)
(b ∘ b⁻¹)(x) ≈ id(x) # => `true`
```
## ADVI
```julia
# Usage in ADVI
d = InverseGamma()
b = DistributionBijector(d)    # (0, ∞) → ℝ
ib = inv(b)                    # ℝ → (0, ∞)
td = transformed(Normal(), ib) # x ∼ 𝓝(0, 1) then f(x) ∈ (0, ∞)
x = rand(td)                   # ∈ (0, ∞)
0 < x < Inf                    # => true ✓

logpdf(td, x) != logpdf(d, x)  # => true ✓
```

## Finite-set Stein Discrepancy (FSSD) test
This is a fun little (and accidental!) "outside-of-Turing"-application of this interface.

I have a package called `KernelGoodnessOfFit` which performs a goodness-of-fit test for a given model distribution `q <: Distribution`. The issue with the FSSD test is that it assumes the distribution to have support on the entire real line, and thus does not work for distributions such as `Beta`. With this interface for `Bijectors` we can simply plug in the `UnivariateTransformed{Beta}` and the test works! (in the sense that the false-negative/rejection ratio is less than the specified size of the test)

I know this is a _frequentist_ application; please don't shoot me!

```julia
using Bijectors
using KernelGoodnessOfFit

using Random; Random.seed!(123)
using ForwardDiff
function Distributions.gradlogpdf(d::ContinuousUnivariateDistribution, x::Real)
    ForwardDiff.derivative(z -> logpdf(d, z), x)
end

p = Beta(2, 3)         # true distribution
td = transformed(p)    # transformed
q = Beta()             # model distribution

α = 0.05               # size of the test
n = 100                # number of samples
num_experiments = 100  # number of experiments to run

# Under H₀: p = q
total = 0.0            # rejection ratio
for i = 1:num_experiments
    xs = reshape(rand(p, n), 1, :)
    ys = td.transform.(xs)

    res = FSSDopt(ys, td; β_H₁ = 0.01)

    total += res.p_val ≤ α
end
total / n              # => 0.03 ≤ 0.05 (✓)

# Under H₁: p ≠ q
total = 0.0            # rejection ratio
for i = 1:num_experiments
    xs = reshape(rand(q, n), 1, :)
    ys = td.transform.(xs)

    res = FSSDopt(ys, td; β_H₁ = 0.01)

    total += res.p_val ≤ α
end
total / n              # => 0.83 "close" to 1.0 (✓)
```
It's of course not directly related to what we do in Turing, but it shows a very neat application that immediately follows from this interface.

## Normalizing flows (thanks to @sharanry!)
```julia
d = MvNormal(zeros(10), ones(10))
b = PlanarLayer(10)

flow = transformed(d, b) # <= TransformedDistribution

y = rand(flow)           # samples from `MvNormal` and transforms using `b`
logpdf(flow, y)          # computes the logpdf using the inverse of `b`

# more efficiently (in certain cases), we can compute logpdf in forward pass
x = rand(d)
logpdf_forward(flow, x)

# Want to fit the flow? No problem!
using Tracker
b = PlanarLayer(10, param)                  # construct parameters using `param`
flow = transformed(d, b)
rand(flow)                                  # <= TrackedArray

x = rand(flow.dist)
Tracker.back!(logpdf_forward(flow, x), 1.0) # backprob through flow
```
Currently we have `PlanarFlow` and `RadialFlow` implemented, but implementing other flows is fairly straightforward.

# Issues / Notes
- [x] Composition in mathematics is read right-to-left, but in the current implementation `transform(Composed{<: Bijector}, x)` is implemented by applying transforms from left-to-right; should this be changed?
- [x] Ideally, we'd implement the `logabsdetjac` "manually" rather than using AD for these known transformations.
- [x] Currently depend on `Turing` for the different backend-types; probably don't want to do this. Need to decide how to handle this.
- [x] Type-stability; difficult with current `Composed` type.
- [X] ~~Should we use `logdetinvjac` or `logdetjac`? They're equivalent up to minus sign, but inside the transformed distributions we'll mostly need the `logdetinvjac` (though could always do `logdetjac(td.transform, transform(td.transform, y))`~~
  - Default: `logabsdetjac(ib::Inversed{<: Bijector}, y) = -logabsdetjac(ib.orig, transform(ib, y))`
  - Can then just override `logabsdetjac(ib::SpecificBijector, y)` to in cases and corresponding for `Inversed{<:SpecificBijector}` to get both.
- [X] ~~The interface is potentially a bit verbose at the moment, can do:~~
  - ~~Rename fields of the transformed distributions to `d::Distribution` and `b::Bijector`~~
  - ~~`transform(d::Distribution, y) = transform(d.transform, y)` and similarily for `inverse`~~


xukai92:
Some update of https://github.com/TuringLang/Bijectors.jl/issues/6#issuecomment-509245069:

I think we should add a `transform` that returns the transformed r.v. only and `forward` returns both the transformed r.v and logdetabsjacob. Why? Because in cases we only need "sample" so the transformed r.v. is necessay while computing the logdetabsjacob would be a waste of time.

xukai92:
I will give a try if things haven't been moved here.

torfjelde:
@xukai92 Agreed!

willtebbutt:
I'm slightly itchy about using a vector to implement the composition operation. As you point out @torfjelde it will probably create some type instability issues. Could we instead use a `Tuple` so that we can guarantee type stability? Or do we imagine that we're going to need to compose a large number of heterogeneous transformations?

torfjelde:
@willtebbutt After messing about with trying to get type-stability for VI, I think making it a `Tuple` is a good idea. If we also make `compose(ts...) = Composed(ts...)` which @xukai92 had in his original proposal, the constructor of composed also becomes type-stable. Granted, this means we lose the "optimization" of turning `b ∘ b⁻¹ = Identity()`. But, a quick test, using `compose(ts...) = `Composed(ts...)`:
 ```
@btime b1 ∘ b2  # => 1.7 μs for Array; 600 ns for Tuple; 25 ns for `Compose(ts...)` with Tuple
```
So it seems worth it to leave the "optimization" up to the user.

**Comment by willtebbutt** [comment]
@torfjelde could you explain the difference between your tuple implementation and splatted implementation please? I had imagined something like
```julia
struct Composed{Tts<:Tuple} <: Bijector
    ts::Tts
end
```
so the type of each of the things being composed should be known. Alternatively, we could just not type the stuff that's contained e.g.
```julia
struct Composed{Tts} <: Bijector
    ts::Tts
end
```
and leave it up to the user to use their preferred iterable data structure i.e. they could choose to use either a `Vector` or some `Tuple`

xukai92:
Using `Tuple` sounds like a very good idea for me. 

I also thinked about this but choosed to use vector only because I want a easy way to do `push!` to a composed transformation + for deep learning the computational bottleneck was on GPUs. But certainly we need this for Turing.jl.

willtebbutt:
> I also thinked about this but choosed to use vector only because I want a easy way to do push! to a composed transformation 

Fair enough @xukai92. How critical is the `push!` interface? Does the composition interface suffice in your opinion?

torfjelde:
> @torfjelde could you explain the difference between your tuple implementation and splatted implementation please?

@willtebbutt The difference is in the constructor. For the tuple-impl we have to do:
```
compose(ts...) = Composed(ts)  # <= type-stable
```
for it to be type-stable. If we drop type-stability from the constructor we can do
```
function compose(ts...) # <= NOT type-stable
    res = []
    
    for b ∈ ts
        if b isa Composed
            # "lift" the transformations
            for b_ ∈ b.ts
                push!(res, b_)
            end
        else
            if (length(res) > 0) && (res[end] == inv(b))
                # remove if inverse
                pop!(res)
            else
                push!(res, b)
            end
        end
    end

    length(res) == 0 ? IdentityBijector : Composed([res...])
end
```
which will
1. "Un-wrap" composed `Compositions` so that we don't get _nested_ compositions.
2. Drop sequential application of bijector and it's inverse, i.e. `b ∘ inv(b) == Identity()`
If we want type-stability in the constructor, I don't see how we can implement the above. I think 1. is quite preferable to have, while 2. can always just be left to the user.

We can of course drop type-stability of constructor but keep the `Tuple` implementation. We'd still benefit from type-stability, but less so. This is sort of similar to what we're currently doing for `TypedVarInfo`: not type-stable constructor, but type-stable operations. I'm just thinking we might be composing functions more often that creating instances of `VarInfo`, thus having a type-stable constructor in this case would be more preferable.

> Alternatively, we could just not type the stuff that's contained

Agree; that's what I did locally to try it out.

xukai92:
> > I also thinked about this but choosed to use vector only because I want a easy way to do push! to a composed transformation
> 
> Fair enough @xukai92. How critical is the `push!` interface? Does the composition interface suffice in your opinion?

It would work but just a little bit trickier as I was playing with an idea on adding or removing the transformations during training. I could build a new model each time I need to change, or I can push or pop.

> @willtebbutt The difference is in the constructor. For the tuple-impl we have to do:

As I said in Slakc I feel this is a little bit over-optimisation. If it doens't give much performance gain we'd better not to do this automatic change which is not expected by user inituitivly.

xukai92:
Or maybe add a flag called to trigger this behaviour.

torfjelde:
> It would work but just a little bit trickier as I was playing with an idea on adding or removing the transformations during training. I could build a new model each time I need to change, or I can push or pop.

One thing to note about this though, depending on how often you add/remove, is that with the type-stable `Bijector` this would lead to possibly noticiable compilation overhead. But if we do the "generic" container approach, the user can then just choose to use `Array` instead.

willtebbutt:
> As I said in Slakc I feel this is a little bit over-optimisation. If it doens't give much performance gain we'd better not to do this automatic change which is not expected by user inituitivly.

The performance gain really depends on the context: as you say if we're working in the DL setting, then this optimisation is unnecessary. However, if we're expecting to use these in other settings then it's probably not unreasonable to ask for type stability. The other option is, again, to provide both interfaces and just expect the `push!` interface to work when a `Vector` of bijectors is provided, and code to be type-stable when a `Tuple` of bijectors (or some other type-stable container) is provided.

Another alternative is to have two different types, which have different performance characteristics and provide different behaviour. I don't have any good suggestions for names though.

@torfjelde : thanks for the explanation of your implementations

> "Un-wrap" composed Compositions so that we don't get nested compositions.

I agree that this is generally tricky to do in a type-stable way, but it's not clear to me why we mind if the user creates nested compositions. Is there something obvious that I'm missing?

> Drop sequential application of bijector and it's inverse, i.e. b ∘ inv(b) == Identity()

I agree with you that this is a nice-to-have rather than something that's especially important.

xukai92:
> The performance gain really depends on the context: as you say if we're working in the DL setting, then this optimisation is unnecessary. However, if we're expecting to use these in other settings then it's probably not unreasonable to ask for type stability. The other option is, again, to provide both interfaces and just expect the push! interface to work when a Vector of bijectors is provided, and code to be type-stable when a Tuple of bijectors (or some other type-stable container) is provided.

I'm talking about optimizations for the type stable implementations, i.e. the un-wrapping and getting rid of inverse operations. I completely agree that we need type stable version.

torfjelde:
I've now just changed it to
```
struct Composed{A} <: Bijector
    ts::A
end
```
with the implementations of `forward`, `inv` and `transform` working for any iterable and reversible container, e.g. `Vector` and `Tuple`. 

> Another alternative is to have two different types, which have different performance characteristics and provide different behaviour. I don't have any good suggestions for names though.

I kind of like the idea of providing two types, if we can do so in a non-confusing manner. Maybe we could something similar to `VarInfo`:
```
const TypedComposed = Composed{<: Tuple}
const UntypedComposed = Composed{<: AbstractVector}
```
Uhmm, the naming is a bit weird though. And we have to make a choice for `∘`: does it automatically create the type-stable or the type-unstable one?

> I agree that this is generally tricky to do in a type-stable way, but it's not clear to me why we mind if the user creates nested compositions. Is there something obvious that I'm missing?

When thinking about it, probably not too much. @xukai92 suggested the only thing is additional stacks being allocated, which I think is correct. In the end, when evaluating `transform`, we're going to have "nested" for-loops for each `Composed`. "Un-wrapping" would circumvent this. But I think you guys are right, it probably won't matter much :+1:

I originally did it because it seemed like what a composition ought to do (though the evaluation is of course still the same) :) But I agree, there are probably very few reasons for it perf-wise :+1:

xukai92:
> And we have to make a choice for ∘: does it automatically create the type-stable or the type-unstable one?

Let make this create type-stable version :)

torfjelde:
> Let make this create type-stable version :)
Great! That's what I defaulted to:)

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+"Transforms the input using the bijector."
+transform(b::T1, y::T2) where {T<:Bijector,T1<:Inversed{T},T2} =
+    error("`transform(b::$T1, y::$T2)` is not implemented.")
+
+"Computes both `transform` and `logabsdetjac` in one forward pass."
+forward(b::T1, y::T2) where {T<:Bijector,T1<:Inversed{T},T2} = 
+    error("`forward(b::$T1, y::$T2)` is not implemented.")
+
+
+transform(b::Bijector) = x -> transform(b, x)
```
willtebbutt: Could you please explain the semantics of this? Should I read `transform` as "give me the transform associated with this Bijector"?

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+
+# defaults implementation for inverses
+logabsdetjac(ib::Inversed{<: Bijector}, y) = - logabsdetjac(ib.orig, transform(ib, y))
+
+inv(b::Bijector) = Inversed(b)
+inv(ib::Inversed{<:Bijector}) = ib.orig
+
+# AD implementations
+function jacobian(b::ADBijector{<: ForwardDiffAD}, y::Real)
+    ForwardDiff.derivative(z -> transform(b, z), y)
```
willtebbutt: To be super picky, could you please put a return statement in from of all of these single line function bodies? (See Turing style guide section on method definitions)

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+    Tracker.gradient(z -> transform(b, z), y)[1]
+end
+function jacobian(b::Inversed{<: ADBijector{<: TrackerAD}}, y::AbstractVector{<: Real})
+    Tracker.jacobian(z -> transform(b, z), y)
+end
+
+# TODO: allow batch-computation, especially for univariate case?
+"Computes the absolute determinant of the Jacobian of the inverse-transformation."
+function logabsdetjac(b::ADBijector, x::Real)
+    log(abs(jacobian(b, x)))
```
willtebbutt: On a single line? Should be under the char limit

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+end
+function logabsdetjac(b::ADBijector, x::AbstractVector{<:Real})
+    fact = lu(jacobian(b, x), check=false)
+    issuccess(fact) ? log(abs(det(fact))) : -Inf # TODO: or smallest possible float?
```
willtebbutt: 1. Return please :)
2. Maybe we want to error if this fails, rather than throw a big large number?

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+    ts::A
+end
+
+compose(ts...) = Composed(ts)
+
+# The transformation of `Composed` applies functions left-to-right
+# but in mathematics we usually go from right-to-left; this reversal ensures that
+# when we use the mathematical composition ∘ we get the expected behavior.
+# TODO: change behavior of `transform` of `Composed`?
+∘(b1::B1, b2::B2) where {B1 <: Bijector, B2 <: Bijector} = compose(b2, b1)
```
willtebbutt: 1. Style: `∘(b1::Bijector, b2::Bijector)` please :)
2. I completely agree with the composition order. Also agree that maybe we want to change the order of stuff in `Composed`.

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+
+inv(ct::Composed) = Composed(map(inv, reverse(ct.ts)))
+
+# TODO: can we implement this recursively, and with aggressive inlining, make this type-stable?
```
willtebbutt: Probably want to dispatch for this. If the container is a tuple, then recursion + inlining is probably a good move. If it's a vector, then probably not.

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+        res = (rv=res′.rv, logabsdetjac=res.logabsdetjac + res′.logabsdetjac)
+    end
+    return res
+end
+
+##############################
+# Example bijector: Identity #
+##############################
+
+struct Identity <: Bijector end
```
willtebbutt: Do we actually need this Bijector for anything? If not, could probably just get rid of it (even if it is nice to have in principle)

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+(b::Logit)(x) = transform(b, x)
+
+logabsdetjac(b::Logit{<:Real}, x::Real) = log((x - b.a) * (b.b - x) / (b.b - b.a))
+forward(b::Logit, x::Real) = (rv=transform(b, x), logabsdetjac=-logabsdetjac(b, x))
+
+
+#######################################################
+# Constrained to unconstrained distribution bijectors #
+#######################################################
+struct DistributionBijector{AD, D} <: ADBijector{AD} where D <: Distribution
```
willtebbutt: Can re-write this more compactly:
```julia
struct DistributionBijector{AD, <:Distribution} <: ADBijector{AD}
```

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+# Simply uses `link` and `invlink` as transforms with AD to get jacobian
+transform(b::DistributionBijector, x) = link(b.dist, x)
+transform(ib::Inversed{<: DistributionBijector}, y) = invlink(ib.orig.dist, y)
+(b::DistributionBijector)(x) = transform(b, x)
+
+"Returns the constrained-to-unconstrained bijector for distribution `d`."
+bijector(d::Distribution) = DistributionBijector(d)
+
+# Transformed distributions
+struct UnivariateTransformed{D, B} <: Distribution{Univariate, Continuous} where {D <: UnivariateDistribution, B <: Bijector}
```
willtebbutt: See previous comment regarding compactness of struct type parameters

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+    dist::D
+    transform::B
+end
+
+struct MultivariateTransformed{D, B} <: Distribution{Multivariate, Continuous} where {D <: MultivariateDistribution, B <: Bijector}
```
willtebbutt: See previous comment regarding compactness of struct type parameters

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+transformed(d::MultivariateDistribution, b::Bijector) = MultivariateTransformed(d, b)
+transformed(d) = transformed(d, bijector(d))
+
+# can specialize further by
+bijector(d::Normal) = IdentityBijector
+bijector(d::Beta{T}) where T <: Real = Logit(zero(T), one(T))
+
+##############################
+# Distributions.jl interface #
+##############################
```
willtebbutt: :heart:

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+
+# size
+Base.length(td::MultivariateTransformed) = length(td.dist)
+
+# logp
+function logpdf(td::UnivariateTransformed, y::T where T <: Real)
```
willtebbutt: Please remove `where`

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+    # logpdf(td.dist, transform(inv(td.transform), y)) .+ logabsdetjac(inv(td.transform), y)
+    logpdf_with_trans(td.dist, transform(inv(td.transform), y), true)
+end
+function _logpdf(td::MultivariateTransformed, y::AbstractVector{T} where T <: Real)
```
willtebbutt: Please use more compact form: `y::AbstractVector{<:Real}`

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+    # logpdf(td.dist, transform(inv(td.transform), y)) .+ logabsdetjac(inv(td.transform), y)
+    logpdf_with_trans(td.dist, transform(inv(td.transform), y), true)
+end
+
+# TODO: implement these using analytical expressions?
+function logpdf_with_jac(td::UnivariateTransformed, y::T where T <: Real)
```
willtebbutt: Please remove `where`

src/interface.jl
```diff
@@ -0,0 +1,252 @@
...
+    z = logabsdetjac(inv(td.transform), y)
+    return (logpdf(td.dist, transform(inv(td.transform), y)) .+ z, z)
+end
+
+function logpdf_with_jac(td::MultivariateTransformed, y::AbstractVector{T} where T <: Real)
```
willtebbutt: Please remove `where`

src/interface.jl
torfjelde: The idea was to convert the `Bijector` into a callable, yeah. It's more a left-over from the previous iteration of the interface. Now I think it's better to implement a caller for the struct instead, e.g. `(b::Logit)(x) = transform(b, x)`.

_But_ the issue is that we can't do this default implementation for all bijectors, i.e. `(b::B)(x) where B <: Bijector` is not possible. So there might be a reason to keep it around? I think the options we have here are:
1. Drop `transform(b, x)` entirely, and require implementation of the "call" method for each bijector. This has the downside of not allowing an over-arching doc-string but instead we have to write a doc-string for each `(b::SpecificBijector)(x)` definition.
2. Drop `transform(b)`, use `transform(b, x)` internally where we need the code to work on general bijectors while providing a call-method for each of the Bijectors defined in Bijectors.jl. Then require the user to implement `transform(b, x)` and let `(b::SpecificBijector)(x)` be optional. The downside with this approach is that I can easily see everyone wanting to use the `(b::SpecificBijector)(x)` call, which might lead to a lot of errors if a package depending on Bijectors.jl is built on the assumption that every Bijector provides a call-method. On the bright side, we can provide over-arching docstrings and it's easier/intuitve for the user to just "inspect" the signature of `transform`.
3. Keep this `transform(b)` method which returns the transformation associated with the Bijector.

I'm partial to 1. myself. As long as we make it very clear in the documentation, it should be fine. And it's probably the most intuitive way. Then, I suppose, we could also just drop the `transform(b, x)` and implement the transformation through the call-method?

Also, I appreciate the italics in "really nice work"! Thanks! It's worth reminding folks that @xukai92 was the one who came up with the `Inversed` and `Composed` idea, which I'm quite fond of myself:)

src/interface.jl
torfjelde: Sorry! I'l fix all the style issues asap

src/interface.jl
torfjelde: It was `logabsdet` before, which would throw an error. I think I agree with you though. The reason why I added it is because of performance and numerical errors might lead to a singular Jacobian. I was thinking it could be useful instead return a very small number, still allowing AD to go through whatever else was included in the computation, e.g. ELBO in ADVI. This way, the parameters could potentially make their way "out of" the numerically unstable domain and optimization could run as normal. I say performance because I was under the impression that it's more expensive to have a `catch` vs. a `if -Inf` outside of the call? Either way, the user has to handle it. Would it be cheaper to warn the user but still use the `-Inf`?

_But_ I don't like it:) It's hacky. One the other hand it made LDA work, but for different reasons. In LDA we need Dirichlet distribution, whose bijector is using clamping (see `_clamp` in the `invlink` function). This lead of course lead to singular Jacobian. Even so, I could start out with singular Jacobian, run it for a while, and eventually it would become non-singular because we'd reach a domain were clamping was not necessary.

src/interface.jl
torfjelde: 1. Fixed, also fixed knowledge. Was left-over from me not properly understanding the type-system:)
2. :+1:

src/interface.jl
torfjelde: Think I got them all :+1:

src/interface.jl
xukai92: I'm up to option 1 and 2.

src/interface.jl
xukai92: Not very sure what's the best thing to do TBH. Was throwing an error the original behaviour? Maybe we should stick to it as this is an interface PR anyway, and solve this issue later. And I'm not very sure if returning `-Inf` would still make the trianing valid as there are no gradient through this function anymore right? Does it mean the gradient is wrong then?

torfjelde: > Was throwing an error the original behaviour? Maybe we should stick to it as this is an interface PR anyway, and solve this issue later. And I'm not very sure if returning -Inf would still make the trianing valid as there are no gradient through this function anymore right? 

Well, before we were using `logabsdet` which throws an error because it uses LU-decomposition to compute the determinant, which is only possible if `det(J) != 0`. Now we're using `log(abs(det(J)))` instead, which means that we're either going to have an error when calling `log(0)` or return `-Inf`; either I think I'm fine with.

Returning `-Inf` doesn't help with computing the gradient, but in whatever computation you could do
```julia
if logabsdetjac(b, x) == -Inf
    # deal with it as you please
    ...
end
```
compared to catching the error and dealing with the error instead.

Or we could go back to using `logabsdet`.

> Does it mean the gradient is wrong then?

Computing the jacobian for the Dirichlet distribution potentially involves clamping, which makes jacobian singular and thus `logdet = -Inf`. So the gradient of this is of course not well-defined, if that's what you mean?

src/interface.jl
torfjelde: I just tried the following:
```julia
_transform(x, b1::Bijector, b2::Bijector) = transform(b2, transform(b1, x))
_transform(x, b::Bijector, bs::Bijector...) = _transform(transform(b, x), bs)  # note ordered differently from above
transform(cb::Composed{<: Tuple}, x) = _transform(x, cb.ts...)
```
And this seems to do the trick!

src/interface.jl
torfjelde: It's use to have as a default, or at the very least, return for distributions like `Normal` and so on.
The only other alternative I can think of is to define `transform` and so on for `Base.identity`, but this is of course not `<: Bijector` and so we lose the ability to dispatch on bijectors only.

So I actually think it's necessary. We can remove it in the future, if we realize it's redundant?:)

src/interface.jl
torfjelde: We need the type for the `dist::D` field, so I don't think we can

willtebbutt: You're totally right. What we could do though is
```julia
struct DistributionBijector{AD, D<:Distribution} <: ADBijector{AD}
```

src/interface.jl
torfjelde: Let's say we go with 1. initially then?

torfjelde:
Just merged @sharanry's normalizing flows PR! :tada: So now this PR also includes normalizing flows, thanks to him :heart:

src/interface.jl
```diff
@@ -0,0 +1,664 @@
...
+    res = forward(b1, x)
+    return _logabsdetjac(res.rv, bs...) + res.logabsdetjac
+end
+logabsdetjac(cb::Composed, x) = _logabsdetjac(x, cb.ts...)
+
+# recursive implementation of `forward`
+function _forward(f, b1::Bijector, b2::Bijector)
+    f1 = forward(b1, f.rv)
+    f2 = forward(b2, f1.rv)
+    return (rv=f2.rv, logabsdetjac=f2.logabsdetjac + f1.logabsdetjac + f.logabsdetjac)
```
sharanry: Since we initialize `f.logabsdetjac` to 0, this gives an error when we send a batch of samples. Please make it `.+` .

src/interface.jl
```diff
@@ -0,0 +1,664 @@
...
+end
+function _forward(f, b::Bijector, bs::Bijector...)
+    f1 = forward(b, f.rv)
+    f_ = (rv=f1.rv, logabsdetjac=f1.logabsdetjac + f.logabsdetjac)
```
sharanry: Same here.

src/interface.jl
```diff
@@ -0,0 +1,664 @@
...
+end
+
+function forward(cb::Composed, x, logjac)
+    rv = x
+    logjac_ = logjac
+    
+    for t in cb.ts
+        res = forward(t, rv)
+        rv = res.rv
+        logjac_ = res.logabsdetjac + logjac_
```
sharanry: Same error can occur here too.

src/interface.jl
```diff
@@ -0,0 +1,664 @@
...
+
+julia> forward(cb, x)
+ERROR: MethodError: no method matching +(::Array{Float64,1}, ::Float64)
+  ...
+julia> forward(cb, x, zeros(size(x, 2)))
+(rv = [1.10887 0.32029 -0.704563; -0.639206 -1.97935 -0.243419], logabsdetjac = [0.018534, 1.46352e-5, 0.00521633])
+```
+    
+"""
+forward(b::Bijector, x) = forward(b, x, zero(eltype(x)))
```
sharanry: @torfjelde This is the root cause of the error. Please read the next comments for context.

torfjelde: This is why we provided the `forward(b::Bijector, x, logjac)` method, _but_ I've now changed it to what you suggested as I agree it's a much better approach!:)

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+# Example bijector: Identity #
+##############################
+
+struct Identity <: Bijector end
+(::Identity)(x) = x
+(::Inversed{Identity})(y) = y
+
+forward(::Identity, x) = (rv=x, logabsdetjac=zero(x))
+
+logabsdetjac(::Identity, y::T) where T <: Real = zero(T)
```
xukai92: Plz remove spaces between `<:` and consider add `{`,`}` to two sides if the line is not too long.

xukai92: Same for all.

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+
+logabsdetjac(b::Logit{<:Real}, x) = @. log((x - b.a) * (b.b - x) / (b.b - b.a))
+
+#############
+# Exp & Log #
+#############
+
+struct Exp <: Bijector end
+struct Log <: Bijector end
+const exp_b = Exp()
```
xukai92: Why do we need this?

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+(b::Scale{<: AbstractVector{<: Real}})(x::AbstractMatrix{<: Real}) = x * b.a
+
+inv(b::Scale) = Scale(inv(b.a))
+inv(b::Scale{<: AbstractVector}) = Scale(inv.(b.a))
+
+# TODO: should this be implemented for batch-computation?
+# There's an ambiguity issue
+#      logabsdetjac(b::Scale{<: AbstractVector}, x::AbstractMatrix)
+# Is this a batch or is it simply a matrix we want to scale differently
+# in each component?
```
xukai92: Does the behaviour different for this two cases?

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+
+####################
+# Simplex bijector #
+####################
+struct SimplexBijector{T} <: Bijector where T end
+
+const simplex_b = SimplexBijector{Val{false}}()
+const simplex_b_proj = SimplexBijector{Val{true}}()
+
+# the following implementations are basically just copy-paste from `invlink` and
```
xukai92: `the` -> `The`

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+function (b::SimplexBijector{Val{proj}})(x::AbstractVector{T}) where {T, proj}
+    y, K = similar(x), length(x)
+
+    ϵ = _eps(T)
+    sum_tmp = zero(T)
+    @inbounds z = x[1] * (one(T) - 2ϵ) + ϵ # z ∈ [ϵ, 1-ϵ]
+    @inbounds y[1] = StatsFuns.logit(z) + log(T(K - 1))
+    @inbounds @simd for k in 2:(K - 1)
+        sum_tmp += x[k - 1]
+        # z ∈ [ϵ, 1-ϵ]
```
xukai92: What are these lines?

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+bijector(d::PositiveDistribution) = log_b
+bijector(d::SimplexDistribution) = simplex_b_proj
+
+_union2tuple(T1::Type, T2::Type) = (T1, T2)
+_union2tuple(T1::Type, T2::Union) = (T1, _union2tuple(T2.a, T2.b)...)
+_union2tuple(T::Union) = _union2tuple(T.a, T.b)
+
+bijector(d::Kolmogorov) = Logit(zero(eltype(d)), zero(eltype(d)))
+for D in _union2tuple(UnitDistribution)[2:end]
+    # Skipping Kolmogorov because it's a DataType
```
xukai92: Where did we skip? I guess it's `[2:end]`. Is it possible to comment it or make it explicity? E.g. use a `if` statement to skip.

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+    @eval bijector(d::$D{T}) where T <: Real = Logit(zero(T), one(T))
+end
+
+# FIXME: Can we make this type-stable?
```
xukai92: Better to leave a name here so people working on it later would know how to contact.

test/interface.jl
```diff
@@ -0,0 +1,242 @@
...
+            Logistic(),
+            LogNormal(1.0, 2.5),
+            Normal(0.1, 2.5),
+            Pareto(),
+            Rayleigh(1.0),
+            TDist(2),
+            TruncatedNormal(0, 1, -Inf, 2),
+        ]
+        
+        for dist in uni_dists
```
xukai92: Have we tested if the inverse function can map things back. Something like
```julia
y = rand(td)
x = inv(td.transform)(y)
@test y == td.transform(x)
```

test/interface.jl
```diff
@@ -0,0 +1,242 @@
...
+        @test b(x) == link(d, x)
+    end
+
+    @testset "Multivariate" begin
+        vector_dists = [
+            Dirichlet(2, 3),
+            Dirichlet([1000 * one(Float64), eps(Float64)]),
+            Dirichlet([eps(Float64), 1000 * one(Float64)]),
+            MvNormal(randn(10), exp.(randn(10))),
+            # MvLogNormal(MvNormal(randn(10), exp.(randn(10)))),
```
xukai92: Why do we comment out this one?

src/interface.jl
torfjelde: In one case we want to return a vector of values with length as long as the batch-size, while in the other case we simply want to return a single number, right?

src/interface.jl
torfjelde: You mean the comments or the entire thing?
This is just copy-pasted from the `logpdf_with_trans`, `link` and `invlink` for `SimplexDistribution`. So those questions are probably better directed at the person who implemented this, but if you're referring to the comment I think it's just stating that we've done a "small" rescaling to ensure that `z` is _not_ on the boundary. I just assumed that we do this because it ensures that the transformation is invertible or differentiable. No closed interval can be isomorphic to the real-line, so by instead considering `[ϵ, 1-ϵ]` we're instead considering the "largest"_open_ interval inside of `[0 ,1]` which indeed _is_ isomorphic to the real line, i.e. we have a bijection:)

If you're asking _why_ we have this comment, I don't know but I'm gladwe do! Personally I find it helpful.

src/interface.jl
torfjelde: You're right, and yeah we can do an `if` statement instead :+1:

src/interface.jl
torfjelde: Good idea; will do!

test/interface.jl
torfjelde: Good catch; we should have such a test

src/interface.jl
torfjelde: Why do we need `exp_b`? Maybe we don't. I just figured that these are stateless and so we could just use a single instance which we plug everywhere instead of constructing `n` different `Exp` or whatever.

I suspect the Julia compiler will just do the most efficient thing for us anyway, so might very well be redundant, but I figured I'd be on the safe side.

Remove and just use `Exp()` instead?

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+import Base: inv, ∘
+
+import Random: AbstractRNG
+import Distributions: logpdf, rand, rand!, _rand!, _logpdf
+
+#######################################
+# AD stuff "extracted" from Turing.jl #
+#######################################
+
+abstract type ADBackend end
```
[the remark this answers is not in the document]
torfjelde: I think it has other use-cases. The one I have/had in mind was if the user wants to implement a `Bijector` but don't want to write out the `logabsdetjac` implementation or no nice expression exists. Then one would only need to implement `(b::UserADBijector{AD})(x)` and `(::Inversed{<: UserADBijector{AD}})(y)` and you'd be good:)

src/interface.jl
```diff
@@ -0,0 +1,655 @@
...
+    # we extract `data` so that we don't returne a `Tracked` type
+    return Tracker.data(Tracker.jacobian(b, y))
+end
+
+# TODO: allow batch-computation, especially for univariate case?
+"Computes the absolute determinant of the Jacobian of the inverse-transformation."
+logabsdetjac(b::ADBijector, x::Real) = log(abs(jacobian(b, x)))
+function logabsdetjac(b::ADBijector, x::AbstractVector{<:Real})
+    fact = lu(jacobian(b, x), check=false)
+    return issuccess(fact) ? log(abs(det(fact))) : -Inf # TODO: do this or not?
```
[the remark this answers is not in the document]
torfjelde: Yeah, I didn't conclude with anything so I just left it as-is. But sure, I'm okay with throwing a warning. Should we make it a "custom" warning or just call `logabsdet` and let it throw the `SingularException`?
I think I like a custom exception so that the user fully understands that the Jacobian is for some reason singular. Just getting a `SingularException` takes a bit more of a mental effort to track down

test/interface.jl
torfjelde: Numerical inaccuracies using `DistributionBijector` (which used `ADBijector` as the transform). But I've now implemented `bijector(d::LogMvNormal) = log_b`, which of course does not suffer from the same, so it's now fixed:)

src/interface.jl
torfjelde: Is this actually a part of the style-guide? Couldn't find it anywhere. I mean that `f(a::AbstractArray{<: Real})` should be `f(a::AbstractArray{<:Real})` and so on, i.e. no space with the `<:`. 

It seems a bit, uhmm, "cramped" to me. The whitespace inbetween makes it much easier to quickly see that the `<:` is _not_ a part of the type-name. E.g. `{T <: Real, S <: AbstractFloat}` vs. `{T<:Real, S<:AbstractFloat}`.

src/interface.jl
torfjelde: A good example is the `logabsdetjac` for `log` and `exp` transforms. It makes sense to let these act on vectors elementwise, etc. but when computing the `logabsdetjac(::Log, x::Vector)` we need to sum over the entires; but what if `x` is actually a vector representing a _batch_, then we should _not_ sum.

src/interface.jl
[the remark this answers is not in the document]
willtebbutt: Agreed. Helpful for prototyping. Might be helpful to add something similar to `CuArrays.allowscalar` to allow the performance-sensitive user to disable this though.

src/interface.jl
willtebbutt: Interesting that this isn't in the style guide. If someone could add it, that would be fantastic.

As regards the actual issue at hand, I'm personally quite pro- the space-free version. Picking a bit of julia at random as evidence, all of the [cholesky](https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/cholesky.jl) code uses even more aggressively space-free conventions than we do, which I'm also not in favour of -- I really do find it hard to read some of that.

Happy to change to a different convention if there's a consensus around something other than what we're doing at the minute and someone is filling to make the required changes to the codebase. Either way though, this should be documented in the style guide.

src/interface.jl
willtebbutt: Yeah, I think either @mohamed82008 or I added these comments at some point (probably @mohamed82008 ). Probably better to keep them around for exactly this type of situation.

src/interface.jl
torfjelde: > Interesting that this isn't in the style guide. If someone could add it, that would be fantastic.
It's not mentioned explicitly as far as I can tell, but there are examples using `{T<:Whatever}`. In those cases the "bad" examples are also using the same convention, so there's not a
```julia
# good
f(a::AbstractArray{<:T}) where {T} = ...

# bad
f(a::AbstractArray{<: T}) where T = ...
```
or something. I could add this example to the "Avoid unnecessary whitespace in these situations" section if you'd like.

> As regards the actual issue at hand, I'm personally quite pro- the space-free version. Picking a bit of julia at random as evidence, all of the cholesky code uses even more aggressively space-free conventions than we do, which I'm also not in favour of -- I really do find it hard to read some of that.

I'm probably missing something but to me they seem to be using the same convention?

willtebbutt: The good / bad examples would be great.

> I'm probably missing something but to me they seem to be using the same convention?

See [here](https://github.com/JuliaLang/julia/blob/c0478d87df6182932b78738be97cc108f0f93248/stdlib/LinearAlgebra/src/cholesky.jl#L347) for example. There isn't any whitespace, and I wouldn't advocate for this.

```julia
# bad
Union{StridedMatrix,RealHermSymComplexHerm{<:Real,<:StridedMatrix}}

# good
Union{StridedMatrix, RealHermSymComplexHerm{<:Real, <:StridedMatrix}}
```
i.e. always adding spaces after commas.

src/interface.jl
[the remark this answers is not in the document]
torfjelde: How about
```julia
struct SingularJacobianException{B} <: Exception where B <: Bijector
    b::B
end
Base.showerror(io::IO, e::SingularJacobianException) = print(io, "jacobian of $(e.b) is singular")

# TODO: allow batch-computation, especially for univariate case?
"Computes the absolute determinant of the Jacobian of the inverse-transformation."
function logabsdetjac(b::ADBijector, x::Real)
    res = log(abs(jacobian(b, x)))
    return isfinite(res) ? res : throw(SingularJacobianException(b))
end

function logabsdetjac(b::ADBijector, x::AbstractVector{<:Real})
    fact = lu(jacobian(b, x), check=false)
    return issuccess(fact) ? log(abs(det(fact))) : throw(SingularJacobianException(b))
end
```
Then
```julia
julia> struct NonInvertibleBijector{AD} <: ADBijector{AD} end

julia> (b::NonInvertibleBijector)(x) = clamp.(x, 0, 1)

julia> b = NonInvertibleBijector{Bijectors.ADBackend()}()
NonInvertibleBijector{Bijectors.ForwardDiffAD}()

julia> Bijectors.jacobian(b, [1.0, 10.0])
2×2 Array{Float64,2}:
 1.0  0.0
 0.0  0.0

julia> logabsdetjac(b, [1.0, 10.0])
ERROR: jacobian of NonInvertibleBijector{Bijectors.ForwardDiffAD}() is singular
Stacktrace:
 [1] logabsdetjac(::NonInvertibleBijector{Bijectors.ForwardDiffAD}, ::Array{Float64,1}) at /home/tor/.julia/dev/Bijectors/src/interface.jl:134
 [2] top-level scope at none:0
```

It's worth pointing out that such a check will not be performed for cases where `logabsdetjac` does _not_ use AD.

src/interface.jl
torfjelde: Now I follow; thanks!

src/interface.jl
torfjelde: What we could do is also parametrize `Scale` with an additional type, specifying the expected _input/output_ and then we can dispatch on this. E.g. for the `Exp` and `Log` we can do

```julia
struct Exp{T} <: Bijector end
struct Log{T} <: Bijector end
const exp_b = Exp{Real}()
const log_b = Log{Real}()

(b::Log)(x) = @. log(x)
(b::Exp)(y) = @. exp(y)

inv(b::Log{T}) where {T} = Exp{T}()
inv(b::Exp{T}) where {T} = Log{T}()

logabsdetjac(b::Log, x) = sum(log.(x))
logabsdetjac(b::Exp, y) = - sum(y)

logabsdetjac(b::Log{<: Real}, x::AbstractVector) = log.(x)
logabsdetjac(b::Exp{<: Real}, y::AbstractVector) = - y
logabsdetjac(b::Log{<: AbstractVector}, x::AbstractMatrix) = sum(log.(x); dims = 1)
logabsdetjac(b::Exp{<: AbstractVector}, y::AbstractMatrix) = - sum(y; dims = 1)
```

This leads to nice method-implementations using multiple dispatch, but feels very hacky :thinking:

torfjelde: OR we could adapt something similar to `Multivariate, Univariate`, etc. in `Distributions`. Then the above would be 
```julia
struct Scale{T, V} <: Bijector
    a::T
end

# sane defaults
(b::Scale)(x) = b.a * x
(b::Scale{<:Real})(x::AbstractArray) = b.a .* x

logabsdetjac(b::Scale, x) = log(abs(b.a))

# more specialized
function logabsdetjac(b::Scale{<:Real, Multivariate}, x::AbstractMatrix)
    return log(abs(b.a)) .* ones(size(x, 2))
end

File: src/interface.jl (hunk 30 above)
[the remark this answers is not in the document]
Comment (xukai92) [reply src/interface.jl]: An exception works too. I don't have strong opinion on this.

PS: I originall was suggesting to keep returning `-Inf` but in the meantime thorwing a warning message not an exception.

File: src/interface.jl (hunk 20 above)
Comment (xukai92) [reply src/interface.jl]: Looks good! Let's stick to Will's version.

Comment (xukai92) [reply src/interface.jl]: "i.e. always adding spaces after commas." is easy to remeber :)

File: src/interface.jl (hunk 21 above)
Comment (xukai92) [reply src/interface.jl]: Yes, I think that's more readble. 

When things are imutable plugging the constructor is fine.

```julia
julia> struct A end

julia> a1 = A()
A()

julia> a2 = A()
A()

julia> a1 == a2
true

julia> mutable struct B end

julia> b1 = B()
B()

julia> b2 = B()
B()

julia> b1 == b2
false
```

src/interface.jl
xukai92: > In one case we want to return a vector of values with length as long as the batch-size, while in the other case we simply want to return a single number, right?

Yes I get your point. Thanks.

The parametric idea looks promising. We prefer the one with `AbstractArray` as it also allows us to put `CuArray` which indicates the device.

src/interface.jl
xukai92: SGTM!

src/interface.jl
torfjelde: Sounds good to me :+1: I'll get it fixed asap

src/interface.jl
torfjelde: > The parametric idea looks promising. We prefer the one with AbstractArray as it also allows us to put CuArray which indicates the device.

That's an interesting idea; didn't think of that :+1:

xukai92:
> Composition in mathematics is read right-to-left, but in the current implementation transform(Composed{<: Bijector}, x) is implemented by applying transforms from left-to-right; should this be changed?

torfjelde:
> > Composition in mathematics is read right-to-left, but in the current implementation transform(Composed{<: Bijector}, x) is implemented by applying transforms from left-to-right; should this be changed?
> 
> Is this still true?

Yes; and it has already caused some confusion I believe. At least, I'm very much for making `compose(ts...) = Composed(reversed(ts))`. The user should really never have to write `Composed((b1, b2, b3))` anymore, so I think changing the order of `Composed.ts` isn't as crucial (and will require a bit more work)

xukai92:
I'm pro to stick to the mathematical convention. I think the current implementation already does so right?

torfjelde:
> I'm pro to stick to the mathematical convention. I think the current implementation already does so right?

Internally, i.e. `Composed.ts`, we represent in a left-to-right manner. So `Composed.ts[1]` is applied, then `Composed.ts[2]`, and so on. This is opposite of mathematical convention, but it's quite convenient for a recursive implementation (though not necessary). But `∘` is implemented so that `b1 ∘ b2 == Composed((b2, b1))`, which results in `b1(b2(x))`, which is what you'd expect from `∘`.

Currently, `compose(ts...) = Composed(ts)` but maybe this should be `compose(ts...) = Composed(reversed(ts))` OR we re-implement the transform of `Composed` to apply `Composed.ts[end]` first, and so on. My initial thought is that this will require an addition `reverse(cb.ts)` in the implementation which is not _as_ performant, right?

xukai92:
I feel it's better to make `∘` and `compose` consistent somehow. `compose(ts...) = Composed(reversed(ts))` sounds good but one the user side it would make confusion IMO. What if I just want a transfom do 1, 2 and 3? Users have to construct things by reversing them first. 

Maybe what we can do is to drop `compose` but implment `composel(ts)` and `composer(ts)` which are simply equivalences for `foldl(∘, ts)` and `foldr(∘, ts)`. In this case things are consistent and clear, indicated by `l` and `r`. And users have the choice to use whichever the one is convenient.

torfjelde:
> I feel it's better to make `∘` and `compose` consistent somehow. `compose(ts...) = Composed(reversed(ts))` sounds good but one the user side it would make confusion IMO. What if I just want a transfom do 1, 2 and 3? Users have to construct things by reversing them first.
> 
> Maybe what we can do is to drop `compose` but implment `composel(ts)` and `composer(ts)` which are simply equivalences for `foldl(∘, ts)` and `foldr(∘, ts)`. In this case things are consistent and clear, indicated by `l` and `r`. And users have the choice to use whichever the one is convenient.

That's not a bad idea :+1: Only thing is though, `composel(ts...) = foldl(∘, ts)` actually leads to right-to-left application because `∘` reverses the order, right? I find this a bit confusing, as I'd expect the `composel` to give me application from left-to-right. I'm also not certain if the fact that the user has to reverse 1, 2, and 3 is much of a problem. If you're already putting them in a list or something, doing it in the reverse order isn't much of an issue. But I'd be fine with doing what you suggested, though maybe `composel(ts...) = foldr(∘, ts)` instead?

There's another thing I've been thinking about; you can do multiple dispatch on something like `Composed{B1, B2, B3}` where `B1`, `B2`, and `B3` are specific types. If all the "standard" ways of constructing a `Composed` are such that we get nested compositions, it becomes difficult to do something like what I just mentioned. Because of this, I would like to have at least one way to construct a composition without nesting.

xukai92:
I guess the way to link `composel` and `composer` to `foldr` and `foldr` introduces extra confusion. Let's forget about `foldr` and `foldr` and simply say `composel` means applying things from the left and `composer` means applying things from the right. How does this sound?

> If all the "standard" ways of constructing a Composed are such that we get nested compositions, it becomes difficult to do something like what I just mentioned.

I didn't mean to get nested compositions via `composel` and `composer`. Simply `composel(ts...) = Composed(ts)` and `composer(ts...) = Composed(reverse(ts))`.

torfjelde:
Ah, follow :+1: Yeah that sounds good!

src/interface.jl
```diff
@@ -0,0 +1,709 @@
...
+
+#############
+# Exp & Log #
+#############
+
+struct Exp <: Bijector end
+struct Log <: Bijector end
+
+(b::Log)(x) = @. log(x)
+(b::Exp)(y) = @. exp(y)
```
sharanry: We often need to apply this to just one of the dimensions. Any ideas on how to go about this?

torfjelde: Me and @xukai92 were discussing this earlier, and we feel like batch computation is something we'll have to have a proper discussion about after this PR is merged. For now assume batch-computation is _not_ supported :/ The difficulty is that in a lot of cases it's ambiguous whether or not something is a batch. And we want to make sure that if we say we support batch-computation, then it's supported for _all_ bijectors, not just on a case-by-case basis.

I know you use it a lot though, so I'd suggest making a different branch where you try to accommodate batch-computation and then we can potentially use this when we're considering the approach we're going to take to support it:)

Discussion: https://github.com/TuringLang/Bijectors.jl/pull/27#discussion_r317863600

xukai92: @sharanry @torfjelde Let's experiment this feature in a different branch. I'm going to merge this PR soon.

xukai92:
We also need to increment the version number in `Project.toml`. Since these are major interface introduction, shall we increment to 0.4.0 @yebai ?

yebai:
> We also need to increment the version number in Project.toml. Since these are major interface introduction, shall we increment to 0.4.0 @yebai ?

Sure, 0.4 sounds good to me.

xukai92:
@torfjelde Can you increment the version please?

torfjelde:
> @torfjelde Can you increment the version please?

Done!

PR #29, opened 2019-07-21
broadcast -> map

mohamed82008:
Tracker seems to fail when broadcasting with a `Bool`, ref https://github.com/FluxML/Tracker.jl/issues/42. In this PR, I replace all instances of broadcasting with map to fix some of the Turing tests. Some of those replacements may not be necessary but just to be on the safe side, I removed all broadcasting.


mohamed82008:
When tests pass, I will merge.

PR #33, opened 2019-07-22
Wishart perf and stability

mohamed82008:
This PR offsets the matrix when `logpdf_with_trans` for `Wishart` and `InverseWishart` fails. It also inlines the `logpdf` call to reuse the `Cholesky` for performance. I unrolled the loop version of the `logpdf_with_trans` since I found it faster when doing reverse_diff in Turing.

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -7,7 +7,6 @@ using LinearAlgebra
 using MappedArrays
 
 export  TransformDistribution,
-        RealDistribution,
         PositiveDistribution,
         UnitDistribution,
         SimplexDistribution,
@@ -350,7 +349,7 @@ function logpdf_with_trans(
     p = dim(d)
     Xcf = cholesky(X, check=false)
     if !issuccess(Xcf)
-        Xcf = cholesky(X + diagm(0 => fill(sqrt(eps(T)) * norm(X, Inf), size(X, 1))))
+        Xcf = cholesky(X + (eps(T) * norm(X)) * I)
     end
     lp = 0.5 * ((df - (p + 1)) * logdet(Xcf) - tr(d.S \ X)) - d.c0
     if transform && isfinite(lp)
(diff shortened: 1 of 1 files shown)


src/Bijectors.jl
```diff
@@ -335,16 +334,33 @@ function invlink(d::PDMatDistribution, Y::AbstractMatrix{<:Real})
...
+        Xcf = cholesky(X + (eps(T) * norm(X)) * I)
+    end
+    lp = getlogp(d, Xcf, X)
     if transform && isfinite(lp)
-        U = cholesky(X).U
-        lp += sum((dim(d) .- (1:dim(d)) .+ 2) .* log.(view(U, diagind(U))))
-        lp += dim(d) * log(2)
+        U = Xcf.U
+        @inbounds @simd for i in 1:dim(d)
+            lp += (dim(d) - i + 2) * log(U[i, i])
```
willtebbutt: Could we please see some timings for this?

mohamed82008:
```julia
using Distributions, LinearAlgebra, Bijectors, BenchmarkTools

function bench_tracker()
	d = Wishart(4, Matrix{Float64}(I, 4, 4))
	X = rand(d)
	@btime Bijectors.logpdf_with_trans($d, $X, true)
end

# No loop 
# 841.956 ns (11 allocations: 800 bytes)

# Loop
# 615.796 ns (6 allocations: 528 bytes)
```

mohamed82008: Sorry for the delay!

mohamed82008:
```julia
function bench_tracker()
	d = InverseWishart(4, Matrix{Float64}(I, 4, 4))
	X = rand(d)
	@btime Bijectors.logpdf_with_trans($d, $X, true)
end

# No loop
# 881.271 ns (12 allocations: 1008 bytes)

# Loop
# 648.804 ns (7 allocations: 736 bytes)
```

Issue #35, opened 2019-08-29
`Bijector` support for batch computation


torfjelde:
## Problem
Currently the `Bijector` interface doesn't "officially" support batch computations. With this I mean, what do you do when a `SimplexBijector` receives an `AbstractMatrix` rather than an `AbstractVector` as is expected? In some cases it's fairly straight forward to do so, e.g. the `SimplexBijector` only makes sense when applied to a vector so when we receive a matrix we know it's a batch and we can just do the right thing. In other cases what to do with a `Real` vs. `AbstractVector{<:Real}` is not so clear, e.g. `Exp` _elementwise_ is a well-defined bijection on `Real`, `AbstractVector`, etc. so what do we do in `logabsdetjac(b::Exp, x)` when `x isa AbstractArray`? Is `x` a batch of inputs so we should return `[y[i] for i = 1:length(x)]`, or is it intended as a vector and we should return `sum(y)` as is need for the distribution `MvLogNormal`?

To drive the point home, what do we do in this situation:
```julia
julia> struct Scale{T}
           a::T
       end

julia> (b::Scale)(x) = b.a .* x

julia> b = Scale(2.0)
Scale{Float64}(2.0)

julia> x = randn()
0.2414826706733002

julia> b(x)
0.4829653413466004

julia> logabsdetjac(b::Scale, x) = log(abs(b.a))
logabsdetjac (generic function with 1 method)

julia> logabsdetjac(b, x)
0.6931471805599453

julia> x = rand(2) # suppose we want to scale vectors elementwise instead
2-element Array{Float64,1}:
 0.7910530695591607 
 0.03333662406026572

julia> b(x) # transform still works nicely
2-element Array{Float64,1}:
 1.5821061391183213 
 0.06667324812053144

julia> logabsdetjac(b, x)
0.6931471805599453

julia> x = rand(2) # batch of 2
2-element Array{Float64,1}:
 0.546411899713563  
 0.16872866221657734

julia> b(x) # transform works on batch as expected
2-element Array{Float64,1}:
 1.092823799427126 
 0.3374573244331547

julia> logabsdetjac(b, x) # NO! Want this to return [logabsdetjac(b, x[1]), logabsdetjac(b, x[2])]!
0.6931471805599453
```
[...]

PR #36, opened 2019-08-30
`Stacked` bijector for stacking together bijectors

torfjelde:
This PR introduces `Stacked <: Bijector` which allows the user to, well, stack univariate and multivariate bijectors together into a single multivariate bijector. I've been using this a lot in a local branch of mine, both for experimentation with Bijectors.jl and improving VI in Turing.jl.

The concept of `Stacked` is fairly simple: the user doesn't have to write `[bijectors[i](x[ranges[i]]) for i = 1:length(ranges)]` to apply different bijectors across a vector where different parts of the vector require different bijectors.

But more than taking this tiny burden off the shoulders of the user, we can ensure that other computations for such a bijector is done correctly and efficiently, and that it works nicely with the rest of Bijectors.jl (e.g. composition). For example, the jacobian of such a `Stacked` is simply a block-diagonal matrix, where the i-th block is `jacobian(bijectors[i], x[ranges[i]])`. This also means that the `logabsdetjac(stacked, x)` is just `sum([logabsdetjac(bijectors[i], x[ranges[i]]) for i = ...])`. Moreover, we can also provide these implementations in a type-stable manner using `@generated` and whatnot.

Similarily to `Composed`, `Stacked` defaults to using a `Tuple` as the internal representation of the stacked bijectors, but allow using other types of containers.

I think the usefulness is most easily seen through some examples.

# Examples
## Simple
```julia
julia> using Bijectors; using Bijectors: Logit, Exp

julia> b1 = inv(Logit(0.0, 1.0))  # image is (0, 1)
Inversed{Logit{Float64}}(Logit{Float64}(0.0, 1.0))

julia> b2 = Bijectors.Exp()       # image is (0, ∞)
Exp()

julia> sb = vcat(b1, b2)  # or equivalently Stacked((b1, b2))
Stacked{Tuple{Inversed{Logit{Float64}},Exp},2}((Inversed{Logit{Float64}}(Logit{Float64}(0.0, 1.0)), Exp()), (1:1, 2:2))

julia> x = [-1.0, -3.0]
2-element Array{Float64,1}:
 -1.0
 -3.0

julia> y = sb(x)
2-element Array{Float64,1}:
 0.2689414213699951  
 0.049787068367863944

julia> (0 ≤ y[1] ≤ 1.0) && (0 < y[2])
true
```
## ADVI with `MvNormal` as base distribution
This `Stacked` allows us to do neat things like take a multivariate distribution as the base-distribution, which is then transformed using bijectors. This is very useful for a mean-field ADVI approach!
```julia
# Original distributions
dists = [
    Beta(),
    InverseGamma(),
    Dirichlet(2, 3)
]

# Construct the corresponding ranges
ranges = []
idx = 1
for i = 1:length(dists)
    d = dists[i]
    push!(ranges, idx:idx + length(d) - 1)

    global idx
    idx += length(d)
end

# Base distribution; mean-field normal
num_params = ranges[end][end]
d = MvNormal(zeros(num_params), ones(num_params))

# Construct the transform
bs = bijector.(dists)     # constrained-to-unconstrained bijectors for dists
ibs = inv.(bs)            # invert, so we get unconstrained-to-constrained
sb = Stacked(ibs, ranges) # => Stacked <: Bijector

# Mean-field normal with unconstrained-to-constrained stacked bijector
td = transformed(d, sb)
y = rand(td)

0.0 ≤ y[1] ≤ 1.0   # => true
0.0 < y[2]         # => true
sum(y[3:4]) ≈ 1.0  # => true
```
This can easily replace the hard-coded `elbo` in the current ADVI implementation in Turing. I have a local branch of Turing for which I have done this, and the resulting implementation is much more general. The only remaining part that needs to be addressed in that case is dealing with constraints on the parameters of the base distribution, e.g. ensuring variances of the underlying `MvNormal` are positive. When that is addressed (which can be done using the bijectors from Bijectors.jl!), we can just plug in arbitrary variational distributions to the `advi` call!

## Normalizing flows
With this `Stacked` bijector, it becomes trivial to create normalizing flows with constrained outputs!
```julia
using Bijectors, Turing, Tracker

# Simple model using Turing
@model gdemo_d() = begin
  s ~ InverseGamma(2, 3)
  m ~ Normal(0, sqrt(s))
  1.5 ~ Normal(m, sqrt(s))
  2.0 ~ Normal(m, sqrt(s))
  return s, m
end

# Computes the logdensity of a Turing.Model
logdensity(model, z) = begin
    varinfo = Turing.VarInfo(Turing.VarInfo(model), Turing.SampleFromUniform(), z)
    model(varinfo)
    return varinfo.logp
end

# Instantiate model
model = gdemo_d()

# Radial flow with an shift, composed with unconstrained-to-constrained transform
d = MvNormal(zeros(2), ones(2))
sb = vcat(bijector(InverseGamma(2, 3)), bijector(Normal()))
b = inv(sb) ∘ RadialLayer(2, param) ∘ Shift(param(zeros(2)))

# Sample from `d`, and transform using `b`
flow = transformed(d, b)

# Sample and transform from flow
z₀, zₖ, logjac, logpdf_q = forward(flow)

# Compute logdensity using the sampled parameters
lp = logdensity(model, zₖ)

# Since `q isa MvNormal` we have analytical expression for entropy
elbo = lp - logjac + entropy(q)

# Backprop the negative ELBO
Tracker.back!(- elbo, 1.0)

# Perform gradient descent using your favorite optimiser!
...
```
By using Flux's `treelike` and whatnot to traverse the Bijectors and their corresponding parameters, the update loop simply becomes
```julia
using Flux: Optimise
using Flux: treelike
@treelike Shift
@treelike RadialFlow
@treelike Stacked
@treelike TransformedDistribution

opt = Optimise.ADAGrad()

for p in Flux.params(flow)
    Δ = Tracker.grad(p)
    Optimise.update!(opt, p, Δ)
end
```
With the changes `Stacked` allows to ADVI in Turing, we can easily just plug the resulting posterior into `advi` to have NFVI + ADVI for (possibly) more flexible VI! It's all very exciting, isn't it?

# Things to address
- [x] Currently we provide the following different ways of constructing a `Stacked`. I think it's up for discussion whether we want the two "default" constructors `Stacked(bs)` and `vcat(bs...)`.
  - `Stacked(bs, ranges)` which is fairly straight forward
  - `Stacked(bs)` which assumes all the bijectors `bs` are _univariate_, i.e. the `ranges` in the original constructor is replaced by `[1:1, 2:2, 3:3, ...]`. Right now this makes it very easy for the user to do the _wrong_ thing, though in the transform performed by `Stacked` we have an `@assert size(x) == size(y)` where `x` is input and `y` is transformed to let the user know they did something wrong:) This can probably be caught upon construction when we've resolved https://github.com/TuringLang/Bijectors.jl/issues/35 in a good manner.
  - `vcat(bs::Bijector...) = Stacked(bs)`, suffers from same issues as `Stacked(bs)`.


src/interface.jl
```diff
@@ -123,7 +124,9 @@ end
 struct SingularJacobianException{B} <: Exception where {B<:Bijector}
     b::B
 end
-Base.showerror(io::IO, e::SingularJacobianException) = print(io, "jacobian of $(e.b) is singular")
+function Base.showerror(io::IO, e::SingularJacobianException)
+    print(io, "jacobian of $(e.b) is singular")
+end
```
xukai92: Is this for line character limit?

src/interface.jl
```diff
@@ -324,13 +401,14 @@ logabsdetjac(b::Scale, x) = log(abs(b.a))
 # Simplex bijector #
 ####################
 struct SimplexBijector{T} <: Bijector where {T} end
+SimplexBijector() = SimplexBijector{Val{true}}()
 
 const simplex_b = SimplexBijector{Val{false}}()
 const simplex_b_proj = SimplexBijector{Val{true}}()
```
xukai92: Is this still used?

src/interface.jl
```diff
@@ -725,3 +811,26 @@ In the case where `d isa Distribution`, this means
 """
 forward(d::Distribution) = forward(GLOBAL_RNG, d)
 forward(d::Distribution, num_samples::Int) = forward(GLOBAL_RNG, d, num_samples)
+
+# utility stuff
+params(td::Transformed) = params(td.dist)
```
xukai92: Is `Transformed` treelike?

src/interface.jl
```diff
@@ -240,6 +243,80 @@ function forward(cb::Composed, x)
...
+    y = _transform(x, sb.ranges, sb.bs...)
+
+    # TODO: maybe tell user to check their ranges?
+    @assert size(y) == size(x)
+
+    return y
+end
+(sb::Stacked)(x::AbstractMatrix{<: Real}) = hcat([sb(x[:, i]) for i = 1:size(x, 2)]...)
+function (sb::Stacked)(x::TrackedArray{A, 2}) where {A}
+    return Tracker.collect(hcat([sb(x[:, i]) for i = 1:size(x, 2)]...))
```
xukai92: Can you remind me of what does `Tracker.collect` do?

test/interface.jl
```diff
@@ -282,6 +282,94 @@ struct NonInvertibleBijector{AD} <: ADBijector{AD} end
         @test f_t == f_a
     end
 
+    @testset "Stacked <: Bijector" begin
```
xukai92: Can we add a simple test that do something like below?
```julia
x = ones(5)
sb = vcat(Exp(), Log(), ...) # stack some simple bijectors
@test sb(x) == [...] # some simple exact answers
```
```

File: src/interface.jl (hunk 1 above)
Comment (xukai92) [reply src/interface.jl]: If so I suggest to do the following
```julia
Base.showerror(io::IO, e::SingularJacobianException) = 
    print(io, "jacobian of $(e.b) is singular")

```

src/interface.jl
torfjelde: Basically convert `Array{TrackedReal}` into `TrackedArray`

src/interface.jl
torfjelde: Not sure? I thought it was nice to have due to this basically corresponding to the default `link` and `invlink`.

src/interface.jl
torfjelde: No, not yet at least. It's `params` from Distributions.jl but yeah it doesn't catch the case where the bijectors themselves are parameterized. It's not clear to me how to do this in general. There's a discussion regarding whether or not we want to depend on Flux over in Sharans PR on batch norm

src/interface.jl
torfjelde: Yup!
But does that follow the [style-guide](https://turing.ml/dev/docs/contributing/style-guide#method-definitions): "Only use short-form function definitions when they fit on a single line"?

src/interface.jl
torfjelde: Should maybe also include a `SimplexBijector(proj::bool) = SimplexBijector{Val{proj}}()`?

src/interface.jl
xukai92: Opps. I think it is me who is not following the style guide.

xukai92: Let's stick to the guide then. Thanks for pointing out this.

src/interface.jl
xukai92: Looks like a good constructor singature to me.

torfjelde:
Btw, what do people think about the "Things to address" part of the PR?

Currently you can construct a `Stacked` using `vcat`; but this will assume they're all 0-dimensional bijectors. In #44 were we introduce the dimensionality into the type, we can specialize `vcat(bs::Bijector{0}...)` which ensures the correct thing is done. We can also do this for `Stacked(bs::Bijector{0}...)`.

src/interface.jl
```diff
@@ -256,6 +259,80 @@ function forward(cb::Composed, x)
...
+end
+
+@generated function _logabsdetjac(
+    x,
+    rs::NTuple{N, UnitRange{Int}},
+    bs::Bijector...
+) where {N}
+    exprs = []
+    for i = 1:N
+        push!(exprs, :(sum(logabsdetjac(bs[$i], x[rs[$i]]))))
```
torfjelde: This `sum` here was to deal with the case when `rs[i] == 2:2` and `bs[i]` is a 0-dim bijector; in this case then the bijector is only guaranteed to be defined for a `Real` so we should ideally pass it `x[i]` if `length(rs[i]) == 1`. Does that make sense?

This is another issue (also the `vcat` one) solved by #44

xukai92:
> Btw, what do people think about the "Things to address" part of the PR?
> 
> Currently you can construct a `Stacked` using `vcat`; but this will assume they're all 0-dimensional bijectors. In #44 were we introduce the dimensionality into the type, we can specialize `vcat(bs::Bijector{0}...)` which ensures the correct thing is done. We can also do this for `Stacked(bs::Bijector{0}...)`.
> 
> But with current master it's very easy for the user to do the _wrong_ thing :/

It's OK to leave it as it as #44 will come soon.

Also, how about`vcat(bs::Bijector{1}...)`?

torfjelde:
> Also, how about `vcat(bs::Bijector{1}...)`?
The issue is that we can't just deduce the corresponding `ranges`. In the case where we have multivariate bijectors, the user needs to explicitly pass in the ranges using `Stacked(bs, ranges)`

xukai92:
I guess we can support `Stacked(bs, ranges)` and `Stacked(bs::StaticBijector{1}...)` then.

torfjelde:
> I guess we can support Stacked(bs, ranges) and Stacked(bs::StaticBijector{1}...) then.

Yup!:)

src/interface.jl
```diff
@@ -434,6 +434,7 @@ end
 
 function (b::SimplexBijector{Val{proj}})(x::AbstractVector{T}) where {T, proj}
     y, K = similar(x), length(x)
+    @assert K > 1
```
xukai92: Can you add assertion msgs as well?

test/interface.jl
xukai92: Is this done?

torfjelde: Yup: L339

test/interface.jl
```diff
@@ -282,6 +282,94 @@ struct NonInvertibleBijector{AD} <: ADBijector{AD} end
...
+            bs = bijector.(dists)     # constrained-to-unconstrained bijectors for dists
+            ibs = inv.(bs)            # invert, so we get unconstrained-to-constrained
+            sb = Stacked(ibs, ranges) # => Stacked <: Bijector
+            x = rand(d)
+            sb(x)
+            @test sb isa Stacked
+
+            td = transformed(d, sb)  # => MultivariateTransformed <: Distribution{Multivariate, Continuous}
+            @test td isa Distribution{Multivariate, Continuous}
+
```
xukai92: L339 looks like an emtpy line.

torfjelde: Yes, I've just added some whitespace to indicate a separation between the two tests, this and the one that "starts" at L341

src/interface.jl
torfjelde: I've also removed the `simplex_b` and `simplex_b_proj` as they are no longer used (Maybe this was actually what you were asking but for some reason Github showed the `SimplexBijector()` line as the one you were referring to)

xukai92:
24-hour merge count down.

xukai92:
In the meantime can you bump up the version number - I will make a new release after this PR is merged.

torfjelde:
I'd suggest one final change: rename `vcat` to `stack`. I've realized it wasn't such a good idea to overload `Base.vcat` since this means we can't actually construct arrays of bijectors using `vcat`. I initially got a bit carried away with making it "intuitive" to just stack bijectors :confused:

PR #37, opened 2019-08-30
[WIP] Updated README with info about new interface

torfjelde:
Feel free to have a look as I go and give feedback :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,3 +61,245 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
 julia> logpdf_with_trans(dist, x, true) # considering the transformation
 -0.7760422307471244
 ```
+
+## `Bijector` interface
+Other than the `logpdf_with_trans` methods, the package also provides a more composable interface through the `Bijector` types. Consider for example the from above with `Beta(2, 2)`.
+
+```julia
+julia> using Bijectors; using Bijectors: Logit
+
+julia> dist = Beta(2, 2)
+Beta{Float64}(α=2.0, β=2.0)
+
+julia> x = rand(dist)
+0.7173326646959575
+
+julia> b = bijector(dist) # bijection (0, 1) → ℝ
+Logit{Float64}(0.0, 1.0)
+
+julia> y = b(x)
+0.9312689879144197
+```
+
+In this case we see that `bijector(d::Distribution)` returns the corresponding constrained-to-unconstrained bijection for `Beta`, which indeed is a `Logit` with `a = 0.0` and `b = 1.0`. The resulting `Logit <: Bijector` has a method `(b::Logit)(x)` defined, allowing us to call it just like any other function. Comparing with the above example, `b(x) == link(dist, x)`. Just to convince ourselves:
+
+```julia
+julia> b(x) == link(dist, x)
+true
+```
+
+What about `invlink`?
+
+```julia
+julia> b⁻¹ = inv(b)
+Inversed{Logit{Float64}}(Logit{Float64}(0.0, 1.0))
+
+julia> b⁻¹(y)
+0.7173326646959575
+
+julia> b⁻¹(y) == invlink(dist, y)
+true
+```
+
+Pretty neat, huh? `Inversed{Logit}` is also a `Bijector` where we've defined `(ib::Inversed{<:Logit})(y)` as the inverse transformation of `(b::Logit)(x)`. Note that it's not always the case that `inv(b) isa Inversed`, e.g. the inverse of `Exp` is simply `Log` so `inv(Exp()) isa Log` is true. Aslo, we can _compose_ bijectors:
+
+```julia
+julia> id = (b ∘ b⁻¹)
+Composed{Tuple{Inversed{Logit{Float64}},Logit{Float64}}}((Inversed{Logit{Float64}}(Logit{Float64}(0.0, 1.0)), Logit{Float64}(0.0, 1.0)))
+
+julia> id(y) == y
+true
+```
+
+This far we've seen that we can replicate the functionality provided by `link` and `invlink`. To replicate `logpdf_with_trans` we instead provide a `TransformedDistribution <: Distribution` implementing the `Distribution` interface from Distributions.jl:
+
+```julia
+julia> using Bijectors: TransformedDistribution
+
+julia> td = transformed(dist)
+TransformedDistribution{Beta{Float64},Logit{Float64},Univariate}(
+dist: Beta{Float64}(α=2.0, β=2.0)
+transform: Logit{Float64}(0.0, 1.0)
+)
+
+julia> td isa UnivariateDistribution
+true
+
+julia> logpdf(td, y)
+-1.0577727579778098
+
+julia> logpdf_with_trans(dist, x, true)
+-1.05777275797781
+```
+
+When computing `logpdf(td, y)` where `td` is the _transformed_ distribution corresponding to `Beta(2, 2)`, it makes more semantic sense to compute the pdf of the _transformed_ variable `y` rather than using the "un-transformed" variable `x` to do so, as we do in `logpdf_with_trans`. With that being said, we can also do
+
+```julia
+julia> logpdf_forward(td, x)
+-1.05777275797781
+```
+
+At this point we've only shown that we can replicate the existing functionality. But we said `TransformedDistribution isa Distribution`, so we also have `rand`:
+
+```julia
+julia> y = rand(td)              # ∈ ℝ
+-0.5231573469209508
+
+julia> x = inv(td.transform)(y)  # transform back to interval [0, 1]
+0.37211423725902915
+```
+
+This can be quite convenient if you have computations assuming input to be on the real line.
+
+But the real utility of `TransformedDistribution` becomes more apparent when using `transformed(dist, b)` for any bijector `b`. To get the transformed distribution corresponding to the `Beta(2, 2)`, we called `transformed(dist)` before. This is simply an alias for `transformed(dist, bijector(dist))`. Remember `bijector(dist)` returns the constrained-to-constrained bijector for that particular `Distribution`. But we can of course construct a `TransformedDistribution` using different bijectors for the same distribution! It's particular useful in something called _Automatic Derivative Variational Inference (ADVI)_.(INSERT REF) An important part of this to approximate a constrained distribution, e.g. `Beta`, as follows:
+1. Sample `x` from a `Normal` with parameters `μ` and `σ`, i.e. `x ~ Normal(μ, σ)`.
+2. Transform `x` to `y` s.t. `y ∈ support(Beta)`, with the transform being a differentiable bijection with a differentiable inverse (a "bijector")
+This then defines a probability density with support same as `Beta`! Of course, it's unlikely that it will be the same, but it's an _approximation_. Creating such a distribution becomes trivial with `Bijector` and `TransformedDistribution`:
+
+```julia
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,3 +61,294 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+```julia
+julia> using Bijectors; using Bijectors: Logit
+
+julia> dist = Beta(2, 2)
+Beta{Float64}(α=2.0, β=2.0)
+
+julia> x = rand(dist)
+0.7173326646959575
+
+julia> b = bijector(dist) # bijection (0, 1) → ℝ
```
yebai: `bijector` ==> `Bijector`?

README.md
```diff
@@ -61,3 +61,294 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
 julia> logpdf_with_trans(dist, x, true) # considering the transformation
 -0.7760422307471244
 ```
+
+## `Bijector` interface
+Other than the `logpdf_with_trans` methods, the package also provides a more composable interface through the `Bijector` types. Consider for example the from above with `Beta(2, 2)`.
```
yebai: "the from above" ==> "above"?

README.md
```diff
@@ -61,3 +61,294 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+julia> td = transformed(dist)
+TransformedDistribution{Beta{Float64},Logit{Float64},Univariate}(
+dist: Beta{Float64}(α=2.0, β=2.0)
+transform: Logit{Float64}(0.0, 1.0)
+)
+
+julia> td isa UnivariateDistribution
+true
+
+julia> logpdf(td, y)
```
yebai: Nice - we can probably replace `logpdf_with_trans` with `logpdf(td::TransformedDistribution, y)` in Turing.

README.md
torfjelde: Is supposed to be "the one from above"; makes more sense?

yebai: ah, that makes a lot more sense!

README.md
torfjelde: Should be possible, thought might require some work? I'm not too familiar with exactly how the transformations happen in Turing, but just had a quick look and was not quite clear to me how to proceed in an efficient way. Using this interface I think the best way would be if we could somehow replace the original distributions with `TransformedDistribution` when we want to perform transformations, and then in `assume` return `inv(transformed_dist.transform)(y), logpdf(transformed_dist, y)`. As long as everything is type-stable, this should be efficient.

I've made an effort to make things in this package type-stable, but there's one `bijector(d::Distribution)` call that is currently _not_ type-stable: `Truncated`. See Issue #38. I wouldn't think this would pose much of an issue for application in Turing.jl since the construction of the distribution will occur in the very first construction of `VarInfo`, whose construction is already unstable.

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+Inversed{Logit{Float64}}(Logit{Float64}(0.0, 1.0))
+
+julia> b⁻¹(y)
+0.7173326646959575
+
+julia> b⁻¹(y) == invlink(dist, y)
+true
+```
+
+Pretty neat, huh? `Inversed{Logit}` is also a `Bijector` where we've defined `(ib::Inversed{<:Logit})(y)` as the inverse transformation of `(b::Logit)(x)`. Note that it's not always the case that `inv(b) isa Inversed`, e.g. the inverse of `Exp` is simply `Log` so `inv(Exp()) isa Log` is true. Aslo, we can _compose_ bijectors:
```
willtebbutt: Typo: `Aslo` -> `Also`

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+```
+
+When computing `logpdf(td, y)` where `td` is the _transformed_ distribution corresponding to `Beta(2, 2)`, it makes more semantic sense to compute the pdf of the _transformed_ variable `y` rather than using the "un-transformed" variable `x` to do so, as we do in `logpdf_with_trans`. With that being said, we can also do
+
+```julia
+julia> logpdf_forward(td, x)
+-1.05777275797781
+```
+
+In the computatio of both `logpdf` and `logpdf_forward` we need to compute `log(abs(det(jacobian(inv(b), y))))` and `log(abs(det(jacobian(b, x))))`, respectively. This computation is available using the `logabsdetjac` method
```
willtebbutt: Typo: compuatio -> computation

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+julia> y = rand(td)              # ∈ ℝ
+-0.5231573469209508
+
+julia> x = inv(td.transform)(y)  # transform back to interval [0, 1]
+0.37211423725902915
+```
+
+This can be quite convenient if you have computations assuming input to be on the real line.
+
+But the real utility of `TransformedDistribution` becomes more apparent when using `transformed(dist, b)` for any bijector `b`. To get the transformed distribution corresponding to the `Beta(2, 2)`, we called `transformed(dist)` before. This is simply an alias for `transformed(dist, bijector(dist))`. Remember `bijector(dist)` returns the constrained-to-constrained bijector for that particular `Distribution`. But we can of course construct a `TransformedDistribution` using different bijectors with the same `dist`. This is particularly useful in something called _Automatic Derivative Variational Inference (ADVI)_.[2] An important part of ADVI is to approximate a constrained distribution, e.g. `Beta`, as follows:
```
willtebbutt: Typo: Derivative -> Differentiation

willtebbutt:
I've left a few comments on typos that you probably would have picked up anyway, but really nice work. It's nice to have some decent documentation.

README.md
cpfiffer: I think it's a good goal to have. One of the things I've had a lot of trouble with while working on Turing's sampler internals was when to link and when not to link. If we just had a bunch of `VarName => TransformedDistribution` pairs, we could theoretically handle all the gradients and everything automatically without anyone ever having to explicitly call `link!(vi, spl)` and `invlink!(vi, spl)`. We could also make sure that it's done in a performant way, so that transformations only occur when they absolutely need to.

torfjelde: Oh yeah, I'm all for making the transition; I just wanted to say that it's probably not going to be as easy as _I_ thought at first:) You guys might have already been aware that's not as straight forward.

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+Inversed{Logit{Float64}}(Logit{Float64}(0.0, 1.0))
+
+julia> b⁻¹(y)
+0.7173326646959575
+
+julia> b⁻¹(y) == invlink(dist, y)
+true
+```
+
+Pretty neat, huh? `Inversed{Logit}` is also a `Bijector` where we've defined `(ib::Inversed{<:Logit})(y)` as the inverse transformation of `(b::Logit)(x)`. Note that it's not always the case that `inv(b) isa Inversed`, e.g. the inverse of `Exp` is simply `Log` so `inv(Exp()) isa Log` is true. Also, we can _compose_ bijectors:
```
xukai92: Add a line break before `Also, we can _compose_ bijectors:`

Also consider add subtitles like "Basic usage", "Composition", etc.

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+-1.595700144883034
+
+julia> logabsdetjac(b, x)
+1.595700144883034
+```
+
+Notice that
+
+```julia
+julia> logabsdetjac(b, x) ≈ - logabsdetjac(b⁻¹, y)
```
xukai92: Maybe `- logabsdetjac(b⁻¹, y)` -> `-logabsdetjac(b⁻¹, y)`, i.e. removing the space.

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+
+Another useful function is the `forward(d::Distribution)` method. It is similar to `forward(b::Bijector)` in the sense that it does a forward pass of the entire process "sample then transform" and returns all the most useful quantities in process using the most efficent computation path.
+
+```julia
+julia> x, y, logjac, logpdf_y = forward(flow) # sample + transform and returns all the useful quantities in one pass
+(x = [-0.387191, 0.761807], y = [-0.677683, 0.0866711] (tracked), logabsdetjac = -0.07475475048737289 (tracked), logpdf = -2.1282560611425447 (tracked))
+```
+
+This method is for example useful when computing quantities such as the _expected lower bound (ELBO)_ between this transformed distribution and some other joint density. If no analytical expression is available, we have to approximate the ELBO by a monte carlo estimate. But one term in the ELBO is the entropy of the base density, which we _do_ know analytically in this case. Using the analytical expression for the entropy and then using a monte carlo estimate for the rest of the terms in the ELBO gives an estimate with lower variance than if we used the monte carlo estimate for the entire expectation.
```
xukai92: `monte carlo` -> `Monte Carlo`

README.md
```diff
@@ -61,3 +61,492 @@ julia> logpdf_with_trans(dist, x, false) # ignoring the transformation
...
+
+For `TransformedDistribution`, together with default implementations for `Distribution`, we have the following methods:
+- `bijector(d::Distribution)`: returns the default constrained-to-unconstrained bijector for `d`
+- `transformed(d::Distribution)`, `transformed(d::Distribution, b::Bijector)`: constructs a `TransformedDistribution` from `d` and `b`.
+- `logpdf_forward(d::Distribution, x)`, `logpdf_forward(d::Distribution, x, logjac)`: computes the `logpdf(td, td.transform(x))` using the forward pass, which is potentially faster depending on the transform at hand.
+- `forward(d::Distribution)`: returns `(x = rand(dist), y = b(x), logabsdetjac = logabsdetjac(b, x), logpdf = logpdf_forward(td, x))` where `b = td.transform`. This combines sampling from base distribution and transforming into one function. The intention is that this entire process should be performed in the most efficient manner, e.g. the `logabsdetjac(b, x)` call might instead be implemented as `- logabsdetjac(inv(b), b(x))` depending on which is most efficient.
+
+# Bibliography
```
xukai92: References seem to be broken?

torfjelde: Uhmm, I've just implemented "manual" references... Can you do proper references in a Github markdown file? E.g. similar to BibTex `\cite`?

README.md
torfjelde: Added a bit more structure / sections; I agree, easier to get an overview now :+1:

README.md
xukai92: You will need to do it manually or comiple the markdown from `.tex`, which is a bit more involved.

xukai92:
What's left to do on this PR?

torfjelde:
> What's left to do on this PR?

Well, if people are happy with the current README, not really anything, just have to remove a TODO. The only thing is that `Stacked` is documented, but this can be included in the PR for `Stacked` instead.

README.md
xukai92: Can you fix this manually before we merge?

torfjelde: Sorry, what's wrong with it currently? That it's not a link?

xukai92: Not sure what are those empty braces `()`? Can you try to make the style consistent to the one in https://github.com/TuringLang/AdvancedHMC.jl, where we also provide hyperlinks to the arXiv version.

PR #39, opened 2019-08-31
Improves type-stability of `bijector(d::Distribution)`

torfjelde:
Related to #40. Also removes the use of a couple of, now, unnecessary `eval` calls. 

Before:
```julia
julia> for d in uni_dists
           c = @code_typed bijector(d)
           if (c.second isa Union) || (c.second == Any)
               println("-----------------")
               println(d, ": ", c.second)
               println("(", minimum(d), ", ", maximum(d), ")")
               println(bijector(d))
           end
       end
-----------------
Arcsine{Float64}(a=2.0, b=4.0): Union{Identity, Logit{Float64}, Composed}
(2.0, 4.0)
Bijectors.Logit{Float64}(2.0, 4.0)
-----------------
Biweight{Float64}(μ=0.0, σ=1.0): Union{Identity, Logit{Float64}, Composed}
(-1.0, 1.0)
Bijectors.Logit{Float64}(-1.0, 1.0)
-----------------
Cosine{Float64}(μ=0.0, σ=1.0): Union{Identity, Logit{Float64}, Composed}
(-1.0, 1.0)
Bijectors.Logit{Float64}(-1.0, 1.0)
-----------------
Epanechnikov{Float64}(μ=0.0, σ=1.0): Union{Identity, Logit{Float64}, Composed}
(-1.0, 1.0)
Bijectors.Logit{Float64}(-1.0, 1.0)
-----------------
Levy{Float64}(μ=0.0, σ=1.0): Union{Identity, Composed{Tuple{Shift{Float64},Log}}}
(0.0, Inf)
Composed{Tuple{Bijectors.Shift{Float64},Bijectors.Log}}((Bijectors.Shift{Float64}(-0.0), Bijectors.Log()))
-----------------
Pareto{Float64}(α=1.0, θ=1.0): Union{Identity, Composed{Tuple{Shift{Float64},Log}}}
(1.0, Inf)
Composed{Tuple{Bijectors.Shift{Float64},Bijectors.Log}}((Bijectors.Shift{Float64}(-1.0), Bijectors.Log()))
-----------------
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(-Inf, 2.0)): Union{Identity, Logit{Float64}, Composed}
(-Inf, 2.0)
Composed{Tuple{Bijectors.Scale{Float64},Composed{Tuple{Bijectors.Shift{Float64},Bijectors.Log}}}}((Bijectors.Scale{Float64}(-1.0), Composed{Tuple{Bijectors.Shift{Float64},Bijectors.Log}}((Bijectors.Shift{Float64}(2.0), Bijectors.Log()))))
```

After:
```julia
julia> for d in uni_dists
           c = @code_typed bijector(d)
           if (c.second isa Union) || (c.second == Any)
               println("-----------------")
               println(d, ": ", c.second)
               println("(", minimum(d), ", ", maximum(d), ")")
               println(bijector(d))
           end
       end
-----------------
Truncated(Normal{Float64}(μ=0.0, σ=1.0), range=(-Inf, 2.0)): Union{Identity, Logit{Float64}, Composed}
(-Inf, 2.0)
Composed{Tuple{Bijectors.Scale{Float64},Composed{Tuple{Bijectors.Shift{Float64},Bijectors.Log}}}}((Bijectors.Scale{Float64}(-1.0), Composed{Tuple{Bijectors.Shift{Float64},Bijectors.Log}}((Bijectors.Shift{Float64}(2.0), Bijectors.Log()))))
```

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -476,10 +476,6 @@ end
 (b::DistributionBijector)(x) = link(b.dist, x)
 (ib::Inversed{<:DistributionBijector})(y) = invlink(ib.orig.dist, y)
 
-
-"Returns the constrained-to-unconstrained bijector for distribution `d`."
-bijector(d::Distribution) = DistributionBijector(d)
-
 # Transformed distributions
 struct TransformedDistribution{D, B, V} <: Distribution{V, Continuous} where {D<:Distribution{V, Continuous}, B<:Bijector}
     dist::D
@@ -514,6 +510,7 @@ transformed(d) = transformed(d, bijector(d))
 
 Returns the constrained-to-unconstrained bijector for distribution `d`.
 """
+bijector(d::Distribution) = DistributionBijector(d)
 bijector(d::Normal) = IdentityBijector
 bijector(d::MvNormal) = IdentityBijector
 bijector(d::PositiveDistribution) = Log()
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -521,7 +521,9 @@ bijector(d::MvLogNormal) = Log()
 bijector(d::SimplexDistribution) = simplex_b_proj
 bijector(d::KSOneSided) = Logit(zero(eltype(d)), zero(eltype(d)))
 
-const BoundedDistribution = Union{Arcsine, Biweight, Cosine, Epanechnikov, Beta, NoncentralBeta}
+const BoundedDistribution = Union{
+    Arcsine, Biweight, Cosine, Epanechnikov, Beta, NoncentralBeta
+}
 bijector(d::BoundedDistribution) = Logit(minimum(d), maximum(d))
 
 const LowerboundedDistribution = Union{Pareto, Levy}
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -516,15 +516,11 @@ bijector(d::MvNormal) = IdentityBijector
 bijector(d::PositiveDistribution) = Log()
 bijector(d::MvLogNormal) = Log()
 bijector(d::SimplexDistribution) = simplex_b_proj
-bijector(d::KSOneSided) = Logit(zero(eltype(d)), zero(eltype(d)))
+bijector(d::KSOneSided) = Logit(zero(eltype(d)), one(eltype(d)))
 
-const BoundedDistribution = Union{
-    Arcsine, Biweight, Cosine, Epanechnikov, Beta, NoncentralBeta
-}
-bijector(d::BoundedDistribution) = Logit(minimum(d), maximum(d))
-
-const LowerboundedDistribution = Union{Pareto, Levy}
-bijector(d::LowerboundedDistribution) = Log() ∘ Shift(- minimum(d))
+bijector_bounded(d, a=minimum(d), b=maximum(d)) = Logit(a, b)
+bijector_lowerbounded(d, a=minimum(d)) = Log() ∘ Shift(-a)
+bijector_upperbounded(d, b=maximum(d)) = Log() ∘ Shift(b) ∘ Scale(- one(typeof(b)))
 
 # FIXME: (TOR) Can we make this type-stable?
 # Can also make a `TruncatedBijector`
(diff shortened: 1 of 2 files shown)


src/interface.jl
```diff
@@ -514,31 +510,27 @@ transformed(d) = transformed(d, bijector(d))
...
-    # Skipping KSOneSided because it's not a parametric type
-    if D == KSOneSided
-        continue
-    end
-    @eval bijector(d::$D{T}) where {T<:Real} = Logit(zero(T), one(T))
-end
+const LowerboundedDistribution = Union{Pareto, Levy}
+bijector(d::LowerboundedDistribution) = Log() ∘ Shift(- minimum(d))
 
 # FIXME: (TOR) Can we make this type-stable?
```
xukai92: Do we need this comment still?

xukai92:
Is it possible to solve the `Truncated` one as well? If not why not?

torfjelde:
> Is it possible to solve the Truncated one as well? If not why not?

Unfortunately not. Whether or not the truncation is bounded above and/or below is only known at run-time. Alternatively, we can fall back to using the AD-based `DistributionBijector` which uses `link` and `invlink` with `logabsdetjac` computed by AD. This would be type-stable, but not as nice as manual implementation.

src/interface.jl
torfjelde: I've left it there for `Truncated`; I'd say leave it so that in the case someone has an idea OR for some reason whether or not a distribution is bounded becomes available at compile-time, this is a reminder to revisit the problem:)

xukai92: I see.

xukai92:
I see. Maybe we can decompose the function into multiple and users at least have a choice to use the type stable version.

```julia
bijector_bounded(d, a=minimum(d), b=maximum(d)) = Logit(a, b)
bijector_lowerbounded(d, a=minimum(d)) = Log() ∘ Shift(-a)
bijector_upperbounded(d, b=maximum(d)) = Log() ∘ Shift(b) ∘ Scale(- one(typeof(b)))

function bijector(d::Truncated)
    a, b = minimum(d), maximum(d)
    lowerbounded, upperbounded = isfinite(a), isfinite(b)
    if lowerbounded && upperbounded
        return bijector_bounded(d)
    elseif lowerbounded
        return bijector_lowerbounded(d)
    elseif upperbounded
        return bijector_upperbounded(d)
    else
        return IdentityBijector
    end
end

# Also we can reuse them for others
bijector(d::BoundedDistribution) = bijector_bounded(d)
bijector(d::LowerboundedDistribution) = bijector_lowerbounded(d)
```

torfjelde:
> I see. Maybe we can decompose the function into multiple and users at least have a choice to use the type stable version.

Yeah I don't see a problem with this :+1: I'll add that tomorrow

torfjelde:
Also, another solution is to create a "runtime"-bounded `Bijector`. Something like `BoundedBijector` or something, with
```julia
struct BoundedBijector{T} <: Bijector
    a::T
    b::T
end

(b::BoundedBijector)(x) = ... # <= same as `link` for `Truncated`
(ib::Inversed{<:BoundedBijector})(y) = ... # <= same as `invlink` for `Truncated`
```

So it becomes a "use it if you have to, but if you know at compile-time what the boundaries are; use something else"?

torfjelde:
Also, due to the fact that there are 3 possible execution paths for `Truncated` with the current implementation, it might not be an issue in most cases due to [union-splitting](https://julialang.org/blog/2018/08/union-splitting).

xukai92:
Maybe actually impl a typed version?

```julia
struct BoundedBijector{T,L,U} <: Bijector end

BoundedBijector(a::T, b::T) where {T} = BoundedBijector{T, Val{a}, Val{b}}

# Resolve dispath based on L and U, e.g.
(b::BoundedBijector{T,Val{:Inf},U})(x) where {T,U} = bijector_upperbounded(d)
```

torfjelde:
> Maybe actually impl a typed version?

But this doesn't solve any thing I think. If you do `BoundedBijector(minimum(d), maximum(d))` inside a function you're going straight into not-type-inferrable territory, right?

I think as soon as we start mixing runtime-information (bounds) with compile-time information we're going to have a bad time :/

xukai92:
In the case where the user define it ahead and keeps using the same one it would be stable right? I believe it would be possible to define it only once in the compile time / first run time for a Turing model.

torfjelde:
Yeah, but we already have that covered through `bijector(d)` for the distributions which we know the `ub` and `lb`. Basically, we'd just be doing
```julia
BoundedBijector{T,Val{:Inf},U}(a) = Exp() ∘ Shift(-a)
```
and so on.

xukai92:
I feel having the interface for `BoundedBijector` is more general and we should fall back those for distributions to `BoundedBijector` version.

xukai92:
Is this ready to merge?

torfjelde:
Yup!

PR #43, opened 2019-09-12
Add BatchNorm flow to enable deeply stacked flows to perform adequately

sharanry:
A slightly modified version of @xukai92's https://github.com/xukai92/Flows.jl/blob/master/src/norm.jl

TODO:

- [x] Add tests
- [x] resolve logabsdetjacob issue

#### Naming
Should it be named `InvertibleBatchNorm` as named in @xukai92's work or just `BatchNorm`? Being a part of `Bijectors.jl`, I feel invertibility is implied.

#### logabsdetjacob test fails
While trying to compare the our implementation of `logabsdetjacob` with the estimated one, I found that they don't match
```julia
using Bijectors, ForwardDiff, LinearAlgebra, Test
z = randn(2, 20)
flow = Bijectors.BatchNorm(2)
forward_diff = log(abs(det(ForwardDiff.jacobian(t -> flow(t), z))))
our_method = sum(forward(flow, z).logabsdetjacob)

flow.active=false
@test inv(flow)(flow(z)) ≈ z rtol=0.2 # passes
@test (inv(flow) ∘ flow)(z) ≈ z rtol=0.2 # passes
@test our_method ≈ forward_diff # fails
```
```julia
Test Failed at In[7]:10
  Expression: our_method ≈ forward_diff
   Evaluated: -3.8534519385617116 ≈ -100.3214336037399
There was an error during testing
```


src/Bijectors.jl
```diff
@@ -6,6 +6,7 @@ using StatsFuns
 using LinearAlgebra
 using MappedArrays
 using Roots
+using Flux
```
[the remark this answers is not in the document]
xukai92: I also prefer to avoid this.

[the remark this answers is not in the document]
MikeInnes: I think there's another issue around that discusses factoring out `treelike` into a package. That seems like a better route forward. For now, if you need it you could just copy-paste.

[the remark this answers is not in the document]
sharanry: @xukai92 @willtebbutt @torfjelde So do you suggest I copy-paste `treelike` for now like @MikeInnes suggested?

src/interface.jl
```diff
@@ -368,6 +368,8 @@ struct Shift{T} <: Bijector
     a::T
 end
 
+Shift(dims::Int, container=Array) = Shift(container(zeros(dims, 1)))
```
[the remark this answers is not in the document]
sharanry: Yeah sorry for that!

src/Bijectors.jl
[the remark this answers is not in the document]
xukai92: How do we currently extract parameters of other flow transformations when using them?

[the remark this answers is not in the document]
torfjelde: We don't:) I've just extracted params by doing using `@treelike` on the structures involved and initializing the bijectors with `Tracker.param`. 

I'm not opposed to using `@treelike` for all bijectors if we copy-paste until it's separated into its own package. I would love to have something that also supports ForwardDiff.jl nicely but not clear how we'd do that.

sharanry:
Flux already has [`BatchNorm`](https://github.com/FluxML/Flux.jl/blob/29eae312b8a9a6bec48a0c71efc58f94c1d018eb/src/layers/normalise.jl#L122). Should we consider renaming this to something else to avoid confusion? Say `BatchNormFlow`.

src/interface.jl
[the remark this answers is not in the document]
sharanry: I have started from a clean branch now.

yebai:
@torfjelde @sharanry anything still missing here? If not, can we review the PR and get it merged?

sharanry:
I think it is ready for review. @xukai92 @torfjelde could you please review this?

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+end
+
+logabsdetjacob(
+    t::T,
+    x;
+    σ²=reshape(t.σ², affinesize(x)...)
+) where {T<:BatchNormFlow} =  (sum(t.logγ - log.(σ² .+ t.ϵ) / 2)) .* typeof(Flux.data(x))(ones(Float32, size(x, 2))')
+
+function _transform(t::BatchNormFlow, x)
+     @assert size(x, ndims(x) - 1) == length(t.μ) "`BatchNormFlow` expected $(length(t.μ)) channels, got $(size(x, ndims(x) - 1))"
```
torfjelde: Indentation level

torfjelde: Also line-length needs to be less than < 92 chars

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+        ϵ = t.ϵ
+    else
+        Tx = eltype(x)
+        dims = length(size(x))
+        axes = [1:dims-2; dims] # axes to reduce along (all but channels axis)
+        μ = mean(x, dims=axes)
+        σ² = sum((x .- μ) .^ 2, dims=axes) ./ m
+        ϵ = Flux.data(convert(Tx, t.ϵ))
+        # Update moving mean/std
+        mtm = Flux.data(convert(Tx, t.momentum))
```
torfjelde: Would it be better to `convert(Tracker.data(Tx), t.ϵ)` instead? Similiarly for second line

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+
+    x̂ = (x .- μ) ./ sqrt.(σ² .+ ϵ)
+    return (rv=γ .* x̂ .+ β, σ²=σ²)
+end
+
+(b::BatchNormFlow)(z) = _transform(b, z).rv
+
+function _forward(t::BatchNormFlow, x)
+    rv, σ² = _transform(t, x)
+    return (rv=rv, logabsdetjacob=logabsdetjacob(t, x; σ²=σ²))
```
torfjelde: `logabsdetjacob` should be `logabsdetjac`

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+end
+
+forward(flow::BatchNormFlow, z) = _forward(flow, z)
+
+# TODO: make this function take kw argument `σ²`
+logabsdetjacob(it::Inversed{T}, y) where {T<:BatchNormFlow} = (xsimilar = y; -logabsdetjacob(inv(it), xsimilar))
```
torfjelde: `logabsdetjacob` should be `logabsdetjac`

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+    @assert t.active == false "`forward(::Inversed{BatchNormFlow})` is only available in test mode but not in training mode."
+    as = affinesize(y)
+    γ = exp.(reshape(t.logγ, as...))
+    β = reshape(t.β, as...)
+    μ = reshape(t.μ, as...)
+    σ² = reshape(t.σ², as...)
+
+    ŷ = (y .- β) ./ γ
+    x = ŷ .* sqrt.(σ² .+ t.ϵ) .+ μ
+    return (rv=x, logabsdetjacob=logabsdetjacob(it, x))
```
torfjelde: `logabsdetjacob` should be `logabsdetjac`

test/norm.jl
```diff
@@ -0,0 +1,16 @@
+using Test
+using Bijectors, ForwardDiff, LinearAlgebra
+using Random: seed!
+
+seed!(1)
+
+@testset "BatchNormFlow" begin
+    z = randn(2, 20)
+    flow = Bijectors.BatchNormFlow(2)
+    flow.active=false
```
torfjelde: `=` → ` = `

src/Bijectors.jl
```diff
@@ -448,4 +449,6 @@ include("interface.jl")
 
 include("norm_flows.jl")
 
+include("norm.jl")
```
torfjelde: Not sure about this filename :/ But I'm very much in favour of restructuring a bit of the project, e.g. `src/bijectors/` is a directory containing all bijectors, and then have `batch_norm.jl` be in this. So maybe we can delay the file-renaming until we do the full restructuring

src/norm.jl
```diff
@@ -0,0 +1,92 @@
+# Code adapted from Flux.jl:
+# https://github.com/FluxML/Flux.jl/blob/68ba6e4e2fa4b86e2fef8dc6d0a5d795428a6fac/src/layers/normalise.jl#L117-L206
+# License: https://github.com/FluxML/Flux.jl/blob/master/LICENSE.md
+
+mutable struct BatchNormFlow{T1,T2} <: Bijector
+    β::T1
+    logγ::T1
+    μ  # moving mean
+    σ² # moving st
```
torfjelde: Untyped? Is this on purpose? This is not the case in the Flux implementation of the batch-norm.

src/norm.jl
```diff
@@ -0,0 +1,92 @@
+# Code adapted from Flux.jl:
+# https://github.com/FluxML/Flux.jl/blob/68ba6e4e2fa4b86e2fef8dc6d0a5d795428a6fac/src/layers/normalise.jl#L117-L206
+# License: https://github.com/FluxML/Flux.jl/blob/master/LICENSE.md
+
+mutable struct BatchNormFlow{T1,T2} <: Bijector
```
torfjelde: Could you add a docstring describing what this does?

Some of the bijectors have this already, most of which don't are self-explanatory (though I'm working on adding docstrings to those too).

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+    β::T1
+    logγ::T1
+    μ  # moving mean
+    σ² # moving st
+    ϵ::T2
+    momentum::T2
+    active::Bool # true when training
+end
+
+BatchNormFlow(dims::Int, container=Array; ϵ=1f-5, momentum=0.1f0) = BatchNormFlow(
```
torfjelde: Is the `container=Array` purposefully not part of the keyword arguments here or should it be after `;`?

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+
+function affinesize(x)
+    dims = length(size(x))
+    channels = size(x, dims-1)
+    affinesize = ones(Int, dims)
+    affinesize[end-1] = channels
+    return affinesize
+end
+
+logabsdetjacob(
```
torfjelde: `logabsdetjacob` should be `logabsdetjac`

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+    t::T,
+    x;
+    σ²=reshape(t.σ², affinesize(x)...)
+) where {T<:BatchNormFlow} =  (sum(t.logγ - log.(σ² .+ t.ϵ) / 2)) .* typeof(Flux.data(x))(ones(Float32, size(x, 2))')
```
torfjelde: Does all of this work nicely with `Tracker`? I.e. we're not getting cases where `TrackedArray` is converted into `Array{<:TrackedReal}`.

src/norm.jl
```diff
@@ -0,0 +1,92 @@
...
+    channels = size(x, dims-1)
+    affinesize = ones(Int, dims)
+    affinesize[end-1] = channels
+    return affinesize
+end
+
+logabsdetjacob(
+    t::T,
+    x;
+    σ²=reshape(t.σ², affinesize(x)...)
```
torfjelde: Remove this. We need the `logabsdetjac` implementations to stick to the `logabsdetjac(b, x)` interface. Otherwise things will break down as soon as we start using compositions.

src/norm.jl
torfjelde: As mentioned above, `logabsdetjacob(t, x; σ²=σ²)` shouldn't be implemented.

I'd suggest the following:
1. Remove the entire implementation of `_forward` (this is completely redundant since it does exactly what `forward` does) and `_transform`.
2. Implement a `_compute_μ_σ²` or something that computes the `μ` and `σ²` (you're welcome to call this function `_transform` if you want). Or you can implement one for each, or whatever. Just separate out the computation of the parameters that are actually being used in different methods:)
3. `forward` should compute both `logabsdetjac` and `rv` in most efficient way; if you want to re-use the computation for `σ²` just copy-paste the implementation of `logabsdetjac` into `forward`. It's no more than a line anyways, so it's fine.

Afterwards `forward` should look something like
```julia
function forward(t::BatchNormFlow, x)
    μ, σ² = _compute_μ_σ²(t, x)  # or just put the computation in here
    γ = exp.(reshape(t.logγ, as...))
    x̂ = (x .- μ) ./ sqrt(σ² + ϵ)
    logjac = (sum(t.logγ - log.(σ² .+ t.ϵ) / 2)) .* typeof(Flux.data(x))(ones(Float32, size(x, 2))')
    return (rv=γ .* x̂, logabsdetjac=logjac)
end
```

src/norm.jl
torfjelde: Also, what is the `xsimilar`? I don't think the return-value is correct here.

yebai:
> Should it be named InvertibleBatchNorm as named in @xukai92's work or just BatchNorm? Being a part of Bijectors.jl, I feel invertibility is implied.


`InvertibleBatchNorm` looks reasonable to me.

src/Bijectors.jl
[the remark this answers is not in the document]
MikeInnes: FYI, treelike has been replaced with `@functor` on Flux master. It should also be easier to get `@functor` types to play nicely with things like ForwardDiff. Duplicating it is entirely fine, but if people end up using it seriously then it'd be very easy to just make it a simple package.

src/norm.jl
sharanry: @torfjelde Could you give an example when this could happen?

sharanry:
@torfjelde @xukai92 Could you take a look at this whenever possible?

src/batch_norm.jl
```diff
@@ -0,0 +1,114 @@
...
+    logγ::T1
+    μ::T2  # moving mean
+    σ²::T2 # moving st
+    ϵ::T3
+    momentum::T3
+    active::Bool # true when training
+end
+
+InvertibleBatchNorm(dims::Int, container=Array; ϵ=1f-5, momentum=0.1f0) = InvertibleBatchNorm(
+    container(zeros(Float32, dims, 1)),
```
xukai92: We should make `Float32` optional.

test/batch_norm.jl
```diff
@@ -0,0 +1,26 @@
...
+    @test inv(flow)(flow(z)) ≈ z
+    @test (inv(flow) ∘ flow)(z) ≈ z
+
+    @test_throws AssertionError forward(flow, randn(2,10))
+    
+    @test logabsdetjac(inv(flow), flow(z)) ≈ - logabsdetjac(flow, z)
+
+    y = flow(z)
+    @test log(abs(det(ForwardDiff.jacobian(flow, z)))) ≈
+     sum(logabsdetjac(flow, z)) rtol=1e-4 # fails return double
```
xukai92: Making `Float32` optional in the constructor could fix this.

src/norm.jl
xukai92: Should be fine with `Trakcer` considering it's adapted from Flux. Would be nice to provide some examples which could be potentiall problematic @torfjelde

test/batch_norm.jl
sharanry: Thanks! That helped.

src/batch_norm.jl
```diff
@@ -0,0 +1,118 @@
...
+                    momentum::AbstractFloat=0.1f0) = (
+                    InvertibleBatchNorm(
+                            container(zeros(dims, 1)),
+                            container(zeros(dims, 1)),
+                            zeros(dims, 1),
+                            ones(dims, 1),
+                            ϵ,
+                            momentum,
+                            true
+                    ))
```
willtebbutt: Style, please consider:
```julia
function InvertibleBatchNorm(
    dims::Int,
    container=Array; 
    ϵ::AbstractFloat=1f-5,
    momentum::AbstractFloat=0.1f0,
)
    return InvertibleBatchNorm(
        container(zeros(dims, 1)),
        container(zeros(dims, 1)),
        zeros(dims, 1),
        ones(dims, 1),
        ϵ,
        momentum,
        true
    )
end
```

src/batch_norm.jl
```diff
@@ -0,0 +1,118 @@
...
+    channels = size(x, dims)
+    affinesize = ones(Int, dims)
+    affinesize[end-1] = channels
+    return affinesize
+end
+
+logabsdetjac(
+    t::T,
+    x
+) where{T<:InvertibleBatchNorm} = forward(t, x).logabsdetjac
```
willtebbutt: Style: please consider:
```julia
logabsdetjac(t::InvertibleBatchNorm, x) = forward(t, x).logabsdetjac
```

src/batch_norm.jl
```diff
@@ -0,0 +1,118 @@
...
+
+function _compute_μ_σ²(t::InvertibleBatchNorm, x)
+    @assert size(x, ndims(x)) ==
```
willtebbutt: Style: this assertion is quite hard to read. Maybe consider the functional macro style:
```julia
@assert(
    size(x, ndims(x)) == length(t.μ),
    "`InvertibleBatchNorm` expected $(length(t.μ)) channels, got $(size(x,ndims(x)))"
)
```

test/runtests.jl
```diff
@@ -4,3 +4,4 @@ Random.seed!(123456)
 
 include("interface.jl")
 include("transform.jl")
+include("batch_norm.jl")
```
willtebbutt: Needs a new-line added. Possibly necessary to set your editor to ensure one is present?

src/norm.jl
torfjelde:
```julia
using Bijectors, Tracker

julia> x = param(ones(1, 2))
Tracked 1×2 Array{Float64,2}:
 1.0  1.0

julia> b = Bijectors.InvertibleBatchNorm(2, Tracker.param)
InvertibleBatchNorm{TrackedArray{…,Array{Float64,2}},Array{Float64,2},Float32}([0.0; 0.0] (tracked), [0.0; 0.0] (tracked), [0.0; 0.0], [1.0; 1.0], 1.0f-5, 0.1f0, true)

julia> logabsdetjac(b, x)
ERROR: MethodError: no method matching Float64(::Tracker.TrackedReal{Float64})
Closest candidates are:
  Float64(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:185
  Float64(::T<:Number) where T<:Number at boot.jl:725
  Float64(::Int8) at float.jl:60
  ...
Stacktrace:
 [1] convert(::Type{Float64}, ::Tracker.TrackedReal{Float64}) at ./number.jl:7
 [2] setindex! at ./array.jl:771 [inlined]
 [3] setindex! at ./multidimensional.jl:414 [inlined]
 [4] copyto!(::Array{Float64,2}, ::TrackedArray{…,Array{Float64,2}}) at ./multidimensional.jl:830
 [5] convert(::Type{Array{Float64,2}}, ::TrackedArray{…,Array{Float64,2}}) at ./array.jl:497
 [6] setproperty!(::InvertibleBatchNorm{TrackedArray{…,Array{Float64,2}},Array{Float64,2},Float32}, ::Symbol, ::TrackedArray{…,Array{Float64,2}}) at ./sysimg.jl:19
 [7] forward(::InvertibleBatchNorm{TrackedArray{…,Array{Float64,2}},Array{Float64,2},Float32}, ::TrackedArray{…,Array{Float64,2}}) at /scratch/tef30/.julia/dev/Bijectors/src/batch_norm.jl:87
 [8] logabsdetjac(::InvertibleBatchNorm{TrackedArray{…,Array{Float64,2}},Array{Float64,2},Float32}, ::TrackedArray{…,Array{Float64,2}}) at /scratch/tef30/.julia/dev/Bijectors/src/batch_norm.jl:53
 [9] top-level scope at none:0
```

test/batch_norm.jl
```diff
@@ -0,0 +1,26 @@
+using Test
+using Bijectors, ForwardDiff, LinearAlgebra
+using Random: seed!
+
+seed!(1)
+
+@testset "InvertibleBatchNorm" begin
+    z = randn(20, 2)
```
torfjelde: Is this supposed to be a batch of 20 2-dim inputs? If so, this should be `(2, 20)`. Distributions.jl use this kind of shape  (because iterating through [:, i] is faster than [i, :]) so also follow the same convention.

src/norm.jl
torfjelde: Changing the `active` field makes it run without issues (though the batch-shape should still be transposed):
```julia
julia> b.active = false
false

julia> logabsdetjac(b, x)
Tracked 1×1 Adjoint{Float64,Array{Float64,2}}:
 -9.999949747714609e-6
```

But I suspect it's supposed to work when `active == true` also, right?

torfjelde: Oh, and the transform and everything related should also work with a single input rather than always requiring a batch. A quick fix is to dispatch differently on `AbstractVector` and `AbstractMatrix`

xukai92: Good point. I never tried this transform with a single vector but we should make it work.

test/batch_norm.jl
xukai92: Yes we should stick to column major for sure as it's more efficient in Julia. The original implementaion from Flux is column major so it looks like a mistake in the test for me. @sharanry can you give a try to fix?

src/norm.jl
torfjelde: > Transform and logabsdetjac with single variable seems to work provided it is in the format (n_dims, batch_size) and it is not tracked. I think we should enforce the input to be (n_dims, batch_size) or another possible solution is adding x = reshape(x, :, 1) everywhere were x is taken as input.

I really don't think we want to enforce `(n_dims, batch_size)`. What I did for the other normalizing flows (they had the same problem), is that I just defined a `(b:Flow)(x::AbstractVector) = b(reshape(x, 1, :))`. You can just do this for the relevant functions, e.g. `logabsdetjac`, and we're good:)

> Do you think this will work for the tracked input problem?

I don't quite see what the purpose of that change is to be honest. From a quick look it doesn't actually seem to have a different effect than the old code. Have you tried this? I don't know exactly what the problem above is, as I haven't looked properly into it yet. I'd suggest looking at the lines referenced in the error and work from there.

sharanry: > ERROR: AssertionError: `InvertibleBatchNorm` expected 2 channels, got 1
This initial error which you had come across is because of the input not taken as column-major. Now, this shouldn't be a problem. 

After fixing this another problem occurs.
[This](https://github.com/TuringLang/Bijectors.jl/pull/43/commits/03a16592ac2de7d948200ad71794ecc2f5debef7) commit is to fix the following error.
```julia
julia> b = InvertibleBatchNorm(2)
InvertibleBatchNorm{Array{Float64,2},Array{Float64,2},Float32}([0.0; 0.0], [0.0; 0.0], [0.0; 0.0], [1.0; 1.0], 1.0f-5, 0.1f0, true)

julia> x = param(ones(2, 1)); b(x)
ERROR: MethodError: no method matching Float64(::Tracker.TrackedReal{Float64})
Closest candidates are:
  Float64(::Real, ::RoundingMode) where T<:AbstractFloat at rounding.jl:194
  Float64(::T<:Number) where T<:Number at boot.jl:741
  Float64(::Int8) at float.jl:60
  ...
Stacktrace:
 [1] convert(::Type{Float64}, ::Tracker.TrackedReal{Float64}) at ./number.jl:7
 [2] setindex! at ./array.jl:769 [inlined]
 [3] setindex! at ./multidimensional.jl:460 [inlined]
 [4] copyto!(::Array{Float64,2}, ::TrackedArray{…,Array{Float64,2}}) at ./multidimensional.jl:877
 [5] convert(::Type{Array{Float64,2}}, ::TrackedArray{…,Array{Float64,2}}) at ./array.jl:482
 [6] setproperty!(::InvertibleBatchNorm{Array{Float64,2},Array{Float64,2},Float32}, ::Symbol, ::TrackedArray{…,Array{Float64,2}}) at ./sysimg.jl:19
 [7] forward(::InvertibleBatchNorm{Array{Float64,2},Array{Float64,2},Float32}, ::TrackedArray{…,Array{Float64,2}}) at /media/sharan/Work/Coding/MachineLearning/Projects/julia/Bijectors.jl/src/batch_norm.jl:93
 [8] (::InvertibleBatchNorm{Array{Float64,2},Array{Float64,2},Float32})(::TrackedArray{…,Array{Float64,2}}) at /media/sharan/Work/Coding/MachineLearning/Projects/julia/Bijectors.jl/src/batch_norm.jl:106
 [9] top-level scope at none:0
```

xukai92:
@sharanry what's the current status of this PR?

sharanry:
@xukai92 [This](https://github.com/TuringLang/Bijectors.jl/pull/43#discussion_r341577819) still needs to be resolved.

src/norm.jl
xukai92: > Good point. I never tried this transform with a single vector but we should make it work.

OK I was wrong here. We never input a single vector to batch norm. Batch norm only works when input is a batch, i.e. more than a vector.

xukai92:
Is the error in https://github.com/TuringLang/Bijectors.jl/pull/43#discussion_r341577819 based on the latest commit or just copied from Tor's comments?

xukai92:
If so, it's unclear to me why assining values to `t.μ` will cause this error at all, as `t.μ` is not tracked (https://github.com/TuringLang/Bijectors.jl/pull/43/files#diff-ebb92afe0a9fbc40f2bb7e4e809345f0R37)

src/norm.jl
torfjelde: > OK I was wrong here. We never input a single vector to batch norm. Batch norm only works when input is a batch, i.e. more than a vector.

Ah, true! Should maybe raise an explicit error for this? So that people like me don't make that mistake again:)

yebai:
Closed in favour of #76

PR #44, opened 2019-09-13
Making dimensionality part of the Bijector type

torfjelde:
# Overview
This PR attempts to solve #35 by introducing the dimensionality of input/output (the same) `N` into the `Bijector` type.

Pros:
1. Solves ambiguities referred to in #35 
2. Catches dimension-mismatch in compositions at compile-time
3. Most standard subtypes of `Bijector{N}` are only well-defined for a specific value of `N`, so in these cases the only change necessary is, say, `Specific1DBijector <: Bijector` → `Specific1DBijector <: Bijector{1}`. E.g. `SimplexBijector` is only well-defined for vector input, so `SimplexBijector <: Bijector{1}`. To get a quick overview of the differences from a user-perspective, have a look at the diff for `tests/interface.jl`. There's barely any change.
4. Batch computation now works nicely:)

Cons:
1. `Log()` looks nicer than `Log{0}()`, etc. But again, due to (3) in pros this is in most cases not a huge issue from a user perspective.
2. Mismatches in _sizes_ are not caught at the moment. Though we can of course add run-time checks for this.

# To-be discussed
- [ ] Is this really the best solution to make batch-computation work? It does put a bit more work on the user for custom `Bijector` since it requires explicit implementation of `logabsdetjac` for both standard case and the batch computation. But of course if the user doesn't want batch-computation, then there's no more work than before.
- [ ] Is this design more restrictive in what we can do with `Bijector`? Could people want to implement something more general than just standard `Real` or `AbstractArray`, and this prohibits it? If so, is this of any priority to us?
- [x] Something similar to StaticArrays.jl was suggested to catch size-mismatches at compile time. After thinking about this for a bit I came to the conclusion that it would get real nasty real quick and would be much more restrictive than we'd like. Also not entirely sure how to go about implementing this as I haven't looked properly into the src of StaticArrays.jl yet.

# TODOs
- [x] Update README.md
- [x] Update docstrings
- [ ] Conclude discussions:)


src/interface.jl
```diff
@@ -38,26 +39,32 @@ end
 ######################
 
 "Abstract type for a `Bijector`."
-abstract type Bijector end
+abstract type Bijector{N} end
```
xukai92: Not compulsory but would be nice to rename this to `AbstractBijector`.

src/interface.jl
```diff
@@ -38,26 +39,32 @@ end
...
+abstract type ADBijector{AD, N} <: Bijector{N} end
 
 """
     inv(b::Bijector)
     Inversed(b::Bijector)
 
 A `Bijector` representing the inverse transform of `b`.
 """
-struct Inversed{B <: Bijector} <: Bijector
+# TODO: can we do something like `Bijector{N}` instead?
```
xukai92: What does this comment mean?

src/interface.jl
```diff
@@ -254,24 +272,26 @@ function forward(cb::Composed, x)
...
 -###########
 -# Stacked #
 -###########
+const ZeroOrOneDimBijector = Union{Bijector{0}, Bijector{1}}
+
 """
     Stacked(bs)
     Stacked(bs, ranges)
-    stack(bs::Bijector...)
+    stack(bs::Bijector{Dim=0}...)
```
xukai92: `Dim=0`?

src/interface.jl
xukai92: Suppose `N` refers to the dimension, maybe rename it to `D`?

src/interface.jl
```diff
@@ -440,57 +471,152 @@ logabsdetjac(b::Logit{<:Real}, x) = @. - log((x - b.a) * (b.b - x) / (b.b - b.a)
...
+_logabsdetjac_scale(a::Real, x::Real, ::Type{Val{0}}) = log(abs(a))
+_logabsdetjac_scale(a::Real, x::AbstractVector, ::Type{Val{0}}) = fill(log(abs(a)), length(x))
+_logabsdetjac_scale(a::Real, x::AbstractVector, ::Type{Val{1}}) = log(abs(a)) * length(x)
+_logabsdetjac_scale(a::Real, x::AbstractMatrix, ::Type{Val{1}}) = fill(log(abs(a)) * size(x, 1), size(x, 2))
+_logabsdetjac_scale(a::AbstractVector, x::AbstractVector, ::Type{Val{1}}) = sum(log.(abs.(a)))
+_logabsdetjac_scale(a::AbstractVector, x::AbstractMatrix, ::Type{Val{1}}) = fill(sum(log.(abs.(a))), size(x, 2))
+_logabsdetjac_scale(a::AbstractMatrix, x::AbstractVector, ::Type{Val{1}}) = log(abs(det(a)))
+_logabsdetjac_scale(a::AbstractMatrix, x::AbstractMatrix{T}, ::Type{Val{1}}) where {T} = log(abs(det(a))) * ones(T, size(x, 2))
+
+# adjoints for 0-dim and 1-dim `Scale` using `Real`
```
xukai92: Capital letter for single line comments.

src/interface.jl
```diff
@@ -440,57 +471,152 @@ logabsdetjac(b::Logit{<:Real}, x) = @. - log((x - b.a) * (b.b - x) / (b.b - b.a)
...
+function _logabsdetjac_scale(a::TrackedReal, x::Real, ::Type{Val{0}})
+    return track(_logabsdetjac_scale, a, data(x), Val{0})
+end
+@grad function _logabsdetjac_scale(a::Real, x::Real, ::Type{Val{0}})
+    return _logabsdetjac_scale(data(a), data(x), Val{0}), Δ -> (inv(data(a)) .* Δ, nothing, nothing)
+end
+
+# need to treat `AbstractVector` and `AbstractMatrix` separately due to ambiguity errors
```
xukai92: Capital letter for single line comments.

src/interface.jl
```diff
@@ -732,21 +866,27 @@ Base.size(td::Transformed) = size(td.dist)
...
+
 function _logpdf(td::MvTransformed, y::AbstractVector{<:Real})
     res = forward(inv(td.transform), y)
-    return logpdf(td.dist, res.rv) .+ res.logabsdetjac
+    return logpdf(td.dist, res.rv) + res.logabsdetjac
 end
 
 function _logpdf(td::MvTransformed{<:Dirichlet}, y::AbstractVector{<:Real})
     T = eltype(y)
     ϵ = _eps(T)
```
xukai92: Where did we define `_eps`?

src/interface.jl
```diff
@@ -440,57 +471,152 @@ logabsdetjac(b::Logit{<:Real}, x) = @. - log((x - b.a) * (b.b - x) / (b.b - b.a)
...
+(b::Scale)(x) = b.a .* x
 (b::Scale{<:Real})(x::AbstractArray) = b.a .* x
-(b::Scale{<:AbstractVector{<:Real}})(x::AbstractMatrix{<:Real}) = x * b.a
+(b::Scale{<:AbstractMatrix})(x::AbstractArray) = b.a * x
+(b::Scale{<:AbstractVector{<:Real}, 2})(x::AbstractMatrix{<:Real}) = b.a .* x
+
+inv(b::Scale{T, D}) where {T, D} = Scale(inv(b.a); dim = Val{D})
+inv(b::Scale{<:AbstractVector, D}) where {D} = Scale(inv.(b.a); dim = Val{D})
+
+# We're going to implement custom adjoint for this
```
xukai92: I didn't check the adjoint implementations. I'm fine with then as long as we have test of them aginst AD based.

src/interface.jl
torfjelde: Sure to both!

src/interface.jl
torfjelde: Since `Inversed{<: Bijector{D}, D}` is specifying the dimensionality `D` _twice_, I was hoping that there would be some way of dropping the second `D`. Don't think this can be done though, and the current approach is probably as good as it's going to get.

src/interface.jl
torfjelde: What's the question? :smile: You think it should be `D=0` instead? Or the fact that we're restricting to only allow `D=0`?

src/interface.jl
torfjelde: In `src/Bijectors.jl`

src/interface.jl
torfjelde: Good point; I'll add that :+1:

Also the reason why I've added these is because of type-issues arising when one input is non-tracked and parameters potentially are tracked, etc. Difficult to know which type to `zero(T)`.

torfjelde: Done

src/interface.jl
torfjelde: Actually, I now remember why I used `N` instead of `D`. 

You're often going to see `D` used as a `D<:Distribution` _and_ `N` is also used to denote the same thing in `AbstractArray{T, N}`, so this is why I decided to do that. And I think I'd prefer sticking with `N`

torfjelde:
Btw, I think it would be a good idea to separate implementations of `Bijector` into their separate files, e.g. instead of implementing `Logit` and `SimplexBijector` in the same file `src/interface.jl`, we move them to `src/bijectors/logit.jl` and `src/bijectors/simplex.jl`, respectively.

I have several new bijectors in the pipeline and IMO such a project structure makes much more sense as the number of bijectors grows.

Question: should I include this in this PR or do it in a separate PR?

xukai92:
> I have several new bijectors in the pipeline and IMO such a project structure makes much more sense as the number of bijectors grows.

Definitely

> Question: should I include this in this PR or do it in a separate PR?

I'm fine if you do it in this PR. But you might need to resolve some conflicts to make other PRs merge afterwards.

torfjelde:
> I'm fine if you do it in this PR. But you might need to resolve some conflicts to make other PRs merge afterwards.

Then I'll do that. And yeah, but probably going to have to do that anyways!:)

src/interface.jl
xukai92: I guess you can do
```julia
struct Inversed{N} <: Bijector{N}
    orig::Bijector{N}
end
```
if you want. But it's not type stable right?

torfjelde:
> Then I'll do that. And yeah, but probably going to have to do that anyways!:)

Actually, let's just get this thing merged and then we can do that after. There's so much work that depends on this PR that I'd like to just get it merged asap

xukai92:
There some conflicts need to be resolved. After that is solved I'm happy to merge it given CI passes.

xukai92:
The only faild CIs are in the nightly Julia version. I guess it's fine to merge it now. I will do so before I sleep if no objection.

torfjelde:
> The only faild CIs are in the nightly Julia version. I guess it's fine to merge it now. I will do so before I sleep if no objection.

Yeah, from a quick look at the logs it seems like it's an issue with Tracker.jl rather than us so will probably be resolved in the future by updating deps.

PR #45, opened 2019-09-13
Definition and short description of Bijector in README

torfjelde:


xukai92:
Lgtm. I can merge it if it's ready.

README.md
```diff
@@ -63,6 +63,14 @@ julia> logpdf_with_trans(dist, x, true) # considering the transformation
 ```
 
 ## `Bijector` interface
+A `Bijector` is a differentiable bijection with a differentiable inverse. That's basically it. 
+
+One particular useful application of such methods is the (very profitable) business of transforming (usually, continuous) probability densities. If we transfrom a random variable `x ~ p(x)` to `y = b(x)` where `b` is a `Bijector`, we also get a canonical density `q(y) = p(b⁻¹(y)) |det J(b⁻¹, y)|` for `y`! Here `J(b⁻¹, y)` is the jacobian of the inverse transform evaluated at `y`. `q` is also known as the _push-forward_ of `p` by `b` in measure theory.
```
willtebbutt: The first sentence here could do with some work. It's not quite clear what `methods` refers to, and `particular useful` and `(very profitable)` kind of say the same things. Maybe go with something like
```
The primary application of `Bijector`s is the (very profitable) business of transforming (usually continuous) probability densities.
```
instead?

Maybe also loose the exclamation mark at the end of the second sentence. Visually, it looks quite cluttered in the rendered version.

torfjelde: I like that :+1:

torfjelde: Did as you suggested:)

xukai92:
Thanks for the effort @torfjelde and the careful review @willtebbutt!

PR #48, opened 2019-10-19
Coupling flows

torfjelde:
## Overview
This PR introduces a `Bijector` called `CouplingLayer`. 

### "Mathematical" definition
For a proper definition, see [Section 3.4, 2].
A `CouplingLayer` consists of a `Bijector` (or rather a *constructor* of a `Bijector`), a `PartitionMask`, and some arbitrary parameter map `f`.

The idea behind a `CouplingLayer` is the following:
1. Take input `x`
2. Partition `x` into 3 parts: `x1, x2, x3`
3. `θ = f(x_2)`
4. Construct `Bijector`: `b = B(θ)`
5. Apply `b` to `x1` and identity-map to the rest: `y1 = b(x1)` and `y2 = x2, y3 = x3`
6. Recombine and return `y = [y1..., y2..., y3...]`

This mapping is indeed bijective with inverse defined as follows:
1. Take input `y`
2. Partition `y` into 3 parts: `y1, y2, y3`
3. `θ = f(y_2)` (same as Step 3 above since `y_2 == x_2` is indeed true)
4. Construct `Bijector`: `b = B(θ)`
5. _Invert_ `b`: `b⁻¹ = inv(b)`
6. Apply `b⁻¹` to `y1` and identity-map to the rest: `x1 = b⁻¹(y1)` and `x2 = y2, x3 = y3`
6. Recombine and return `x = [x1..., x2..., x3...]`

In this case, `f` can be any mapping with the correct input- and output-dimensions, e.g. a neural network.

### Interface
The current implementation is as follows:

``` julia
struct CouplingLayer{B, M, F} <: Bijector{1} where {B, M <: PartitionMask, F}
    mask::M
    θ::F
end
CouplingLayer(B, mask::M, θ::F) where {M, F} = CouplingLayer{B, M, F}(mask, θ)

# ...

function (cl::CouplingLayer{B})(x::AbstractVector) where {B}
    # partition vector using `cl.mask::PartitionMask`
    x_1, x_2, x_3 = partition(cl.mask, x)

    # construct bijector `B` using θ(x₂)
    b = B(cl.θ(x_2))

    # recombine the vector again using the `PartitionMask`
    return combine(cl.mask, b(x_1), x_2, x_3)
end

# ...
```
In addition to implementing a `Bijector`, it also has the following two methods:

``` julia
"Returns the constructor of the coupling law."
coupling(cl::CouplingLayer{B}) where {B} = B

"Returns the coupling law constructed from `x`."
function couple(cl::CouplingLayer{B}, x::AbstractVector) where {B}
    # partition vector using `cl.mask::PartitionMask`
    x_1, x_2, x_3 = partition(cl.mask, x)

    # construct bijector `B` using θ(x₂)
    b = B(cl.θ(x_2))

    return b
end
```

And an example use-case:

``` julia
julia> m = PartitionMask(3, [1], [2]) # <= going to use x[2] to parameterize transform of x[1]
PartitionMask{SparseArrays.SparseMatrixCSC{Float64,Int64}}(
  [1, 1]  =  1.0, 
  [2, 1]  =  1.0, 
  [3, 1]  =  1.0)

julia> cl = CouplingLayer(Shift, m, identity) # <= will do `y[1:1] = x[1:1] + x[2:2]`;

julia> x = [1., 2., 3.];

julia> cl(x)
3-element Array{Float64,1}:
 3.0
 2.0
 3.0
 
 julia> cl = CouplingLayer(θ -> Shift(θ[1]), m, identity) # <= results in the same as above
 
 julia> cl(x)
3-element Array{Float64,1}:
 3.0
 2.0
 3.0
```
A more exciting example is something like
```julia
julia> using Flux

julia> using Bijectors: Shift, Scale

julia> nn = Chain(Dense(1, 2, relu), Dense(2, 2));

julia> cl = CouplingLayer(θ -> Shift(θ[2]) ∘ Scale(θ[1]), m, nn);

julia> cl(x)
Tracked 3-element Array{Float64,1}:
 -0.8811154365539551
  2.0               
  3.0
```

## Issues / Things to discuss
- [x] `B` _needs_ to be known at compile-time, otherwise the construction `CouplingLayer(B, mask, f)` throws an error. I don't have a reproducible example at the moment, but I recall having an issue at some point where `B` was an an anonymous function created in some nested scopes which resulted into a `DataType` error of sorts. I don't think this should be prohibitive in any cases, but might in certain contrived examples be annoying. I'm honestly not entirely sure why this happens or how it works...
- [ ] There is some overlap in the part `B` and `f` plays. For example you can either do `B = θ -> Shift(θ[1])` and `f = identity`, or `B = Shift` and `f(x2) = x2[1]`. Originally I did this because I intended for `B` to be `Type{BT} where BT <: Bijector` but I don't think this works. And so now `B` is just assumed to map the output of `f` to a `Bijector`, which I don't think we can enforce using the type-system. Also, this separation makes sense due to the maps having two different purposes, but at the same time, in implementation, we could probably just do `B ∘ f` and just map `x2` to the `Bijector` in one. I don't like that either though.
- [x] Role of `PartitionMask`. I've introduced `PartitionMask` to make the this partitioning operation AD-, matrix-, and (possibly) GPU-friendly, instead of just doing `x1, x2, x3 = x[1:n1], x[n1 + 1:n2], x[n2 + 1:end]`. `PartitionaMask` internally uses three `SparseArray` to perform `partition` and `combine`. Does this seem reasonable?

# TODOs
- [ ] Add to README

## References
[1] Dinh, L., Krueger, D., & Bengio, Y., Nice: non-linear independent components estimation, CoRR, (),  (2014). 
[2] Kobyzev, I., Prince, S., & Brubaker, M. A., Normalizing flows: introduction and ideas, CoRR, (),  (2019).


src/couplings.jl
```diff
@@ -0,0 +1,244 @@
...
+    x_1, x_2, x_3 = partition(cl.mask, x)
+
+    # construct bijector `B` using θ(x₂)
+    b = B(cl.θ(x_2))
+
+    # recombine the vector again using the `PartitionMask`
+    return combine(cl.mask, b(x_1), x_2, x_3)
+end
+function (cl::CouplingLayer{B})(x::AbstractMatrix) where {B}
+    return hcat([cl(x[:, i]) for i = 1:size(x, 2)]...)
```
mohamed82008: Not great when there are many columns. Use `reduce`.

src/couplings.jl
```diff
@@ -0,0 +1,244 @@
...
+    return PartitionMask(A_1, A_2, spzeros(n, 0))
+end
+PartitionMask(x::AbstractVector, indices) = PartitionMask(length(x), indices)
+
+"""
+    combine(m::PartitionMask, x_1, x_2, x_3)
+
+Combines `x_1`, `x_2`, and `x_3` into a single vector.
+"""
+@inline combine(m::PartitionMask, x_1, x_2, x_3) = m.A_1 * x_1 .+ m.A_2 * x_2 .+ m.A_3 * x_3
```
mohamed82008: Nice use of sparse matrices and transposes but I doubt this will be faster or more Tracker and GPU friendly than simply `collect(sparsevec(inds1, x1, n)) + collect(sparsevec(inds2, x2, n)) + ...`. Some of these functions may need to be defined for `TrackedArray` or `GPUArray` but nothing should stop that in principle. Also do the indices need to be all over the place, or can we assume contiguity of the partitions?

src/couplings.jl
```diff
@@ -0,0 +1,244 @@
...
+3-element Array{Float64,1}:
+ 3.0
+ 2.0
+ 3.0
+```
+
+# References
+[1] Kobyzev, I., Prince, S., & Brubaker, M. A., Normalizing flows: introduction and ideas, CoRR, (),  (2019). 
+"""
+struct CouplingLayer{B, M, F} <: Bijector{1} where {B, M <: PartitionMask, F}
```
mohamed82008: `B` can be a field. Type parameters must be `isbits` or Symbols. I don't think we need to commit to this extra restriction. `B` can be any function that returns a `Bijector`.

src/couplings.jl
```diff
@@ -0,0 +1,244 @@
...
+    x_1, x_2, x_3 = partition(cl.mask, x)
+    b = B(cl.θ(x_2))
+
+    # `B` might be 0-dim in which case it will treat `x_1` as a batch
+    # therefore we sum to ensure such a thing does not happen
+    return sum(logabsdetjac(b, x_1))
+end
+
+function logabsdetjac(cl::CouplingLayer{B}, x::AbstractMatrix) where {B}
+    r = [logabsdetjac(cl, x[:, i]) for i = 1:size(x, 2)]
```
mohamed82008: I think we can delegate this column-wise iteration to the `logabsdetjac` function. We can then define efficient adjoints for `logabsdetjac` when `x` is `TrackedArray`. If the tracking comes from `cl`, and no scalar operations happen in `logabsdetjac`, I think `Tracker` should just work without requiring additional adjoint definitions.

src/couplings.jl
torfjelde: You mean
```julia
foldl(hcat, [cl(x[:, i]) for i = 1:size(x, 2)])
```
?

src/couplings.jl
torfjelde: It should be more GPU-friendly though, or at least make it easier to support GPUs, since you can instantiate `PartitionMask` with the corresponding GPU arrays. E.g. `PartitionMask(cu(A_1), cu(A_2), cu(A_3))`, right?

src/couplings.jl
torfjelde: > Type parameters must be isbits or Symbols.
So this is probably the issue I mention in things to discuss; is there somewhere where I can read about _when_ something is `isbits`?:)

And your suggestion seems like a good solution; I'll do that :+1:

src/couplings.jl
torfjelde: >  We can then define efficient adjoints for logabsdetjac when x is TrackedArray.
I was thinking about something similar the other week, but at the time it seemed to me like it wouldn't work. I believe my reason was just that since the only requirement is that `B` _returns_ a `Bijector`, this could have `Tracked` parts coming from somewhere else, i.e. you define `B` by closing over some `Tracked` thing. I'm uncertain how this works with defining custom adjoints :confused: 

>  I think we can delegate this column-wise iteration to the logabsdetjac function. 
I'm not entirely certain what you mean here. Do you want to make it so that `logabsdetjac` _always_ takes a matrix as input?

src/couplings.jl
torfjelde: Should we just combine `B` and `F` then? The only way we use these are `B(cl.θ(x_2))`, so we could just instead remove the `B` field and let `F` be `B ∘ cl.θ`.

src/couplings.jl
torfjelde: > Also do the indices need to be all over the place, or can we assume contiguity of the partitions?

I introduce `Permute <: Bijector` in a different PR, so we _could_ assume contiguity of partitions with some additional work to make things easier on the user. But is this worth it? I guess with my current implementation it won't make much difference, but do you have different approach that makes this more efficient (assuming contiguity of partitions)?

Also, I'll have a GPU available to work on soon, at which point I'll start making everything compatible with GPU.

src/couplings.jl
torfjelde: A couple of reasons why I actually prefer the current approach vs. the two we've touched upon here:
- With `B` and `F` separate, we can dispatch differently on the coupling law `B` returning a `Bijector` and the parameter map `F`. E.g. define an `Affine(W, b) = Shift(b) ∘ Scale(W)` and then define `(cl::CouplingLayer{typeof(Affine)})(x)`. In _most_ cases this functionality won't be used, but at least with the current approach we have the option of doing this.
- With `B` in the type, we can do `CouplingLayer(Shift, M, F)` and things are type-stable. If we add it as a _field_, `CouplingLayer(Shift, mask, F)` results in a `UnionAll` type of the `B` field and thus we lose type-stability. As a result, to guarantee type-stability, we'd have combine `B` and `F` as mentioned above. But this means that we cannot dispatch differently on the coupling law and parameter map used, etc.

src/couplings.jl
mohamed82008: Yes

src/couplings.jl
mohamed82008: Sorry I missed your comments. You can try both and see if the simple version works on the GPU.

src/couplings.jl
mohamed82008: Note that the type of `B` can be a type parameter. I don't see how this will harm type stability compared to directly using `B` itself as a type parameter.

src/couplings.jl
mohamed82008: I mean to define `logabsdetjac` for matrix `x`.

src/couplings.jl
torfjelde:
```julia
using Test

using Bijectors
using Bijectors: CouplingLayer, Shift, Scale, PartitionMask, partition, combine

### On GPU
using CuArrays, SparseArrays
x = Float32.(randn(100_000))
x_gpu = cu(x)

mask = PartitionMask(length(x), collect(1:100), collect(500:1000))
# Convert directly into `CuSparseMatrixCSC` instead of doing `sparse ∘ cu`
# due to the fact that `cu` will first load the full non-sparse matrix into
# memory before converting into sparse using `sparse`.
mask_gpu = PartitionMask(
    CUSPARSE.CuSparseMatrixCSC.(
        convert.(SparseMatrixCSC{Float32, Int}, (mask.A_1, mask.A_2, mask.A_3))
    )...
);

x1, x2, x3 = partition(mask_gpu, x_gpu)
y = combine(mask_gpu, x1, x2, x3)
```
and then
```julia
using BenchmarkTools
@btime begin
    x1, x2, x3 = partition($mask_gpu, $x_gpu)
    y = combine($mask_gpu, x1, x2, x3)
end 
# Memory usage is wrong since not accounting for GPU memory
# => 1.488 ms (203 allocations: 7.22 KiB)

@btime begin
    x1, x2, x3 = partition($mask, $x)
    y = combine($mask, x1, x2, x3)
end
# => 573.984 μs (15 allocations: 3.82 MiB)
```
Works without any issues. Though I ran into an issue where if I was using `Float64` for the sparse matrix, `transpose(mask_gpu.A_1) * x_gpu` would silently do _nothing_. This is of course an issue with `CUSPARSE.jl` rather than with `Bijectors.jl`. 

With that being said, the above is about 3X slower on the GPU than on CPU. Not sure whether or not this is expected?

src/couplings.jl
torfjelde: So there's a slightly annoying issue.

In the case where `b = B(cl.θ(x_2))` returns a `Bijector{0}`, `logabsdetjac(b::Bijector{0}, x::AbstractMatrix)` is generally _not_ defined. Only `logabsdetjac(b::Bijector{0}, x::Union{Real, AbstractVector})` are defined.

I'm not entirely sure what to do in those cases. With the current implementation everything still works just fine. I do agree though, the current approach is sub-optimal :/

torfjelde: Also, there's the issue of what to do with `B(cl.θ(x_2))`. Not all bijectors will take a `Vector` instead of `Real` as parameter and have this imply that the evaluation of this `Bijector` works on a "batch" rather than a single input. E.g. if `cl.θ(x_2)` returns a `(param_size, batch_size)`, this doesn't necessarily mean that `B(...)(x)[i] == B(cl.θ(x_2)[i, :])(x[i, :])`.

mohamed82008: Can't we tell this from the "dimension" of the bijector?

torfjelde: Yeah, but not the "dimension" of the _parameters_ of it. It basically comes down to the fact that the following isn't true in general
```julia
# LHS might not even work
B(cl.θ(x_2))(x)[i, :] != B(cl.θ(x_2)[i, :])(x[i, :])
```

mohamed82008: I see, ignore my comment then. I will look into this again some other time to see if there is a way to make it work.

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
+using SparseArrays
+
+"""
+    PartitionMask{A}(A_1::A, A_2::A, A_3::A) where {A}
+
+This is used to partition and recombine a vector into 3 disjoint "subvectors".
```
xukai92: Why do we partition inputs into 3 disjoint subspaces? It is to support "multi-scale" flow or something?

xukai92:
What's the suggested way to implement a conditional coupling flow, e.g. to model p(x|y)? I guess I can concatenate x and y and set the mask to use the y part as the input of a NN. Is there a better way which avoids concatenating?

src/bijectors/coupling_layer.jl
torfjelde: You split into 3 parts (x₁, x₂, x₃) s.t. you can then define a mapping b(x₁; θ(x₂)) and x₃ isn't used for anything. Alternatively, you could just define θ(x₂, x₃) (i.e. only split into 2 parts) but then you again have to do a "filtering" on indices inside θ. And we can make it so that if you do `PartitionMask(A_1, A_2)`, then the last part is just an empty array (we already have such a constructor based on indices directly, i.e. `PartitionMask(n::Int, indices1, indices2)`)

torfjelde:
> What's the suggested way to implement a conditional coupling flow, e.g. to model p(x|y)? I guess I can concatenate x and y and set the mask to use the y part as the input of a NN. Is there a better way which avoids concatenating?

Yeah, exactly. I think in general that is the best approach.

src/bijectors/coupling_layer.jl
xukai92: I see. It makes sense. Thanks.

xukai92:
OK. It doesn't sound like very efficient in practice (e.g. when both x and y are large and on GPU). Do you think we could make it easier by making a specific type of mask?

torfjelde:
> OK. It doesn't sound like very efficient in practice (e.g. when both x and y are large and on GPU). Do you think we could make it easier by making a specific type of mask?

The issue is that of composition of `Bijector`. If the `Bijector` now takes two inputs; how do you compose with all other bijectors? Though I agree it would be nice to have, it's unclear how we'd go about doing this. I think we'd have to go down the route of defining input and output-spaces and so on, which will be _a lot_ of work both for us and the user.

xukai92:
That's a good point. Also I think I find a workaround without concatenating.

```julia
function (cl::CouplingLayer{B})(xt::Tuple) where {B}
    x_1, x_2 = xt
    b = B(cl.θ(x_2))
    return b(x_1)
end
```

torfjelde:
> That's a good point. Also I think I find a workaround without concatenating.

Yeah, so that is okay as long as you don't try to compose with other bijectors. Alternatively, depending on what you're doing, you can also just define `θ` as a closure over `y` and make `x_2` empty. Then you can use the standard `CouplingLayer`. This works nice if `y` represents something static, e.g. data, that you're not ever going to change.

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
...
+Combines `x_1`, `x_2`, and `x_3` into a single vector.
+"""
+@inline combine(m::PartitionMask, x_1, x_2, x_3) = m.A_1 * x_1 .+ m.A_2 * x_2 .+ m.A_3 * x_3
+
+"""
+    partition(m::PartitionMask, x)
+
+Partitions `x` into 3 disjoint subvectors.
+"""
+@inline partition(m::PartitionMask, x) = (transpose(m.A_1) * x, transpose(m.A_2) * x, transpose(m.A_3) * x)
```
xukai92: This is not type-safe. If `x` is `Array{Float32}`, paritions are returned as `Array{Float64}`. Can you fix this?

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
...
+    return PartitionMask(A_1, A_2, spzeros(n, 0))
+end
+PartitionMask(x::AbstractVector, indices) = PartitionMask(length(x), indices)
+
+"""
+    combine(m::PartitionMask, x_1, x_2, x_3)
+
+Combines `x_1`, `x_2`, and `x_3` into a single vector.
+"""
+@inline combine(m::PartitionMask, x_1, x_2, x_3) = m.A_1 * x_1 .+ m.A_2 * x_2 .+ m.A_3 * x_3
```
xukai92: This it not type-safe either.

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
...
+    PartitionMask(A_1::A, A_2::A, A_3::A) where {A <: AbstractMatrix{<:Real}} = new{A}(A_1, A_2, A_3)
+end
+
+function PartitionMask(
+    n::Int,
+    indices_1::AbstractVector{Int},
+    indices_2::AbstractVector{Int},
+    indices_3::AbstractVector{Int}
+)
+    A_1 = spzeros(n, length(indices_1));
```
xukai92:
```suggestion
    A_1 = spzeros(Bool, n, length(indices_1))
```

xukai92: Similar to others

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
...
+    A_2 = spzeros(n, length(indices_2));
+    A_3 = spzeros(n, length(indices_3));
+
+    for (i, idx) in enumerate(indices_1)
+        A_1[idx, i] = 1.0
```
xukai92:
```suggestion
        A_1[idx, i] = true
```

xukai92: Similar to others

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
...
+    indices_3::AbstractVector{Int}
+) = PartitionMask(n, indices_1, [i for i in 1:n if i ∉ (indices_1 ∪ indices_3)], indices_3)
+
+"""
+    PartitionMask(n::Int, indices)
+
+Assumes you want to _split_ the vector, where `indices` refer to the 
+parts of the vector you want to apply the bijector to.
+"""
+function PartitionMask(n::Int, indices)
```
xukai92: Similar to this one.

src/bijectors/coupling_layer.jl
```diff
@@ -0,0 +1,244 @@
...
+    # partition vector using `cl.mask::PartitionMask`
+    x_1, x_2, x_3 = partition(cl.mask, x)
+
+    # construct bijector `B` using θ(x₂)
+    b = B(cl.θ(x_2))
+
+    # recombine the vector again using the `PartitionMask`
+    return combine(cl.mask, b(x_1), x_2, x_3)
+end
+function (cl::CouplingLayer{B})(x::AbstractMatrix) where {B}
```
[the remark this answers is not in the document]
xukai92: Same question for logabsdetjac and those for inverse

xukai92 (dismissed):

src/bijectors/coupling_layer.jl
torfjelde: Isn't this correct behavior though? I see what you're saying, but think the fix is to just define the partitions also be `Float32`, no?

src/bijectors/coupling_layer.jl
torfjelde: See my comment below

src/bijectors/coupling_layer.jl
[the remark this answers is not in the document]
torfjelde: It isn't absolutely necessary, but otherwise we need `cl.θ` to be able to handle both `AbstractVector` and `AbstractMatrix` i.e. more work for the user.

Worth mentioning that this was written _before_ we introduced dimensionality, so these might not be necessary anymore!:)

src/bijectors/coupling_layer.jl
torfjelde: Is this actually more efficient when it comes to matrix multiplication and such? Thinking about both CPU and GPU perf here.

torfjelde: Alternatively, for convenience we can do
```julia
function PartitionMask(
    n::Int,
    indices_1::AbstractVector{Int},
    indices_2::AbstractVector{Int},
    indices_3::AbstractVector{Int},
   T::Type = Float64
)
    A_1 = spzeros(T, n, length(indices_1));
```

src/bijectors/coupling_layer.jl
xukai92: Why would we expect to return an array of a different type after partitioning?

src/bijectors/coupling_layer.jl
xukai92: I'm not sure about the perfomrance. This is attempting to solve the type issue.

Your suggestion works only if users manually input the type. I don't think it's ideal to have this. Why partitioning cannot preserve the array type it works on?

src/bijectors/coupling_layer.jl
torfjelde: In the sense that `PartitionMask` also has a parameter in the type, we leave type-promotion to the multiplication.

I see what you're saying though, but I don't think we're going to be able to do that in a consistent way. What if we have `PartitionMask{SparseArray}` and `x::SparseVector`, how do we ensure that we also return a `SparseVector`? Yeah, we can implement that, but we're not going to be able to cover all these different combinations, right?

xukai92: > Yeah, we can implement that, but we're not going to be able to cover all these different combinations, right?

I really mean the `eltype` should not be changed.

xukai92: If we only doing partitioning, it is expected.

torfjelde: > I really mean the eltype should not be changed.
Yeah, but I honestly don't see how we can do that consistently :confused: 

And I don't think it's such a huge problem as this is something that generally comes up with transformations using matrix multiplication and the type is right there in the definition of `PartitionArray{A}`. It' similar to how if you do `Shift{Float64}  and use that on a `Float32`, things will be converted.

I _do_ however agree that with partitioning it's not so natural to think that this will make a difference, but, again, I don't see how we can consistently ensure that we don't change the `eltype` when we're using matrix multiplication under the hood :confused: 

To do that, I think we'll have to drop the use of matrix multiplication, but that also means that we drop GPU support, right? Could potentially define specific ones when using `CuArray`?

devmotion: > I do however agree that with partitioning it's not so natural to think that this will make a difference, but, again, I don't see how we can consistently ensure that we don't change the eltype when we're using matrix multiplication under the hood

I assume you should be fine if the element type of the mask is `Bool` - e.g., `3f0 * true isa Float32` and `3.0 * true isa Float64`.

xukai92: > Yeah, but I honestly don't see how we can do that consistently 😕

> I assume you should be fine if the element type of the mask is `Bool`

Yes that's why I suggest changing mask to `Bool` in other comments.

torfjelde: > I assume you should be fine if the element type of the mask is Bool - e.g., 3f0 * true isa Float32 and 3.0 * true isa Float64.

Ah, now I understand why you provided the other suggested change @xukai92; sorry about that! For a simple example I did see a slight perf difference, but probably not enough for us to not make the change. Do you know if it will work nicely on GPU?

xukai92: Yes it works with GPU but you would need `A_1`, `A_2` and `A_3` also on GPU, e.g. by doing so
```julia
import Flux
Flux.@functor PartitionMask
mask = PartitionMask(...) |> Flux.gpu
```
Of course if we make the whole coupling layer a functor it would automatically happen when do `Flux.gpu` on the whole layer.

torfjelde: Ah, but what you suggest won't result in a sparse matrix :confused: I just blew the memory of my GPU out trying the example of GPU usage I have in another comment from above. And converting directly do sparse arrays on GPU doesn't seem to work:
```julia
julia> A_1
100×10 SparseMatrixCSC{Bool,Int64} with 10 stored entries:
  [1 ,  1]  =  1
  [2 ,  2]  =  1
  [3 ,  3]  =  1
  [4 ,  4]  =  1
  [5 ,  5]  =  1
  [6 ,  6]  =  1
  [7 ,  7]  =  1
  [8 ,  8]  =  1
  [9 ,  9]  =  1
  [10, 10]  =  1

julia> CUSPARSE.CuSparseMatrixCSC(A_1)
ERROR: MethodError: no method matching CuArrays.CUSPARSE.CuSparseMatrixCSC(::Array{Int64,1}, ::Array{Int64,1}, ::Array{Bool,1}, ::Tuple{Int64,Int64})
Closest candidates are:
  CuArrays.CUSPARSE.CuSparseMatrixCSC(::Array{Ti,1}, ::Array{Ti,1}, ::Array{T,1}, ::Tuple{Int64,Int64}) where {T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64}, Ti<:Integer} at /home/tor/.julia/packages/CuArrays/1njKF/src/sparse/array.jl:207
  CuArrays.CUSPARSE.CuSparseMatrixCSC(::CuArray{Ti,N,P} where P where N, ::CuArray{Ti,N,P} where P where N, ::CuArray{T,N,P} where P where N, ::Tuple{Int64,Int64}) where {T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64}, Ti<:Integer} at /home/tor/.julia/packages/CuArrays/1njKF/src/sparse/array.jl:208
  CuArrays.CUSPARSE.CuSparseMatrixCSC(::CuArray{Ti,N,P} where P where N, ::CuArray{Ti,N,P} where P where N, ::CuArray{T,N,P} where P where N, ::Any, ::Tuple{Int64,Int64}) where {T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64}, Ti<:Integer} at /home/tor/.julia/packages/CuArrays/1njKF/src/sparse/array.jl:209
Stacktrace:
 [1] CuArrays.CUSPARSE.CuSparseMatrixCSC(::SparseMatrixCSC{Bool,Int64}) at /home/tor/.julia/packages/CuArrays/1njKF/src/sparse/array.jl:219
 [2] top-level scope at REPL[56]:1
```

I guess we could be using `Bool` if on CPU and `Float32` on GPU? I remember there being an error where if the input was `CuArray{Float64}` and the partition masks were `CuArray{Float32}`, it would just fail silently, i.e. returned `nothing`. So I guess in the GPU case we're going to require the input and the partition mask to have the _same_  `eltype` anyways :confused:

torfjelde: What do you think? Should I make it `Bool` for now and then we can deal with the pains of GPU later? In the mean-time, it's of course possible to provide the matrices directly and so it's _possible_ to use GPU with sparse-arrays, but slightly inconvenient.

xukai92: Let's make it `Bool` for now and to resolve the bad memory usage due to dense matrix issue on GPU later.

torfjelde: Aight!

xukai92:
Can you remind me what's left to be done for this PR?

torfjelde:
The design is the main bit. I'm not quite happy with how it's currently done, and would like to re-think it.  Essentially what's being mentioned in the PR as "up-for-discussion". Would love to hear thoughts!

torfjelde:
The ReverseDiff.jl error is fixed in https://github.com/JuliaDiff/ReverseDiff.jl/pull/148, hence the tests should pass once we have a new release of ReverseDiff.jl.

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,273 @@
...
+function logabsdetjac(cl::Coupling, x::AbstractMatrix)
+    r = [logabsdetjac(cl, x[:, i]) for i = 1:size(x, 2)]
+
+    # FIXME: this really needs to be handled in a better way
+    # We need to return a `TrackedArray`
+    if Tracker.istracked(r[1])
+        return Tracker.collect(r)
+    else
+        return r
+    end
```
torfjelde, line 272: This is leftovers from me trying to ensure that everything is `TrackedArray` rather than `Array{<:TrackedReal}`. I don't think this is longer something we strive for in Bijectors.jl (e.g. see discussion in #99 ). Sooo maybe just remove this?

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,273 @@
...
+)
+
+"""
+    PartitionMask(n::Int, indices)
+
+Assumes you want to _split_ the vector, where `indices` refer to the 
+parts of the vector you want to apply the bijector to.
+"""
+function PartitionMask(n::Int, indices; sp_type = Bool)
+    indices_2 = [i for i in 1:n if i ∉ indices]
```
devmotion, line 122:
```suggestion
    indices_2 = setdiff(1:n, indices)
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,273 @@
...
+    kwargs...
+)
+
+"""
+    PartitionMask(n::Int, indices)
+
+Assumes you want to _split_ the vector, where `indices` refer to the 
+parts of the vector you want to apply the bijector to.
+"""
+function PartitionMask(n::Int, indices; sp_type = Bool)
```
devmotion, line 121: I think it would be more natural to make `sp_type` a type parameter (it's the element type of `A` anyways) and then define
```suggestion
PartitionMask(n::Int, indices) = PartitionMask{Bool}(n, indices)
function PartitionMask{T}(n::Int, indices) where T
```
I'm wondering though why you do not want to hardcode `Bool` here? `Bool` is the number type that affects promotion the least, so in contrast to integers etc (usually) you do not have to be worried about promoting inputs.

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,273 @@
...
+) = PartitionMask(n, indices_1, indices_2, nothing; kwargs...)
+
+PartitionMask(
+    n::Int,
+    indices_1::AbstractVector{Int},
+    indices_2::AbstractVector{Int},
+    indices_3::Nothing;
+    kwargs...
+) = PartitionMask(
+    n, indices_1, indices_2, [i for i in 1:n if i ∉ (indices_1 ∪ indices_2)];
```
devmotion, line 100:
```suggestion
    n, indices_1, indices_2, setdiff(1:n, indices_1, indices_2);
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,273 @@
...
+)
+
+PartitionMask(
+    n::Int,
+    indices_1::AbstractVector{Int},
+    indices_2::Nothing,
+    indices_3::AbstractVector{Int};
+    kwargs...
+) = PartitionMask(
+    n, indices_1, [i for i in 1:n if i ∉ (indices_1 ∪ indices_3)], indices_3;
```
devmotion, line 111:
```suggestion
    n, indices_1, setdiff(1:n, indices_1, indices_3), indices_3;
```

src/bijectors/coupling.jl
torfjelde: > I think it would be more natural to make sp_type a type parameter 

Yeah, you're completely right :+1: 

> I'm wondering though why you do not want to hardcode Bool here? 

https://github.com/TuringLang/Bijectors.jl/pull/48#discussion_r376639755 

In addition, it's more generality at essentially zero cost, so to me it seems like something we'd do just in case someone really needs to be some other type?

torfjelde: > I think it would be more natural to make sp_type a type parameter

Should we do `PartitionMask{T, A<:AbstractArray{T}}` then?

devmotion: Yes, that's what I had in mind.

torfjelde: Done :+1:

torfjelde:
Tests are passing now :tada:

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+    end
+
+    for (i, idx) in enumerate(indices_2)
+        A_2[idx, i] = one(T)
+    end
+
+    return PartitionMask(A_1, A_2, spzeros(T, n, 0))
+end
+function PartitionMask{T}(x::AbstractVector, indices) where {T}
+    return PartitionMask(length(x), indices)
```
devmotion, line 136: This is not tested, but I assume it should be
```suggestion
    return PartitionMask{T}(length(x), indices)
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+struct PartitionMask{T, A}
+    A_1::A
+    A_2::A
+    A_3::A
+
+    # Only make it possible to construct using matrices
+    PartitionMask(A_1::A, A_2::A, A_3::A) where {T<:Real, A <: AbstractMatrix{T}} = new{T, A}(A_1, A_2, A_3)
+end
+
+PartitionMask(args...; kwargs...) = PartitionMask{Bool}(args...; kwargs...)
```
devmotion, line 60: It seems there are no keyword arguments anymore?
```suggestion
PartitionMask(args...) = PartitionMask{Bool}(args...)
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+    indices_1::AbstractVector{Int},
+    indices_2::AbstractVector{Int};
+) where {T} = PartitionMask{T}(n, indices_1, indices_2, nothing)
+
+PartitionMask{T}(
+    n::Int,
+    indices_1::AbstractVector{Int},
+    indices_2::AbstractVector{Int},
+    indices_3::Nothing;
+    kwargs...
```
devmotion, line 98:
```suggestion
    indices_3::Nothing,
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+) where {T} = PartitionMask{T}(n, indices_1, indices_2, setdiff(1:n, indices_1, indices_2))
+
+PartitionMask{T}(
+    n::Int,
+    indices_1::AbstractVector{Int},
+    indices_2::Nothing,
+    indices_3::AbstractVector{Int};
+    kwargs...
```
devmotion, line 106: Not tested, but I guess:
```suggestion
    indices_3::AbstractVector{Int},
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+# References
+[1] Kobyzev, I., Prince, S., & Brubaker, M. A., Normalizing flows: introduction and ideas, CoRR, (),  (2019). 
+"""
+struct Coupling{F, M} <: Bijector{1} where {F, M <: PartitionMask}
+    θ::F
+    mask::M
+end
+
+function Coupling(θ, n::Int)
+    idx = Int(floor(n / 2))
```
devmotion, line 201: It is better to use
```suggestion
    idx = div(n, 2)
```
(or the corresponding infix operator)

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+    return Coupling(θ, PartitionMask(n, 1:idx))
+end
+
+function Coupling(cl::Coupling{B}, mask::PartitionMask) where {B}
```
devmotion, line 205:
```suggestion
function Coupling(cl::Coupling, mask::PartitionMask)
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,267 @@
...
+    x_1, x_2, x_3 = partition(cl.mask, x)
+    b = cl.θ(x_2)
+
+    # `B` might be 0-dim in which case it will treat `x_1` as a batch
+    # therefore we sum to ensure such a thing does not happen
+    return sum(logabsdetjac(b, x_1))
+end
+
+function logabsdetjac(cl::Coupling, x::AbstractMatrix)
+    r = [logabsdetjac(cl, x[:, i]) for i = 1:size(x, 2)]
```
devmotion, line 258: Maybe use views?

torfjelde:
```suggestion
    r = [logabsdetjac(cl, view(x, :, i)) for i = 1:size(x, 2)]
```
?

src/bijectors/coupling.jl
torfjelde: Thoughts on this @devmotion ?

src/bijectors/coupling.jl
devmotion: Yep, or
```suggestion
    r = [logabsdetjac(cl, x[:, i]) for i in axes(x, 2)]
```

torfjelde: Nice :+1: 

Btw, do you have any thoughts on this: https://github.com/TuringLang/Bijectors.jl/pull/48#discussion_r488610184?

src/bijectors/coupling.jl
devmotion: Hmm I think this is ugly, but I know that arrays of TrackedReal are not efficient. However, I guess e.g. if in most cases you just compute the sum of the values then it should be fine?

So if Bijectors in general is not worried about arrays of TrackedReal anymore, then I would be fine without special cases here.

torfjelde: Yeah, I remember @mohamed82008 almost preferred having `Array{<:TrackedReal}` because apparently `sum` is faster on this. Though it used to cause a lot of issues with compositions due to the inconsistency, but that was fixed on Tracker's end. We make take a perf hit with compositions though, as I'm guessing 'TrackedArray + TrackedArray' is more efficient than `TrackedArray + Array{<:TrackedReal}` 😕 

I don't know tbh. But this is such an ugly fix anyways, that we might as well just remove it for now 👍

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,265 @@
...
+function logabsdetjac(cl::Coupling, x::AbstractMatrix)
+    r = [logabsdetjac(cl, view(x, :, i)) for i in axes(x, 2)]
+
+    # FIXME: this really needs to be handled in a better way
+    # We need to return a `TrackedArray`
+    if Tracker.istracked(r[1])
+        return Tracker.collect(r)
+    else
+        return r
+    end
```
torfjelde, line 264:
```suggestion
    return [logabsdetjac(cl, view(x, :, i)) for i in axes(x, 2)]
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,257 @@
...
+        A_1[idx, i] = one(T)
+    end
+
+    for (i, idx) in enumerate(indices_2)
+        A_2[idx, i] = one(T)
+    end
+
+    for (i, idx) in enumerate(indices_3)
+        A_3[idx, i] = one(T)
+    end
```
devmotion, line 82: I noticed that you can just
```suggestion
    A_1 = sparse(indices_1, 1:length(indices_1), one(T), n, length(indices_1))
    A_1 = sparse(indices_2, 1:length(indices_2), one(T), n, length(indices_2))
    A_1 = sparse(indices_3, 1:length(indices_3), one(T), n, length(indices_3))
```

src/bijectors/coupling.jl
```diff
@@ -0,0 +1,257 @@
...
+    # Like doing:
+    #    A[1, 1] = 1.0
+    #    A[3, 2] = 1.0
+    for (i, idx) in enumerate(indices)
+        A_1[idx, i] = one(T)
+    end
+
+    for (i, idx) in enumerate(indices_2)
+        A_2[idx, i] = one(T)
+    end
```
devmotion, line 129:
```suggestion
    # sparse arrays <3
    A_1 = sparse(indices, 1:length(indices), one(T), n, length(indices))
    A_2 = sparse(indices_2, 1:length(indices_2), one(T), n, length(indices_2))
```

src/bijectors/coupling.jl
torfjelde: Woah, that's much, much better! Thanks!

devmotion (approved):
LGTM, just bump the version, then it's ready to go!

torfjelde:
Awesome, and thanks for the great feedback!

Issue #46, opened 2019-09-15
`Composed` improvements


torfjelde:
I've re-visted some ideas I had earlier on but didn't have quite the knowledge nor time to explore. As a result I think I've come up with some really nice improvements to the `Composed` bijector.

The  changes suggested here already exists on a local branch of mine and tests are passing without issues. So if people seem happy with this I'll make a PR asap.

# Nested compositions is bleh
We're currently not doing anything to "flatten" the compositions, e.g.
```julia
julia> using Bijectors

julia> b = Bijectors.Exp()
Bijectors.Exp()                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                               
julia> map(typeof, (b ∘ b ∘ b).ts)
(Bijectors.Exp, Composed{Tuple{Bijectors.Exp,Bijectors.Exp}})
```
Notice that it we get _nested_ compositions here.

Personally I'd like `∘` to encourage a _flat_ structure, so we instead get:
```julia
julia> map(typeof, (b ∘ b ∘ b).ts)
(Bijectors.Exp, Bijectors.Exp, Bijectors.Exp)
```
The above was accomplished by simply introducing the following lines:
```julia
∘(b1::Bijector, b2::Bijector) = composel(b2, b1)  # <= already exists
∘(b1::Composed, b2::Bijector) = composel(b2, b1.ts...)       # new 
∘(b1::Bijector, b2::Composed) = composel(b2.ts..., b1)       # new
∘(b1::Composed, b2::Composed) = composel(b2.ts..., b1.ts...) # new
```
If the user really wants to use _nested_ compositions rather than this flat structure, they can call `composel` and/or `composer` explicitly. I struggle to see a case where you'd like to group bijectors together but wouldn't already have access to this grouping outside of the composition _but_ I'd still like to allow both possibilities. This solution does exactly that.

Other than just being "nice", it has two advantages over nesting:
1. Dispatching on different compositions is easier since you only need to consider `Composed{Tuple{B1, B2, B3}}` rather than all combinations of nestedness.
2. Helps with performance

# Recursive implementations ain't so good
Currently we're using a recursive implementation for the following methods for `Composed`:
- `(cb::Composed)(x)`
- `forward(cb::Composed, x)`
- `logabsdetjac(cb::Composed, x)`

[...]

Issue #40, opened 2019-08-31
Type-stability of `bijector(d::Distribution)` not complete


torfjelde:
I've made an effort to make things in the interface type-stable, but there's one `bijector(d::Distribution)` call that is currently not type-stable: `Truncated`.

For `bijector(d::Truncated)` we need to decide which bijectors to compose depending on whether or not it's only lower-bounded, upper-bounded or both.

Also, at the moment, some other univariate distributions (all the ones defined in union `TransformDistributions`) also have unstable bijector calls because they are also using the implementation as `Truncated`, e.g. `ArcSine`, but these we can just implement on a case-by-case basis to get type-stability in all cases except Truncated.

In practice, I don't think type-stability of `bijector(d::Distribution)` is that important, as one will usually construct the distributions at the very beginning of the computation. Even so, for the cases where this is not the case it would be nice to have type-stability.

Also, in the case of `Truncated{<:Distribution}` the union is of size 3 which can possibly be dealt with by union-splitting since default split-size is 4.

```julia
julia> for d in uni_dists
           c = @code_typed bijector(d)
           if (c.second isa Union) || (c.second == Any)
               println("-----------------")
               println(d, ": ", c.second)
               println("(", minimum(d), ", ", maximum(d), ")")
               println(bijector(d))
           end
       end
-----------------
Arcsine{Float64}(a=2.0, b=4.0): Union{Identity, Logit{Float64}, Composed}
(2.0, 4.0)
Bijectors.Logit{Float64}(2.0, 4.0)
-----------------
Biweight{Float64}(μ=0.0, σ=1.0): Union{Identity, Logit{Float64}, Composed}
(-1.0, 1.0)
Bijectors.Logit{Float64}(-1.0, 1.0)
-----------------
Cosine{Float64}(μ=0.0, σ=1.0): Union{Identity, Logit{Float64}, Composed}
(-1.0, 1.0)
Bijectors.Logit{Float64}(-1.0, 1.0)
-----------------
Epanechnikov{Float64}(μ=0.0, σ=1.0): Union{Identity, Logit{Float64}, Composed}
(-1.0, 1.0)
Bijectors.Logit{Float64}(-1.0, 1.0)
-----------------
Levy{Float64}(μ=0.0, σ=1.0): Union{Identity, Composed{Tuple{Shift{Float64},Log}}}
(0.0, Inf)
[...]

PR #52, opened 2019-11-11
Improvements to `Composed` and  `TruncatedBijector` fixing #40

torfjelde:
# Overview
This PR introduces most of the _performance_ improvements to `Composed` discussed in #46 and the `TruncatedBijector` discussed in #40 making `bijector(d::Distribution)` type-stable and providing a good "default" bijector.

In addition to the above, `Test.@inferred` is now used extensively in `test/interface.jl` to try and catch instability for the `<:Tuple` versions of `Composed` and `Stacked`.

Finally the project structure has been changed a bit (for the better). Implementations of `Bijector` are now found in the `src/bijectors` directory, e.g. `src/bijectors/scale.jl`. This makes it much easier for people unfamiliar with the code to contribute, plus easier to reason about in general.

# Changes
## `Composed`
From a user-perspective, the major change is that `∘` will now "flatten" the composition structure in the sense that
```julia
julia> using Bijectors; using Bijectors: Exp

julia> b = Exp()
Exp{0}()

julia> cb1 = b ∘ b;

julia> cb2 = b ∘ b;

julia> (cb1 ∘ cb2).ts # <= different
(Exp{0}(), Exp{0}(), Exp{0}(), Exp{0}())

julia> (cb1 ∘ cb2).ts isa NTuple{4, Exp{0}}
true
```
while in master it looks like
```julia
julia> using Bijectors; using Bijectors: Exp
[ Info: Recompiling stale cache file /home/tor/.julia/compiled/v1.1/Bijectors/39uFz.ji for Bijectors [76274a88-744f-5084-9051-94815aaf08c4]

julia> b = Exp()
Exp{0}()

julia> cb1 = b ∘ b;

julia> cb2 = b ∘ b;

julia> (cb1 ∘ cb2).ts # <= different
(Composed{Tuple{Exp{0},Exp{0}},0}((Exp{0}(), Exp{0}())), Composed{Tuple{Exp{0},Exp{0}},0}((Exp{0}(), Exp{0}())))

julia> (cb1 ∘ cb2).ts isa Tuple{Composed, Composed}
true
```
Note that this is _only_ the case for `∘`. In some cases it might be desirable to keep certain operations together, e.g. `Composed{Tuple{Scale, Shift}}}` represents an affine transformation. Therefore `composel` and `composer` will _not_ do this "flattening", but instead preserve the structure of the compositions.

The above is also described in the docstring of `Composed`.


src/bijectors/adbijector.jl
```diff
@@ -0,0 +1,53 @@
+"""
+Abstract type for a `Bijector{N}` making use of auto-differentation (AD) to
+implement `jacobian` and, by impliciation, `logabsdetjac`.
+"""
+abstract type ADBijector{AD, N} <: Bijector{N} end
+
+# AD implementations
+function jacobian(b::ADBijector{<:ForwardDiffAD}, x::Real)
```
xukai92: Can we merge the forward and inverse functions?

src/bijectors/adbijector.jl
```diff
@@ -0,0 +1,53 @@
...
+end
+
+function jacobian(b::ADBijector{<:TrackerAD}, x::Real)
+    return Tracker.data(Tracker.gradient(b, x)[1])
+end
+function jacobian(b::Inversed{<:ADBijector{<:TrackerAD}}, y::Real)
+    return Tracker.data(Tracker.gradient(b, y)[1])
+end
+function jacobian(b::ADBijector{<:TrackerAD}, x::AbstractVector{<:Real})
+    # We extract `data` so that we don't returne a `Tracked` type
```
xukai92: Typo. `returne` -> `return`

src/bijectors/scale.jl
```diff
@@ -0,0 +1,80 @@
+struct Scale{T, N} <: Bijector{N}
+    a::T
+end
+
+Scale(a::T; dim::Val{D} = Val(0)) where {T<:Real, D} = Scale{T, D}(a)
+Scale(a::A; dim::Val{D} = Val(N)) where {T, D, N, A<:AbstractArray{T, N}} = Scale{A, D}(a)
+
+(b::Scale)(x) = b.a .* x
```
xukai92: Is there a particular reason we sometimes use `@.` and sometimes do the dot operations explictly.

src/bijectors/truncated.jl
```diff
@@ -0,0 +1,92 @@
+#######################################################
+# Constrained to unconstrained distribution bijectors #
+#######################################################
+struct TruncatedBijector{T} <: Bijector{0}
```
xukai92: Does the parameter here make things type stable?

xukai92:
Looks good to me overall!

src/bijectors/adbijector.jl
torfjelde: You mean `Union{Inversed{B}, B}` where `B<:ADBijector{<:ForwardDiff}`? If so, that seems like a good idea to me:)

src/bijectors/scale.jl
torfjelde: I don't have a specific style that I follow for this. It's more a matter of how many dots I need to write out:)
I usually go with `@.` if I have more than 2 operations I want to fuse

src/bijectors/truncated.jl
torfjelde: Depends on what the alternative is I suppose, but for example
```julia
struct TruncatedBijector <: Bijector{0}
    lb
    ub
end
```
would at least not be type-stable. But you could of course fix the types to be, say, `Float64`, but that seems unnecessary

src/bijectors/adbijector.jl
xukai92: Yes.

xukai92:
@torfjelde Can I merge this now?

torfjelde:
> @torfjelde Can I merge this now?

Yup! Though maybe I should bump the version number? I guess this would _technically_ be a breaking change since the circ-operation now has different behavior. Should I bump the minor- or patch-version?

src/bijectors/composed.jl
```diff
@@ -0,0 +1,190 @@
...
+    res = x
+    for b ∈ cb.ts
+        res = b(res)
+    end
+
+    return res
+end
+
+# recursive implementation like this allows type-inference
+_transform(x, b1::Bijector, b2::Bijector) = b2(b1(x))
```
torfjelde: Implement this without recursion using `@generated` like we've done for `forward`.

torfjelde:
Should be good now @kai!:)

xukai92:
Thanks for the great work!

PR #59, opened 2019-12-28
Do not use `Val` as type parameter

devmotion:
I guess the type parameter `Val{proj}` of `SimplexBijector` can just be replaced with `proj`. That seems a bit simpler to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,13 +335,13 @@ dim: 4
 
 julia> # Construct the transform
        bs = bijector.(dists)     # constrained-to-unconstrained bijectors for dists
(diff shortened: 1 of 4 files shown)


devmotion:
The test error is unrelated and fixed by https://github.com/TuringLang/Bijectors.jl/pull/60.

mohamed82008:
@torfjelde well most of the time we are only doing the `proj = true` for sampling so the runtime check is not necessary here. In general, runtime condition checking in big functions is negligible. But for small functions in a loop, the runtime condition can inhibit some optimizations like SIMD if constant propagation (not the most reliable in my experience) doesn't kick in.

PR #60, opened 2019-12-28
Update dependencies and fix deprecations

devmotion:
This PR updates the dependencies of Bijectors, adds package compatibilities (I took lower bounds from Turing), and fixes some deprecations.


devmotion:
The updates fix the current errors due to Pkg dependencies on Julia nightly (see, e.g., https://ci.appveyor.com/project/TuringLang/bijectors-jl/builds/29787947/job/tegs9jryfyla563a) but expose a type inference problem on Julia nightly.

src/bijectors/stacked.jl
```diff
@@ -80,12 +80,15 @@ function (sb::Stacked{<:AbstractArray, N})(x::AbstractVector{<:Real}) where {N}
     return y
 end
 
-# HACK: `reshape` to get around the fact that `hcat` isn't type-stable
-(sb::Stacked)(x::AbstractMatrix{<:Real}) = reshape(foldl(hcat, [sb(x[:, i]) for i = 1:size(x, 2)]), size(x)...)
+@views function (sb::Stacked)(x::AbstractMatrix{<:Real})
+    init = reshape(sb(x[:, 1]), :, 1)
+    return mapreduce(i -> sb(x[:, i]), hcat, 2:size(x, 2); init = init)
```
mohamed82008: No need to define `init` separately. This is equivalent to using `1:size(x, 2)` and dropping the `init` kwarg.

mohamed82008:
Oh and thanks for the PR btw :)

src/bijectors/stacked.jl
mohamed82008: Hmm actually I double checked and your implementation here is type stable but the compiler is struggling to infer my suggestion.

devmotion: Yes, this change is only for fixing the type inference problem.

devmotion:
I'm in favour of removing the Manifest file and not depending on ForwardDiff and Tracker as well :+1: Actually I was wondering about the Manifest file and the Requires approach, but since I didn't know what's the general opinion about it I just stuck to the current approach (with Manifest file and ForwardDiff and Tracker as hard dependencies - even though it might seem otherwise due to the dependency on Requires and the optional block in `Bijectors.jl` ForwardDiff and Tracker are already dependencies and loaded in `interface.jl`).

cpfiffer:
I've been removing the Manifest everywhere else -- I think it's bad form to leave it in the repo.

src/bijectors/scale.jl
```diff
@@ -24,62 +24,4 @@ _logabsdetjac_scale(a::Real, x::AbstractMatrix, ::Val{1}) = fill(log(abs(a)) * s
 _logabsdetjac_scale(a::AbstractVector, x::AbstractVector, ::Val{1}) = sum(log.(abs.(a)))
 _logabsdetjac_scale(a::AbstractVector, x::AbstractMatrix, ::Val{1}) = fill(sum(log.(abs.(a))), size(x, 2))
 _logabsdetjac_scale(a::AbstractMatrix, x::AbstractVector, ::Val{1}) = log(abs(det(a)))
-_logabsdetjac_scale(a::AbstractMatrix, x::AbstractMatrix{T}, ::Val{1}) where {T} = log(abs(det(a))) * ones(T, size(x, 2))
```
torfjelde: Is there a reason why this line was removed?

devmotion: It's not actually removed, somehow I just removed the newline character at the end of the line.

torfjelde: Ah, np then:)

mohamed82008:
I like the `compat` idea.

devmotion:
I like this idea as well, so I moved the optional functionality to a `compat` directory.

PR #61, opened 2019-12-29
Fix test and bijector of MvLogNormal

devmotion:
This PR fixes an incorrect test (instead of testing different distributions one distribution was fixed and tested multiple times). This fix exposed a bug in the definition of `bijector(d::MvLogNormal)`, for which a fix is included as well.


devmotion:
The test error is unrelated and fixed by https://github.com/TuringLang/Bijectors.jl/pull/60.

PR #63, opened 2020-01-01
Fix optional dependencies

devmotion:
Turing's CI tests show that the optional dependencies on ForwardDiff and Tracker are not correctly implemented. As noted in the [documentation of Requires](https://github.com/MikeInnes/Requires.jl/blob/master/README.md#requiresjl), one has to use `using/import .ForwardDiff` and `using/import .Tracker` in the included files instead of `using/import ForwardDiff` and `using/import Tracker`.

Unfortunately, just adding a dot in front of every occurrence of `ForwardDiff` and `Tracker` does not work since ["relative-import qualifiers are only valid in `using` and `import` statements"](https://docs.julialang.org/en/v1/manual/modules/#Relative-and-absolute-module-paths-1) and hence it is not possible to call `.ForwardDiff.jacobian` and `.Tracker.jacobian` - instead the `jacobian` functions have to be imported! However, this leads to namespace problems since Bijectors, ForwardDiff and Tracker all define their own `jacobian` function.

The PR solves this problem by defining `jacobian(b, x) = _jacobian(b, x)` in Bijectors and implementing separate modules for ForwardDiff and Tracker compatibility in which `_jacobian` is imported from Bijectors and `jacobian` from ForwardDiff and Tracker, respectively.

Moreover, since the convenience macro `Tracker.@grad` [hardcodes a call to `Tracker._forward`](https://github.com/FluxML/Tracker.jl/blob/128c29d46d51e182319bbdbae066dc279bdee2af/src/Tracker.jl#L62) it can't be used and instead `_forward` has to be implemented directly.

Running Turing's tests with this PR shows that indeed the approach in this PR fixes the current implementation.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,13 +14,11 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 
 [compat]
 Distributions = "0.21.11"
-ForwardDiff = "0.10.3"
 MappedArrays = "0.2.2"
 Reexport = "0.2"
 Requires = "1"
 Roots = "0.8.4"
 StatsFuns = "0.8, 0.9.3"
-Tracker = "0.2.3"
 julia = "1"
 
 [extras]
--- a/src/bijectors/adbijector.jl
+++ b/src/bijectors/adbijector.jl
@@ -13,6 +13,7 @@ end
 
 # concrete implementations with optional dependencies ForwardDiff and Tracker
 function jacobian end
+const _jacobian = jacobian
 
 # TODO: allow batch-computation, especially for univariate case?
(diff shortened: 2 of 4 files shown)


mohamed82008:
Why do we need `.ForwardDiff.jacobian`? When you do `using/import ForwardDiff`, the identifier `ForwardDiff` gets loaded to the namespace, so you can do `ForwardDiff.jacobian` no problem.

devmotion:
Yes, you're right. Actually I'm not sure why I ended up with this implementation, I guess everything else didn't work. I put together a new and much simpler [PR](https://github.com/TuringLang/Bijectors.jl/pull/64).

PR #64, opened 2020-01-05
Fix optional dependencies

devmotion:
Fixes the warning message in https://github.com/TuringLang/Turing.jl/pull/1052. Supersedes https://github.com/TuringLang/Bijectors.jl/pull/63.


mohamed82008:
Thanks for this PR @devmotion!

PR #65, opened 2020-01-07
Remove duplicates in SimplexBijector and SimplexDistribution

devmotion:
This PR removes some code duplications in the implementation of SimplexBijector and SimplexDistribution.

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -171,39 +171,15 @@ end
 ###########
 
 const SimplexDistribution = Union{Dirichlet}
-function _clamp(x::T, dist::SimplexDistribution) where T
-    bounds = (zero(T), one(T))
-    clamped_x = clamp(x, bounds...)
-    DEBUG && @debug "x = $x, bounds = $bounds, clamped_x = $clamped_x"
-    return clamped_x
-end
+
+_clamp(x, ::SimplexDistribution) = _clamp(x, SimplexBijector())
 
 function link(
     d::SimplexDistribution,
-    x::AbstractVector{T},
-    ::Type{Val{proj}} = Val{true}
-) where {T<:Real, proj}
-    y, K = similar(x), length(x)
-
-    ϵ = _eps(T)
-    sum_tmp = zero(T)
-    @inbounds z = x[1] * (one(T) - 2ϵ) + ϵ # z ∈ [ϵ, 1-ϵ]
-    @inbounds y[1] = StatsFuns.logit(z) + log(T(K - 1))
-    @inbounds @simd for k in 2:(K - 1)
-        sum_tmp += x[k - 1]
-        # z ∈ [ϵ, 1-ϵ]
-        # x[k] = 0 && sum_tmp = 1 -> z ≈ 1
-        z = (x[k] + ϵ)*(one(T) - 2ϵ)/((one(T) + ϵ) - sum_tmp)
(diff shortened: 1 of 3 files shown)


src/transformed_distribution.jl
```diff
@@ -39,7 +39,7 @@ bijector(d::Normal) = Identity{0}()
 bijector(d::MvNormal) = Identity{1}()
 bijector(d::PositiveDistribution) = Log{0}()
 bijector(d::MvLogNormal) = Log{1}()
-bijector(d::SimplexDistribution) = SimplexBijector{true}()
+bijector(d::SimplexDistribution, ::Type{Val{proj}} = Val{true}) where proj = SimplexBijector{proj}()
```
torfjelde: IMHO, it's slightly nicer to hardcode usage of `SimplexBijector` with the correct `proj` value into the `link` and `invlink` definitions rather than changing the definition of `bijector` to accomodate the current definitions of `link` and `invlink`. Ideally, we'll just completely phase out the use of `link` and `invlink` at some point, and so I feel it's better to keep the `<: Bijector` interface consistent.

Therefore I'd suggest reverting this change back.

src/Bijectors.jl
```diff
@@ -171,39 +171,15 @@ end
...
-        y[K] = zero(T)
-    else
-        y[K] = one(T) - sum_tmp - x[K]
-    end
-
-    return y
+    x::AbstractVecOrMat{<:Real},
+    proj = Val{true}
+)
+    return bijector(d, proj)(x)
```
torfjelde: Following my other comment, I suggested doing the following:

```julia
function link(
    d::SimplexDistribution,
    x::AbstractVecOrMat{<:Real},
    ::Type{Val{proj}} = Val{true}
) where {proj}
    return SimplexBijector{proj}()(x)
end
```

src/Bijectors.jl
```diff
@@ -239,58 +215,12 @@ function link_jacobian(
...
-    @inbounds if proj
-        x[K] = _clamp(one(T) - sum_tmp, d)
-    else
-        x[K] = _clamp(one(T) - sum_tmp - y[K], d)
-    end
-    return x
+    y::AbstractVecOrMat{<:Real},
+    proj = Val{true}
+)
+    return inv(bijector(d, proj))(y)
```
torfjelde: Following my other comment, I suggest the following:
```julia
function invlink(
    d::SimplexDistribution,
    y::AbstractVecOrMat{<:Real},
    ::Type{Val{proj} = Val{true}}
) where {proj}
    return inv(SimplexBijector{proj}())(y)
end
```

src/transformed_distribution.jl
devmotion: I don't have a strong opinion on this, so I just reverted the change. I agree that the change introduced an unfortunate inconsistency in the `bijector` interface. The main motivation for it was to be able to obtain an instance of `SimplexBijector{false}` with `bijector` - that's not possible anymore in the updated version. However, I guess it's fine to just use the default `SimplexBijector{true}()` since users can still obtain the non-standard `SimplexBijector{false}()` without `bijector`.

devmotion:
Is there anything else that you want me to change in this PR? :slightly_smiling_face:

mohamed82008:
Let me review it tonight then I will merge after if all is good, sorry for the delay.

PR #66, opened 2020-01-09
Remove Tracker implementation of Shift bijector

devmotion:
This PR removes the special implementation of `logabsdetjac` of the Shift bijector for TrackedReal's and TrackedVector's.

### DIFF
--- a/src/bijectors/shift.jl
+++ b/src/bijectors/shift.jl
@@ -18,7 +18,16 @@ inv(b::Shift{T, N}) where {T, N} = Shift{T, N}(-b.a)
 # FIXME: implement custom adjoint to ensure we don't get tracking
 logabsdetjac(b::Shift{T, N}, x) where {T, N} = _logabsdetjac_shift(b.a, x, Val(N))
 
-_logabsdetjac_shift(a::Real, x::Real, ::Val{0}) = zero(eltype(x))
-_logabsdetjac_shift(a::Real, x::AbstractVector{T}, ::Val{0}) where {T<:Real} = zeros(T, length(x))
-_logabsdetjac_shift(a::T1, x::AbstractVector{T2}, ::Val{1}) where {T1<:Union{Real, AbstractVector}, T2<:Real} = zero(T2)
-_logabsdetjac_shift(a::T1, x::AbstractMatrix{T2}, ::Val{1}) where {T1<:Union{Real, AbstractVector}, T2<:Real} = zeros(T2, size(x, 2))
+_logabsdetjac_shift(a::Real, x::Real, ::Val{0}) = zero(Base.promote_typeof(a, x))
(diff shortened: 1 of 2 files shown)


src/bijectors/shift.jl
```diff
@@ -18,7 +18,16 @@ inv(b::Shift{T, N}) where {T, N} = Shift{T, N}(-b.a)
...
 
-_logabsdetjac_shift(a::Real, x::Real, ::Val{0}) = zero(eltype(x))
-_logabsdetjac_shift(a::Real, x::AbstractVector{T}, ::Val{0}) where {T<:Real} = zeros(T, length(x))
-_logabsdetjac_shift(a::T1, x::AbstractVector{T2}, ::Val{1}) where {T1<:Union{Real, AbstractVector}, T2<:Real} = zero(T2)
-_logabsdetjac_shift(a::T1, x::AbstractMatrix{T2}, ::Val{1}) where {T1<:Union{Real, AbstractVector}, T2<:Real} = zeros(T2, size(x, 2))
\ No newline at end of file
+_logabsdetjac_shift(a::Real, x::Real, ::Val{0}) = zero(Base.promote_typeof(a, x))
+function _logabsdetjac_shift(a::Real, x::AbstractVector{<:Real}, ::Val{0})
+    T = Base.promote_eltype(a, x)
+    return zeros(T, length(x))
```
torfjelde: The issue with this is that you'll end up with `Array{TrackedReal}` rather than `TrackedArray`. This the reason why I implemented the adjoints by hand in the first place. Back when I implemented this, this kind of issue was caught by this testset: https://github.com/devmotion/Bijectors.jl/blob/0c499b42d556d05cbe7f4ac4a703c8a4636c49b2/test/interface.jl#L315-L335

Buuut it seems to work just fine now, even if there's an issue of conversion. I think `Tracker.extract_grad!(b.a)` used to fail back in the day if the conversion hadn't been done properly. I at least remember introducing the above testset and the adjoints because `Array{TrackedReal}` was causing issues.

Anyways, I do prefer this resulting in a `TrackedArray` rather than array of `TrackedReal`.

torfjelde: Oh, and you can see this by just inserting `@info zeros(T, length(x))` in the above to see that you end up with `Array{TrackedReal}` instead of `TrackedArray`

devmotion: I see, I wasn't aware of this problem! I was satisfied since all tests passed :smile: So I'll just close this PR, since it's not helpful then. However, maybe it would be good to actually add a test that checks that the result is a `TrackedArray`?

torfjelde: No problem:) As I said, this issue used to be caught by the testset mentioned above, but it seems like it's been patched by updates to Tracker.jl.

And yeah, should definitively add some tests for that:)

PR #67, opened 2020-01-09
Bump Distributions version

mohamed82008:


devmotion:
Just a note (since all tests pass): 0.22 introduces a breaking change in the constructors of `MvNormal` (https://github.com/JuliaStats/Distributions.jl/pull/1019) (now all matrices, i.e. even scaled identity matrices, are treated as covariance matrices), so maybe that's good to be aware of.

mohamed82008:
@devmotion Thanks for the note. I don't think we use that though.

PR #76, opened 2020-02-03
Port batch norm

xukai92:


yebai:
is this a replacement for #43?

torfjelde:
> is this a replacement for #43?

Seems like it is! If so, have you addressed my previous comments about the tracking @xukai92 ? Kind of difficult to keep track of with a new PR.

xukai92:
> is this a replacement for #43?

Yes

> Seems like it is! If so, have you addressed my previous comments about the tracking @xukai92 ? Kind of difficult to keep track of with a new PR.

I'm not sure if we still want to make this Tracker compatible. The current implementation is only supposed to work with Zygote due to the use of `@functor`, `trainable()` and `istraining()` (related issue for the last one: https://github.com/FluxML/Flux.jl/issues/909).

PR #77, opened 2020-02-03
Taking bijectors autodiff seriously

mohamed82008:
This PR supersedes #74 and includes the changes in the mt/array_dist_and_multi branch. In this PR:
1. I try to make sure all bijectors are Tracker and Zygote compatible including `PlanarLayer`, `RadialLayer` and `Stacked`.
2. For Tracker support, I try to make sure that we never fallback on arrays of `TrackedReal` and always have `TrackedArray`s returned by the multidimensional bijectors.
3. Type stability of bijectors with `TrackedArray` inputs is maintained for most cases. The `DistributionBijector` is not fully covered yet.
4. Custom adjoints are defined for for the Simplex bijector, inverse bijector and logabsdetjac. This should make the Dirichlet distribution in Turing faster after removing the ForwardDiff-based adjoints defined there, but I didn't test it yet.
5. Support for DistributionAD's `filldist` and `arraydist` distributions is also added here. This means that this PR depends on https://github.com/TuringLang/DistributionsAD.jl/pull/19. 
6. Finally, `Inversed` is renamed to `Inverse`. A major release will be made after this PR is merged anyways so I thought I would squeeze in that breaking change as well.


src/bijectors/planar_layer.jl
```diff
@@ -15,73 +15,93 @@ using Roots # for inverse
 
 # TODO: add docstring
 
-# FIXME: using `TrackedArray` for the parameters, we end up with
-# nested tracked structures; don't want this.
-mutable struct PlanarLayer{T1,T2} <: Bijector{1}
+mutable struct PlanarLayer{T1 <: AbstractVector{<:Real}, T2 <: Real} <: Bijector{1}
```
torfjelde: Style-guide: `{T1<:AbstractVector{<:Real}, T2<:Real}`

src/bijectors/planar_layer.jl
```diff
@@ -15,73 +15,93 @@ using Roots # for inverse
...
+    return u .+ (planar_flow_m(x) - x) .* w ./ sum(abs2, w)   # from A.1
 end
 
 function PlanarLayer(dims::Int, container=Array)
-    w = container(randn(dims, 1))
-    u = container(randn(dims, 1))
-    b = container(randn(1))
+    w = container(randn(dims))
+    u = container(randn(dims))
+    b = randn()
```
torfjelde: I think this came up in the discussion in some other PR, but I believe the reasoning for making `b` an array is so that in the case were one is using Tracker, one can use `TrackedArray` and update `b` by doing elementwise assignment. When this is a `Real` we have to reconstruct the `PlanarLayer`. 

It's kind of ugly to have it as a one-element array, but at the same time I understand the original reasoning as it's quite convenient. I'm not certain what I prefer, personally.

src/bijectors/planar_layer.jl
```diff
@@ -15,73 +15,93 @@ using Roots # for inverse
...
-    psi = ψ(z, flow.w, flow.b)
-    log_det_jacobian = log.(abs.(1.0 .+ psi' * u_hat))          # from eq(12)
-    return (rv=transformed, logabsdetjac=vec(log_det_jacobian)) # from eq(10)
+    psi = ψ(z, flow.w, flow.b) .+ zero(eltype(u_hat))
+    if psi isa AbstractVector
+        T = eltype(psi)
+    else
+        T = typeof(vec(psi))
+    end
+    log_det_jacobian::T = log.(abs.(1.0 .+ psi' * u_hat)) # from eq(12)
```
torfjelde: Why is all this needed? Seems strange that the type of `log_det_jacobian` isn't inferrable, no?

src/bijectors/radial_layer.jl
```diff
@@ -13,78 +13,85 @@ using Roots # for inverse
...
     β::T1
     z_0::T2
 end
 
 function RadialLayer(dims::Int, container=Array)
-    α_ = container(randn(1))
-    β = container(randn(1))
-    z_0 = container(randn(dims, 1))
+    α_ = randn()
+    β = randn()
```
torfjelde: Same comment as for `PlanarLayer`

src/bijectors/radial_layer.jl
```diff
@@ -13,78 +13,85 @@ using Roots # for inverse
...
-    # from A.2
-    α = softplus.(flow.α_)
-
-    # from A.2
-    β_hat = @. -α + softplus(flow.β)
-
-    r = sqrt.(sum((z .- flow.z_0).^2; dims = 1))
-    transformed = @. z + β_hat * h(α, r) * (z - flow.z_0)   # from eq(14)
-    return (transformed=transformed, α=α, β_hat=β_hat, r=r)
+function _transform(flow::RadialLayer, z::AbstractVecOrMat)
```
torfjelde: Can just remove this definition and replace the uses of `_transform` with `_radial_transform`, no?

src/bijectors/scale.jl
```diff
@@ -7,27 +7,21 @@ function Scale(a::Union{Real,AbstractArray}; dim::Val{D} = Val(ndims(a))) where
...
 _logabsdetjac_scale(a::Real, x::AbstractVector, ::Val{1}) = log(abs(a)) * length(x)
 _logabsdetjac_scale(a::Real, x::AbstractMatrix, ::Val{1}) = fill(log(abs(a)) * size(x, 1), size(x, 2))
-_logabsdetjac_scale(a::AbstractVector, x::AbstractVector, ::Val{1}) = sum(log.(abs.(a)))
-_logabsdetjac_scale(a::AbstractVector, x::AbstractMatrix, ::Val{1}) = fill(sum(log.(abs.(a))), size(x, 2))
-_logabsdetjac_scale(a::AbstractMatrix, x::AbstractVector, ::Val{1}) = log(abs(det(a)))
-_logabsdetjac_scale(a::AbstractMatrix, x::AbstractMatrix{T}, ::Val{1}) where {T} = log(abs(det(a))) * ones(T, size(x, 2))
\ No newline at end of file
+_logabsdetjac_scale(a::AbstractVector, x::AbstractVector, ::Val{1}) = sum(x -> log(abs(x)), a)
+_logabsdetjac_scale(a::AbstractVector, x::AbstractMatrix, ::Val{1}) = fill(sum(x -> log(abs(x)), a), size(x, 2))
+_logabsdetjac_scale(a::AbstractMatrix, x::AbstractVector, ::Val{1}) = logabsdet(a)
```
torfjelde: `logabsdet` returns a 2-tuple, right? So should this be `logabsdet(a)[1]`?

src/bijectors/scale.jl
```diff
@@ -7,27 +7,21 @@ function Scale(a::Union{Real,AbstractArray}; dim::Val{D} = Val(ndims(a))) where
...
+_logabsdetjac_scale(a::AbstractMatrix, x::AbstractMatrix{T}, ::Val{1}) where {T} = logabsdet(a) * ones(T, size(x, 2))
```
torfjelde: Same here

src/bijectors/stacked.jl
```diff
@@ -37,74 +37,65 @@ struct Stacked{Bs, N} <: Bijector{1} where N
...
 Stacked(bs) = Stacked(bs, tuple([i:i for i = 1:length(bs)]...))
 
-isclosedform(b::Stacked) = all(isclosedform.(b.bs))
+isclosedform(b::Stacked) = all(isclosedform, b.bs)
 
 stack(bs::Bijector{0}...) = Stacked(bs)
 
 # For some reason `inv.(sb.bs)` was unstable... This works though.
 inv(sb::Stacked) = Stacked(map(inv, sb.bs), sb.ranges)
-@generated function inv(sb::Stacked{A}) where {A <: Tuple}
```
torfjelde: The reason why this is here, rather than just using the `map` impl from above is that for large `Stacked`, i.e. large number of stacked bijectors, the result of `map` isn't type-inferrable. Thus, I don't think we should remove this

src/bijectors/stacked.jl
```diff
@@ -37,74 +37,65 @@ struct Stacked{Bs, N} <: Bijector{1} where N
...
+    init = sum(logabsdetjac(b.bs[1], x[b.ranges[1]]))
+    init + sum(2:N) do i
+        sum(logabsdetjac(b.bs[i], x[b.ranges[i]]))
+    end
 end
+
 function logabsdetjac(b::Stacked, x::AbstractMatrix{<:Real})
-    return [logabsdetjac(b, x[:, i]) for i = 1:size(x, 2)]
+    @views init = vcat(logabsdetjac(b, x[:,1]))
+    return mapreduce(vcat, Base.Iterators.drop(eachcol(x), 1); init=init) do c
```
torfjelde: `eachcol` isn't available in Julia-1.0

src/compat/tracker.jl
```diff
@@ -1,105 +1,428 @@
...
+
+Base.prevfloat(r::TrackedReal) = track(prevfloat, r)
+@grad function prevfloat(r::Real)
+    prevfloat(data(r)), Δ -> Δ
+end
+Base.nextfloat(r::TrackedReal) = track(nextfloat, r)
+@grad function nextfloat(r::Real)
+    nextfloat(data(r)), Δ -> Δ
+end
+for i = 0:2, c = Tracker.combinations([:AbstractArray, :TrackedArray, :TrackedReal, :Number], i), f = [:hcat, :vcat]
```
torfjelde: Can you explain why this is necessary? Is there an issue with tracking when one of the arguments are `TrackedReal`? 
If so, should this maybe be a fix to https://github.com/FluxML/Tracker.jl/blob/128c29d46d51e182319bbdbae066dc279bdee2af/src/lib/array.jl#L172-L176? Ofc can leave it here as a quickfix for now until that is fixed in Tracker.jl, but maybe at least add a comment about it?

src/compat/tracker.jl
```diff
@@ -1,105 +1,428 @@
...
 # end
 # @grad function _logabsdetjac_scale(a::TrackedMatrix, x::AbstractVector, ::Val{1})
 #     throw
 # end
 
 # implementations for Stacked bijector
 function logabsdetjac(b::Stacked, x::TrackedMatrix{<:Real})
-    return Tracker.collect([logabsdetjac(b, x[:, i]) for i = 1:size(x, 2)])
+    @views init = vcat(logabsdetjac(b, x[:, 1]))
+    return mapreduce(vcat, drop(eachcol(x), 1); init = init) do c
```
torfjelde: `eachcol` not available in Julia 1.0

src/compat/tracker.jl
```diff
@@ -1,105 +1,428 @@
...
+    end
 end
 # TODO: implement custom adjoint since we can exploit block-diagonal nature of `Stacked`
 function (sb::Stacked)(x::TrackedMatrix{<:Real})
-    init = reshape(sb(x[:, 1]), :, 1)
-    return Tracker.collect(mapreduce(i -> sb(x[:, i]), hcat, 2:size(x, 2); init = init))
-end
\ No newline at end of file
+    @views init = reshape(sb(x[:, 1]), :, 1)
+    return mapreduce(hcat, drop(eachcol(x), 1); init = init) do c
```
torfjelde: `eachcol` not available in Julia 1.0

src/compat/tracker.jl
```diff
@@ -1,105 +1,428 @@
...
+    xd = data(x)
+    return logabsdetjac(b, xd), Δ -> begin
+        @views init = reshape(simplex_logabsdetjac_gradient(xd[:,1]) * Δ[1], :, 1)
+        (nothing, mapreduce(hcat, drop(eachcol(xd), 1), drop(Δ, 1); init = init) do c, g
+            simplex_logabsdetjac_gradient(c) * g
+        end)
+    end
+end
+
+(ib::Inverse{<:Logit{<:TrackedReal}})(y::AbstractArray) = ib.(y)::TrackedArray
```
torfjelde: Are these actually useful? Seems you'd still have unstability due to the element-type of `TrackedArray` being unknown, no?

src/compat/distributionsad.jl
```diff
@@ -0,0 +1,246 @@
...
+    _logpdf_with_trans_pd(d, X, transform)
+end
+function logpdf_with_trans(
+    d::TuringWishart,
+    X::AbstractArray{<:AbstractMatrix{<:Real}},
+    transform::Bool
+)
+    _logpdf_with_trans_pd.(Ref(d), X, transform)
+end
+function link(d::TuringWishart, X::AbstractMatrix{<:Real})
```
torfjelde: Regarding _all_ these `link` and `invlink` definitions: would it maybe be a good idea to instead define a corresponding `Bijector` for these, and just let `link` and `invlink` be a proxy for evaluating this `Bijector` instead? This way we don't end up having to repeat ourselves when we inevitably implement these bijectors.

src/bijectors/simplex.jl
```diff
@@ -37,10 +49,22 @@ function (b::SimplexBijector{proj})(x::AbstractVector{T}) where {T, proj}
...
-function (b::SimplexBijector{proj})(X::AbstractMatrix{T}) where {T<:Real, proj}
-    Y, K, N = similar(X), size(X, 1), size(X, 2)
+function (b::SimplexBijector)(X::AbstractMatrix)
+    _simplex_bijector(X, b)
+end
+function (b::SimplexBijector)(
+    Y::AbstractMatrix,
+    X::AbstractMatrix,
+)
+    _simplex_bijector!(Y, X, b)
```
torfjelde: Seems like if some bijectors suddenly start taking multiple arguments, things can be a bit confusing. Granted, I guess we could establish the convention that if a `Bijector` takes two arguments, it's _always_ inplace?

This also kind of brings up the discussion we had waaay back about whether to use `(b::Bijector)(x)` or `transform(b, x)` (thus also allowing `transform!(b, x, y)`).

src/bijectors/simplex.jl
```diff
@@ -130,7 +178,332 @@ function logabsdetjac(b::SimplexBijector, x::AbstractVector{T}) where T
...
+        g[k-1] -= (1/(z + ϵ) - 1/((one(T) + ϵ) - z)) * x[k] / ((one(T) + ϵ) - sum_tmp)^2 - 1
+        g[k] -= (1/(z + ϵ) - 1/((one(T) + ϵ) - z)) * 1 / ((one(T) + ϵ) - sum_tmp)
+    end
+    return g
+end
 
 function logabsdetjac(b::SimplexBijector, x::AbstractMatrix{<:Real})
-    return [logabsdetjac(b, x[:, i]) for i = 1:size(x, 2)]
+    @views init = vcat(logabsdetjac(b, x[:,1]))
+    return mapreduce(c -> logabsdetjac(b, c), vcat, drop(eachcol(x), 1); init = init)
```
torfjelde: `eachcol` not in Julia 1.0

mohamed82008:
Thanks for your comments @torfjelde. Finally, came back to this PR. I will address/respond to your comments one by one.

src/bijectors/logit.jl
```diff
@@ -8,7 +8,11 @@ struct Logit{T<:Real} <: Bijector{0}
...
-(b::Logit)(x) = @. logit((x - b.a) / (b.b - b.a))
-(ib::Inversed{<:Logit{<:Real}})(y) = @. (ib.orig.b - ib.orig.a) * logistic(y) + ib.orig.a
+(b::Logit)(x::Real) = @. logit((x - b.a) / (b.b - b.a))
+(b::Logit)(x) = b.(x)
 
-logabsdetjac(b::Logit{<:Real}, x) = @. - log((x - b.a) * (b.b - x) / (b.b - b.a))
+(ib::Inverse{<:Logit})(y::Real) = (ib.orig.b - ib.orig.a) * logistic(y) + ib.orig.a
+(ib::Inverse{<:Logit})(y) = ib.(y)
+
+_logabsdetjac(b::Logit, x::Real) = -log((x - b.a) * (b.b - x) / (b.b - b.a))
```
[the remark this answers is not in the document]
mohamed82008: No reason. I will change. Thanks.

src/bijectors/planar_layer.jl
mohamed82008: You cannot update a `TrackedArray` in-place. `Array{TrackedReal}` yes. I think reconstructing the whole thing is reasonable. No need to use a 1 element array. If we really must make it mutable, we can use a `Ref`. A `Ref` is lighter on the compiler than an array.

src/bijectors/planar_layer.jl
mohamed82008: Whenever `TrackedArray` and broadcasting are involved, it seems that type inference jumps out of the window.
```julia
julia> using Tracker

julia> f(x) = log.(abs.(x))

julia> @code_warntype f(param(rand(3)))
Variables
  x::TrackedArray{…,Array{Float64,1}}

Body::Any
1 ─ %1 = Base.broadcasted(Main.abs, x)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(abs),Tuple{TrackedArray{…,Array{Float64,1}}}}
│   %2 = Base.broadcasted(Main.log, %1)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(log),Tuple{Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(abs),Tuple{TrackedArray{…,Array{Float64,1}}}}}}
│   %3 = Base.materialize(%2)::Any
└──      return %3
```

src/bijectors/radial_layer.jl
mohamed82008: Sure but this is more organized. I don't want to extract fields from `flow` in too many places.

src/bijectors/scale.jl
mohamed82008: Yes I think so. Nice catch!

src/bijectors/stacked.jl
mohamed82008: Ok I will bring it back with a comment. Thanks.

src/bijectors/stacked.jl
mohamed82008: I think it's defined in Compat. I can depend on that.

src/compat/tracker.jl
mohamed82008: True. This needs to be added in Tracker. I just put it here as a quick hack. Tracker was having issues with a single argument `vcat` and other `vcat` methods with `TrackedReal`.

src/compat/tracker.jl
mohamed82008: Hmm I thought I checked they helped, but now that I am double checking they don't seem to add much. I will remove them. Thanks!

src/compat/distributionsad.jl
mohamed82008: Sounds good.

src/bijectors/simplex.jl
mohamed82008: Hmm just curious what bijector takes 2 arguments? Can we just make the convention that it always takes 1, maybe a tuple if we really have to?

src/bijectors/logit.jl
```diff
@@ -8,7 +8,11 @@ struct Logit{T<:Real} <: Bijector{0}
     b::T
 end
 
-(b::Logit)(x) = @. logit((x - b.a) / (b.b - b.a))
-(ib::Inversed{<:Logit{<:Real}})(y) = @. (ib.orig.b - ib.orig.a) * logistic(y) + ib.orig.a
+(b::Logit)(x::Real) = @. logit((x - b.a) / (b.b - b.a))
```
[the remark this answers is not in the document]
mohamed82008: Done

mohamed82008:
I will add more tests now analogous to DistributionsAD but for `logpdf_with_trans` instead of `logpdf`.

mohamed82008:
All the AD tests are there now. Some tests are still failing so I will try to fix them tomorrow.

mohamed82008:
The `Dirichlet` AD test with matrix `X` is still failing along with some type stability tests. I hope to fix the remaining failures and merge this PR in 2 days.

src/Bijectors.jl
```diff
@@ -67,13 +67,23 @@ function maphcat(f, args...)
...
+function eachcolmaphcat(f, x1, x2)
+    out = [f(x1[:,i], x2[i]) for i in 1:size(x1, 2)]
+    init = reshape(out[1], :, 1)
+    return reduce(hcat, drop(out, 1); init = init)
+end
+function eachcolmaphcat(f, x)
+    out = map(f, eachcol(x))
+    init = reshape(out[1], :, 1)
+    return reduce(hcat, drop(out, 1); init = init)
+end
```
devmotion: What is the reason for defining `mapvcat`, `maphcat`, `eachcolmaphcat`, and `eachcolmapvcat` instead of, e.g., using `mapreduce(hcat, f, eachcol(x))` for `eachcolmaphcat(f, x)`? Type instabilities? Lately I managed to solve some type stability issues with `mapreduce` by using `reshape(x, Val(...))`, IIRC - maybe something similar could be applicable here, depending on your use case?

mohamed82008: No type stability is not an issue. It's just that Zygote doesn't like `eachcol` or `mapreduce`, they error. So I define these functions to be used by Tracker then define their adjoint using something that Zygote can differentiate.

devmotion: Ah I see :slightly_smiling_face:

mohamed82008:
Now only some `MvNormal` correctness tests are left to fix. Hopefully, I will fix them and merge this today.

mohamed82008:
The MvNormal issue is a DistributionsAD issue https://github.com/TuringLang/DistributionsAD.jl/issues/39. So I will merge this as soon as other tests pass and then go back to DAD to fix it.

src/bijectors/planar_layer.jl
torfjelde: This issue just came up: https://github.com/TuringLang/Bijectors.jl/issues/121, and I remembered that we had a discussion exactly about this part of the code but couldn't remember the conclusion. It also seemed weird that I had "approved" such a change because it breaks with the users expectations. So it seems that the fact that it was marked as resolved made it go under my radar :confused:

So for future reference: I think it's a good idea to let the "asker" resolve the question rather than the "answerer" :+1:

torfjelde: Also, just to clarify: I think `b` should still be wrapped in `container` and change the default `container` to just be `identity` (since `randn` returns an `Array` anyways) or we add another argument specific for the `Real` parameter. It's useful when you're using, say, `Tracker.param`.

torfjelde: Maybe just rename the argument to `wrapper` or something

mohamed82008: Sounds good, sorry about that!

torfjelde: No worries; might also have been my fault in the sense that I wasn't quick enough to address the comment after you answered :+1:

PR #79, opened 2020-02-11
[WIP] Rational Quadratic Spline

torfjelde:
EDIT: Realize I've needlessly started this branch from `tor/couplings`, so I'll close this PR and make a new one starting from `master`.

Implementation of Rational Quadratic Spline (RQS) [1] as a `Bijector`.

## Notes
- At the moment `RationalQuadraticSpline` takes "unconstrained" parameters, i.e. expects input on ℝᵈ, despite the fact that under the hood we transform `widths` and `heights` to lie on a (d-1)-simplex S ⊂ ℝᵈ and `derivatives` to be positive. I initially didn't do this way, but instead let the user choose between passing the parameters wrapped in a `Unconstrained` struct, in which case we would perform the transformation, or as raw values, in which case it was assumed the parameters indeed was contained within their respective domains. I dropped the use of `Unconstrained` to reduce "new" code in this PR and since this is very similar to an issue we've expressed interest in solving in more generality.

## TODOs
- [ ] Added Compat.jl as a requirement due to how much easier life becomes. That okay?
- [ ] Need to clean up parts of the implementation, e.g. batched multivariate impls.
- [ ] The way we currently do batching in `forward` for RQS is not the most efficient. So maybe drop the use of "merging" named tuples and instead implement "by hand" a batched version of `forward`.

## References
1. Durkan, C., Bekasov, A., Murray, I., & Papamakarios, G., Neural Spline Flows, CoRR, arXiv:1906.04032 [stat.ML], (),  (2019).


torfjelde:
Closed in favour of #80

PR #81, opened 2020-02-11
LeakyReLU

torfjelde:
Implementation of LeakyReLU as a `Bijector`, which defines the invertible mapping

    x ↦ x if x ≥ 0 else αx

where α > 0.

## TODOs
- [x] Clean up testing a bit and make sure we've captured everything.


torfjelde:
Tests fail due to type-instability in Tracker.jl (in particular: https://github.com/FluxML/Tracker.jl/issues/42).

The question is: do we care? I'm not sure what the current state of the "should we drop support for Tracker.jl?"-discussion is :confused:

devmotion:
Some test errors on Travis are real and not Tracker-related though. Test errors on Zygote should be fixed by the next release of DistributionsAD (see https://github.com/TuringLang/DistributionsAD.jl/pull/108#issuecomment-689643838).

devmotion:
> The question is: do we care? I'm not sure what the current state of the "should we drop support for Tracker.jl?"-discussion is

IMO Tracker is well supported so far (in particular after dropping support for Julia < 1.3), e.g., all supported distributions in DistributionsAD work with Tracker (https://github.com/TuringLang/DistributionsAD.jl/blob/master/test/ad/distributions.jl). So if possible, I think we should support it.

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+# (N=0) Univariate case
+function (b::LeakyReLU{<:Any, 0})(x::Real)
+    mask = x < zero(x)
+    return mask * b.α * x + (1 - mask) * x
+end
+(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = b.(x)
+
+function (ib::Inverse{<:LeakyReLU, 0})(y::Real)
+    mask = y < zero(y)
+    return mask * (x / b.α) + (1 - mask) * x
```
devmotion, line 28:
```suggestion
    return mask * (x / ib.orig.α) + (1 - mask) * x
```

xukai92:
Why do we implement our own instead of using https://github.com/FluxML/NNlib.jl/blob/master/src/activation.jl#L73?

src/bijectors/leaky_relu.jl
torfjelde: Ah, thanks! But how was this not caught by the tests? o.O I'll check.

torfjelde:
> Why do we implement our own instead of using https://github.com/FluxML/NNlib.jl/blob/master/src/activation.jl#L73?

I had no idea this existed! It might also not have at the time when I first implemented LeakyReLU. But that's neat! I guess we could just do:
```julia
(b::LeakyReLU)(x) = NNlib.leakyrelu(x, b.α)
(ib::Inverse{<:LeakyReLU})(y) = NNlib.leakyrelu(y, inv(ib.orig.α))
```
I'll try that :+1: Thanks!

torfjelde:
Of course
```julia
(ib::Inverse{<:LeakyReLU})(y) = NNlib.leakyrelu(y, inv(ib.orig.α))
```
doesn't work. Their impl assumes that the `α ∈ (0, 1)`. So this would mean that we could only make use of their impl for the forward pass, which is not *that* useful, right?

torfjelde:
> Tests fail due to type-instability in Tracker.jl (in particular: FluxML/Tracker.jl#42).

So it seems like this doesn't fail after all! It might have been me just using a global scope, whoops...

src/bijectors/leaky_relu.jl
devmotion: It was, by the tests that run on Travis.

torfjelde:
Ah, I didn't know that Travis is the only one testing the bijectors at this point... Then there are still issues with not Tracker and Broadcasting not working togheter.

torfjelde:
@devmotion I see what you meant now, but there are still issues related to Tracker.jl making it type unstable.

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+function forward(b::LeakyReLU{<:Any, 0}, x::Real)
+    mask = x < zero(x)
+    J = mask * b.α + (1 - mask) * one(x)
+    return (rv=J * x, logabsdetjac=log(abs(J)))
+end
+
+# Batched version
+function forward(b::LeakyReLU{<:Any, 0}, x::AbstractVector)
+    mask = x .< zero(eltype(x))
+    J = mask .* b.α .+ (1 .- mask) .* one(eltype(x))
```
devmotion, line 52: Maybe
```suggestion
    mask = x .< zero(eltype(x))
    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
```
would be more efficient (since it avoids the allocation of `mask`) and more stable (since it does not require computations based on types).

As a side remark, IMO it is a bit unfortunate that currently so many almost identical implementations of a function are needed to define a bijector. Maybe this could be resolved by defining defaults on a high level for basically all bijectors (or a subtype of BaseBijectors, similar to BaseKernels in KernelFunctions) that call just in one or two methods that users would have to implement if their bijectors belong to these simple standard groups. With Julia >= 1.3 it is also possible to define methods for abstract functions such as
```julia
(f::Bijector{T,0})(x::AbstractVecOrMat{<:Real}) where T = map(f, x)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+    return (rv=J .* x, logabsdetjac=log.(abs.(J)))
+end
+
+# (N=1) Multivariate case, with univariate parameter `α`
+function (b::LeakyReLU{<:Any, 1})(x::AbstractVecOrMat)
+    mask = x .< zero(eltype(x))
+    return mask .* b.α .* x .+ (1 .- mask) .* x
```
devmotion, line 59: Similar here:
```suggestion
    return @. (x < zero(x)) * b.α * x + (x > zero(x)) * x
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+end
+
+function (ib::Inverse{<:LeakyReLU, 1})(y::AbstractVecOrMat)
+    mask = x .< zero(eltype(y))
+    return mask .* (y ./ ib.orig.α) .+ (1 .- mask) .* y
```
devmotion, line 64: And here:
```suggestion
    return @. (x < zero(x)) * y / ib.orig.α + (x > zero(x)) * y
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+end
+
+function logabsdetjac(b::LeakyReLU{<:Any, 1}, x::AbstractVecOrMat)
+    # Is really diagonal of jacobian
+    mask = x .< zero(eltype(x))
+    J = mask .* b.α .+ (1 .- mask) .* one(eltype(x))
```
devmotion, line 70:
```suggestion
    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+    end
+end
+
+# We implement `forward` by hand since we can re-use the computation of
+# the Jacobian of the transformation. This will lead to faster sampling
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 1}, x::AbstractVecOrMat)
+    # Is really diagonal of jacobian
+    mask = x .< zero(eltype(x))
+    J = mask .* b.α .+ (1 .- mask) .* one(eltype(x))
```
devmotion, line 85:
```suggestion
    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+LeakyReLU(α::T; dim::Val{N} = Val(D)) where {D, T<:AbstractArray{<:Real, D}, N} = LeakyReLU{T, N}(α)
+
+up1(b::LeakyReLU{T, N}) where {T, N} = LeakyReLU{T, N + 1}(b.α)
+
+# (N=0) Univariate case
+function (b::LeakyReLU{<:Any, 0})(x::Real)
+    mask = x < zero(x)
+    return mask * b.α * x + (1 - mask) * x
+end
+(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = b.(x)
```
devmotion, line 24: No need for broadcasting here (in particular with the advanced broadcasting machinery and the Tracker issues in mind):
```suggestion
(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = map(b, x)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+
+function (ib::Inverse{<:LeakyReLU, 0})(y::Real)
+    mask = y < zero(y)
+    return mask * (y / ib.orig.α) + (1 - mask) * y
+end
+(ib::Inverse{<:LeakyReLU{<:Any}, 0})(y::AbstractVector{<:Real}) = ib.(y)
```
devmotion, line 30:
```suggestion
(ib::Inverse{<:LeakyReLU{<:Any}, 0})(y::AbstractVector{<:Real}) = map(ib, y)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+
+LeakyReLU(α::T; dim::Val{N} = Val(0)) where {T<:Real, N} = LeakyReLU{T, N}(α)
+LeakyReLU(α::T; dim::Val{N} = Val(D)) where {D, T<:AbstractArray{<:Real, D}, N} = LeakyReLU{T, N}(α)
+
+up1(b::LeakyReLU{T, N}) where {T, N} = LeakyReLU{T, N + 1}(b.α)
+
+# (N=0) Univariate case
+function (b::LeakyReLU{<:Any, 0})(x::Real)
+    mask = x < zero(x)
+    return mask * b.α * x + (1 - mask) * x
```
devmotion, line 22: Could also write
```suggestion
    return mask * b.α * x + !mask * x
```
to keep promotions to the lowest possible level.

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+end
+(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = b.(x)
+
+function (ib::Inverse{<:LeakyReLU, 0})(y::Real)
+    mask = y < zero(y)
+    return mask * (y / ib.orig.α) + (1 - mask) * y
```
devmotion, line 28:
```suggestion
    return mask * (y / ib.orig.α) + !mask * y
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+end
+(ib::Inverse{<:LeakyReLU{<:Any}, 0})(y::AbstractVector{<:Real}) = ib.(y)
+
+function logabsdetjac(b::LeakyReLU{<:Any, 0}, x::Real)
+    mask = x < zero(x)
+    J = mask * b.α + (1 - mask) * one(x)
+    return log(abs(J))
+end
+logabsdetjac(b::LeakyReLU{<:Real, 0}, x::AbstractVector{<:Real}) = logabsdetjac.(b, x)
```
devmotion, line 37: Not sure if this is more efficient or not:
```suggestion
logabsdetjac(b::LeakyReLU{<:Real, 0}, x::AbstractVector{<:Real}) = map(x -> logabsdetjac(b, x), x)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,95 @@
...
+
+
+# We implement `forward` by hand since we can re-use the computation of
+# the Jacobian of the transformation. This will lead to faster sampling
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 0}, x::Real)
+    mask = x < zero(x)
+    J = mask * b.α + (1 - mask) * one(x)
```
devmotion, line 45:
```suggestion
    J = mask * b.α + !mask * one(x)
```

src/bijectors/leaky_relu.jl
torfjelde: > would be more efficient (since it avoids the allocation of mask) and more stable (since it does not require computations based on types).

Won't this call `zero(x)` n times though, i.e. have the same allocation? Other than that, it seems like a good idea :+1: 

> As a side remark

100%. 

> Maybe this could be resolved by defining defaults on a high level for basically all bijectors (or a subtype of BaseBijectors, similar to BaseKernels in KernelFunctions) that call just in one or two methods that users would have to implement if their bijectors belong to these simple standard groups.

I'm unaware of what KernelFunctions.jl does. Are they simply not making the `struct` callable? If so, it was a deliberate design-decision to go with making structs callable when we started out. We were debating whether or not to do this or include some `transform` method so we could define "one method to rule them all" on an abstract type. Ended up going with the "callable struct" approach, but it def has it's issues, i.e. redundant code. 

I've recently played around a bit with actually using a `transform` method under the hood, but adding a macro which allows you to say `@deftransform function transform(b, x) ... end` and we just add a `(b::Bijector)(x) = transform(b, x)` just after the method declaration. This would also allow us to implement in-place versions of all the methods, i.e. `transform!(b, x, out)`, `logabsdetjac(b, x, out)`, and so on. Thoughts?

> With Julia >= 1.3 it is also possible to define methods for abstract functions such as

Woah, really? If so that would solve the problem, no?

devmotion: > I'm unaware of what KernelFunctions.jl does. Are they simply not making the struct callable? If so, it was a deliberate design-decision to go with making structs callable when we started out. We were debating whether or not to do this or include some transform method so we could define "one method to rule them all" on an abstract type. Ended up going with the "callable struct" approach, but it def has it's issues, i.e. redundant code.

No, KernelFunctions API is based on callable structs only (using a function was removed a while ago). But e.g. translation-invariant kernels are built all in the same way, usually they use some simple function to evaluate the distance between the inputs (e.g. using Distances) and then they apply some nonlinear mapping afterwards. With this special structure also all kind of optimizations are possible when constructing kernel matrices etc. Hence there is a special type of such kernels (SimpleKernel IIRC), and then users just define their metric and the nonlinear mapping and get everything else for free. There's more information here: https://juliagaussianprocesses.github.io/KernelFunctions.jl/dev/create_kernel/

torfjelde: To add on to the above, because it sucks to have to implement `(b::Bijector)`, `logabsdetjac` and then finally `forward`, would it be an idea to add a macro that would allow you to define all in one go? 

I.e. something like the following. IMHO it feels a bit like "overkill" (plus I bet the macro below is not the cleanest way of doing it), but curious what you think. 
```julia
"""
    @bijector function forward(b::Bijector, x) ... end

Takes the method `forward` and uses it to define both `transform`
and `logabsdetjac`, while ensuring that any shared computation is
taken advantage of in `forward`.
"""
macro bijector(expr)
    def = MacroTools.splitdef(expr)
    body = def[:body]
    args = def[:args]

    # Figure out what is shared, and what isn't
    shared_exprs = [] # beginning of `forward`
    transform_exprs = []
    logjac_exprs = []
    tail_exprs = [] # goes at the end of `forward`

    for expr in body.args
        # When using `rv = ...` and `logabsdetjac = ...`
        if Meta.isexpr(expr, :(=)) && expr.args[1] == :rv
            push!(transform_exprs, Expr(:return, expr.args[2]))
            push!(tail_exprs, expr)
        elseif Meta.isexpr(expr, :(=)) && expr.args[1] == :logabsdetjac
            push!(logjac_exprs, Expr(:return, expr.args[2]))
            push!(tail_exprs, expr)
        else
            # If still sharing, add those expressions
            issharing = (length(transform_exprs) == 0) && (length(logjac_exprs) == 0)
            if issharing
                push!(shared_exprs, expr)
            else
                push!(tail_exprs, expr)
            end
        end
    end

    # Add the shared computation
    transform_full_exprs = copy(shared_exprs)
    append!(transform_full_exprs, transform_exprs)

    logjac_full_exprs = copy(shared_exprs)
    append!(logjac_full_exprs, logjac_exprs)

    # Remove the redundant macro's from the `forward` body
    forward_full_exprs = vcat(shared_exprs, tail_exprs)

    return quote
        $(Bijectors).@deftransform function $(Bijectors).transform($(args...))
            $(transform_full_exprs...)
        end

        function $(Bijectors).logabsdetjac($(args...))
            $(logjac_full_exprs...)
        end

        function $(Bijectors).forward($(args...))
            $(forward_full_exprs...)
        end
    end
end
```
in which case something like
```julia
@macroexpand Bijectors.@bijector function forward(b::Bijectors.Exp{0}, x::Real)
    # Shared computation
    z = exp(x)
    
    rv = begin
        # `(b::Bijector)(x)` goes here
        z
    end
    logabsdetjac = begin
        # `logabsdetjac(b, x)` goes here
        log(z)
    end

    return (rv = rv, logabsdetjac = logabsdetjac)
end
```
is expanded to
```julia
$(Expr(:meta, :doc))
function (Bijectors).transform(var"#4#b"::(Bijectors.Bijectors).Exp{0}, var"#7#x"::Bijectors.Real)
    var"#5#z" = Bijectors.exp(var"#7#x")
    return var"#5#z"
end

(var"#4#b"::(Bijectors.Bijectors).Exp{0})(var"#9#x"::Bijectors.Real) = begin
    (Bijectors).transform(var"#4#b", var"#9#x")
end

function (Bijectors).logabsdetjac(var"#11#b"::(Bijectors.Bijectors).Exp{0}, var"#12#x"::Bijectors.Real)
    var"#10#z" = Bijectors.exp(var"#12#x")
    return Bijectors.log(var"#10#z")
end
function (Bijectors).forward(var"#16#b"::(Bijectors.Bijectors).Exp{0}, var"#17#x"::Bijectors.Real)
    var"#13#z" = Bijectors.exp(var"#17#x")
    var"#14#rv" = var"#13#z"
    var"#15#logabsdetjac" = Bijectors.log(var"#13#z")
    return (rv = var"#14#rv", logabsdetjac = var"#15#logabsdetjac")
end
```

torfjelde: > But e.g. translation-invariant kernels are built all in the same way, usually they use some simple function to evaluate the distance between the inputs (e.g. using Distances) and then they apply some nonlinear mapping afterwards.

Ah, I see. I don't think it's so easy to do for bijectors as we have much more structure than just "forward" evaluation, no?

torfjelde:
Okay, so _now_ the only issue is https://github.com/FluxML/Tracker.jl/issues/42

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,91 @@
...
+    return mask * b.α * x + !mask * x
+end
+(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = map(b, x)
+
+function (ib::Inverse{<:LeakyReLU, 0})(y::Real)
+    mask = y < zero(y)
+    return mask * (y / ib.orig.α) + !mask * y
+end
+(ib::Inverse{<:LeakyReLU{<:Any}, 0})(y::AbstractVector{<:Real}) = map(ib, y)
+
```
devmotion, line 31: Couldn't this just be
```suggestion
Base.inv(b::LeakyReLU) = LeakyReLU(inv.(b.a))
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,91 @@
...
+# (N=1) Multivariate case, with univariate parameter `α`
+function (b::LeakyReLU{<:Any, 1})(x::AbstractVecOrMat)
+    return @. (x < zero(x)) * b.α * x + (x > zero(x)) * x
+end
+
+function (ib::Inverse{<:LeakyReLU, 1})(y::AbstractVecOrMat)
+    return @. (y < zero(y)) * y / ib.orig.α + (y > zero(x)) * y
+end
+
```
devmotion, line 64: With the suggestion above this could be dropped I guess:
```suggestion
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,91 @@
...
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 0}, x::Real)
+    mask = x < zero(x)
+    J = mask * b.α + !mask * one(x)
+    return (rv=J * x, logabsdetjac=log(abs(J)))
+end
+
+# Batched version
+function forward(b::LeakyReLU{<:Any, 0}, x::AbstractVector)
+    mask = x .< zero(eltype(x))
```
devmotion, line 51:
```suggestion
```

src/bijectors/leaky_relu.jl
torfjelde: Ah, of course! Nice!

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,81 @@
...
+    return (rv=J * x, logabsdetjac=log(abs(J)))
+end
+
+# Batched version
+function forward(b::LeakyReLU{<:Any, 0}, x::AbstractVector)
+    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
+    return (rv=J .* x, logabsdetjac=log.(abs.(J)))
+end
+
+# (N=1) Multivariate case, with univariate parameter `α`
```
devmotion, line 50: The type parameters do not ensure that the parameter is univariate, do they?

torfjelde: Ah, yeah that comment is wrong. I realized that we could just have the same impl for `Real` and `AbstractVector` and just made it `Any`; forgot to change the comment.

torfjelde, line 50:
```suggestion
# (N=1) Multivariate case, with real or vector parameter `α`
```

devmotion: Maybe just
```suggestion
# (N=1) Multivariate case
```
since you don't restrict the parameter at all?

devmotion:
> Okay, so now the only issue is FluxML/Tracker.jl#42

Are you sure that this causes the test error? The linked issue seems to be about broadcasting issues with Tracker if the method is not defined generically enough that result in a MethodError (if Tracker doesn't know how to handle it, it falls back to ForwardDiff - but this requires that the method allows for inputs of type ForwardDiff.Dual) whereas here it seems to be an inference problem.

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,81 @@
...
+up1(b::LeakyReLU{T, N}) where {T, N} = LeakyReLU{T, N + 1}(b.α)
+
+# (N=0) Univariate case
+function (b::LeakyReLU{<:Any, 0})(x::Real)
+    mask = x < zero(x)
+    return mask * b.α * x + !mask * x
+end
+(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = map(b, x)
+
+Base.inv(b::LeakyReLU) = LeakyReLU(inv.(b.a))
```
devmotion, line 26: Oops sorry, it should of course be
```suggestion
Base.inv(b::LeakyReLU) = LeakyReLU(inv.(b.α))
```

torfjelde:
> Are you sure that this causes the test error? 

Looking closer, it seems like you're right. It might not be related after all :confused:

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,81 @@
...
+up1(b::LeakyReLU{T, N}) where {T, N} = LeakyReLU{T, N + 1}(b.α)
+
+# (N=0) Univariate case
+function (b::LeakyReLU{<:Any, 0})(x::Real)
+    mask = x < zero(x)
+    return mask * b.α * x + !mask * x
+end
+(b::LeakyReLU{<:Any, 0})(x::AbstractVector{<:Real}) = map(b, x)
+
+Base.inv(b::LeakyReLU) = LeakyReLU(inv.(b.α))
```
devmotion, line 26: The dimension can't be inferred:
```suggestion
function Base.inv(b::LeakyReLU{<:Any,N}) where N
    invα = inv.(b.α)
    return LeakyReLU{typeof(invα),N}(invα)
end

```

devmotion:
I had a closer look at the Tracker issue and made the following interesting observation:
```julia
julia> using Tracker, Test

julia> f(x::AbstractVecOrMat) = @. x < zero(x)

julia> @inferred(f(Tracker.param(rand(5,4))))
ERROR: return type BitArray{2} does not match inferred return type Any
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at REPL[21]:1

julia> f(x::AbstractVecOrMat) = @. x < 0

julia> @inferred(f(Tracker.param(rand(5,4))))
5×4 BitArray{2}:
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0

julia> f(x::AbstractVecOrMat) = @. x < $(zero(eltype(x)))

julia> @inferred(f(Tracker.param(rand(5,4))))
5×4 BitArray{2}:
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0

julia> g(x) = x < zero(x)

julia> f(x::AbstractVecOrMat) = g.(x)
f (generic function with 2 methods)

julia> @inferred(f(Tracker.param(rand(5,4))))
5×4 BitArray{2}:
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
 0  0  0  0
```
I don't know why the first approach does not work but I guess the Tracker could be fixed by defining two functions `isnegative(x) = x < zero(x)` and `ispositive(x) = x > zero(x)`, and writing
```julia
J = @. isnegative(x) * a * x + ispositive(x) * x
```
instead of
```julia
J = @. (x < zero(x)) * a * x + (x > zero(x)) * x
```

torfjelde:
This is some strange stuff:
```julia
julia> f(x::AbstractVecOrMat) = @. x < zero(x)

julia> @code_warntype f(x)
Variables
  x::TrackedArray{…,Array{Float64,2}}

Body::Any
1 ─ %1 = Base.broadcasted(Main.zero, x)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(zero),Tuple{TrackedArray{…,Array{Float64,2}}}}
│   %2 = Base.broadcasted(Main.:<, x, %1)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(<),Tuple{TrackedArray{…,Array{Float64,2}},Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(zero),Tuple{TrackedArray{…,Array{Float64,2}}}}}}
│   %3 = Base.materialize(%2)::Any
└──      return %3

julia> g(x::AbstractVecOrMat) = @. x < 0

julia> @code_warntype g(x)
Variables
  #self#::Core.Compiler.Const(g, false)
  x::TrackedArray{…,Array{Float64,2}}

Body::BitArray{2}
1 ─ %1 = Base.broadcasted(Main.:<, x, 0)::Core.Compiler.PartialStruct(Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(<),Tuple{TrackedArray{…,Array{Float64,2}},Int64}}, Any[Core.Compiler.Const(<, false), Core.Compiler.PartialStruct(Tuple{TrackedArray{…,Array{Float64,2}},Int64}, Any[TrackedArray{…,Array{Float64,2}}, Core.Compiler.Const(0, false)]), Core.Compiler.Const(nothing, false)])
│   %2 = Base.materialize(%1)::BitArray{2}
└──      return %2

julia> function h(x::AbstractVecOrMat)
           z = zero(x)
           return @. x < z
       end
h (generic function with 1 method)

julia> @code_warntype h(x)
Variables
  #self#::Core.Compiler.Const(h, false)
  x::TrackedArray{…,Array{Float64,2}}
  z::Array{Float64,2}

Body::BitArray{2}
1 ─      (z = Main.zero(x))
│   %2 = Base.broadcasted(Main.:<, x, z)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(<),Tuple{TrackedArray{…,Array{Float64,2}},Array{Float64,2}}}
│   %3 = Base.materialize(%2)::BitArray{2}
└──      return %3
```
Sooo I think we fix it without custom-adjoints, i.e. by just not broadcasting the `zero`.

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,85 @@
...
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 0}, x::Real)
+    mask = x < zero(x)
+    J = mask * b.α + !mask * one(x)
+    return (rv=J * x, logabsdetjac=log(abs(J)))
+end
+
+# Batched version
+function forward(b::LeakyReLU{<:Any, 0}, x::AbstractVector)
+    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
```
torfjelde, line 50:
```suggestion
    z = zero(x)
    o = one(x)
    J = @. (x < z) * b.α + (x > z) * o
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,85 @@
...
+    return (rv=J .* x, logabsdetjac=log.(abs.(J)))
+end
+
+# (N=1) Multivariate case
+function (b::LeakyReLU{<:Any, 1})(x::AbstractVecOrMat)
+    return @. (x < zero(x)) * b.α * x + (x > zero(x)) * x
```
torfjelde, line 56:
```suggestion
    z = zero(x)
    return @. (x < z) * b.α * x + (x > z) * x
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,85 @@
...
+end
+
+function logabsdetjac(b::LeakyReLU{<:Any, 1}, x::AbstractVecOrMat)
+    # Is really diagonal of jacobian
+    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
```
torfjelde, line 61:
```suggestion
    z = zero(x)
    o = one(x)
    J = @. (x < z) * b.α + (x > z) * o
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,85 @@
...
+        return vec(sum(log.(abs.(J)); dims = 1))  # sum along column
+    end
+end
+
+# We implement `forward` by hand since we can re-use the computation of
+# the Jacobian of the transformation. This will lead to faster sampling
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 1}, x::AbstractVecOrMat)
+    # Is really diagonal of jacobian
+    J = @. (x < zero(x)) * b.α + (x > zero(x)) * one(x)
```
torfjelde, line 75:
```suggestion
    J = let z = zero(x), o = one(x)
        @. (x < z) * b.α + (x > z) * o
    end
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,93 @@
...
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 0}, x::Real)
+    mask = x < zero(x)
+    J = mask * b.α + !mask * one(x)
+    return (rv=J * x, logabsdetjac=log(abs(J)))
+end
+
+# Batched version
+function forward(b::LeakyReLU{<:Any, 0}, x::AbstractVector)
+    J = let z = zero(x), o = one(x)
```
torfjelde, line 50:
```suggestion
    J = let T = eltype(x), z = zero(T), o = one(T)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,93 @@
...
+        @. (x < z) * b.α + (x > z) * o
+    end
+    return (rv=J .* x, logabsdetjac=log.(abs.(J)))
+end
+
+# (N=1) Multivariate case
+function (b::LeakyReLU{<:Any, 1})(x::AbstractVecOrMat)
+    return let z = zero(x)
```
torfjelde, line 58:
```suggestion
    return let z = zero(eltype(x))
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,93 @@
...
+        @. (x < z) * b.α * x + (x > z) * x
+    end
+end
+
+function logabsdetjac(b::LeakyReLU{<:Any, 1}, x::AbstractVecOrMat)
+    # Is really diagonal of jacobian
+    J = let z = zero(x), o = one(x)
```
torfjelde, line 65:
```suggestion
    J = let T = eltype(x), z = zero(T), o = one(T)
```

src/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,93 @@
...
+        return vec(sum(log.(abs.(J)); dims = 1))  # sum along column
+    end
+end
+
+# We implement `forward` by hand since we can re-use the computation of
+# the Jacobian of the transformation. This will lead to faster sampling
+# when using `rand` on a `TransformedDistribution` making use of `LeakyReLU`.
+function forward(b::LeakyReLU{<:Any, 1}, x::AbstractVecOrMat)
+    # Is really diagonal of jacobian
+    J = let z = zero(x), o = one(x)
```
torfjelde, line 81:
```suggestion
    J = let T = eltype(x), z = zero(T), o = one(T)
```

src/bijectors/leaky_relu.jl
devmotion, line 50: IMO that's not a good fix since it destroys the whole point of just having one broadcast expression that is fused together. Now we would allocate two additional vectors of ones and zeros just for Tracker.

Did you check if we can work around the Tracker bug by using a separate function `ispositive(x) = x > zero(x)` and `isnegative(x) = x < zero(x)` that we can include in the broadcast expression instead of the explicit `x < zero(x)` etc statements?

torfjelde: > Now we would allocate two additional vectors of ones and zeros just for Tracker.

Not two vectors; just two _numbers_, right? (I realize the above impl was wrong, it was supposed to contain `eltype`)

torfjelde: Also, even if we use `isnegative`, we're running into the same issue because we're broadcasting over `one(x)`, so we would have to also have this in a separate function, e.g. `mul1(x, y) = x * one(y)` or something :confused:

devmotion: With `eltype` it's just a number - IMO it's still unfortunate since there would be no need for this type-based computation if Tracker would be fixed (I opened a PR: https://github.com/FluxML/Tracker.jl/pull/85). In general it's better to not use types if not needed since the instances contain more information (similar argument for why generated functions should be used only if needed). It *should* just work with this single broadcast expression here.

Apart from that, are the `let` blocks actually needed?

devmotion: BTW it seems the tests still fail?

torfjelde: It still breaks, but now the only one failing is `LeakyReLU{..., 1}`. Again it hits a snag at `materlialize`:
```julia
Variables
  b::LeakyReLU{Float64,1}
  x::TrackedArray{…,Array{Float64,2}}
  z::Tracker.TrackedReal{Float64}

Body::Any
1 ─ %1 = Bijectors.eltype(x)::Core.Compiler.Const(Tracker.TrackedReal{Float64}, false)
│        (z = Bijectors.zero(%1))
│   %3 = Base.broadcasted(Bijectors.:<, x, z)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(<),Tuple{TrackedArray{…,Array{Float64,2}},Tracker.TrackedReal{Float64}}}
│   %4 = Base.getproperty(b, :α)::Float64
│   %5 = Base.broadcasted(Bijectors.:*, %3, %4)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(*),Tuple{Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(<),Tuple{TrackedArray{…,Array{Float64,2}},Tracker.TrackedReal{Float64}}},Float64}}
│   %6 = Base.broadcasted(Bijectors.:>, x, z)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(>),Tuple{TrackedArray{…,Array{Float64,2}},Tracker.TrackedReal{Float64}}}
│   %7 = Base.broadcasted(Bijectors.:*, %6, x)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(*),Tuple{Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(>),Tuple{TrackedArray{…,Array{Float64,2}},Tracker.TrackedReal{Float64}}},TrackedArray{…,Array{Float64,2}}}}
│   %8 = Base.broadcasted(Bijectors.:+, %5, %7)::Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(+),Tuple{Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(*),Tuple{Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(<),Tuple{TrackedArray{…,Array{Float64,2}},Tracker.TrackedReal{Float64}}},Float64}},Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(*),Tuple{Base.Broadcast.Broadcasted{Tracker.TrackedStyle,Nothing,typeof(>),Tuple{TrackedArray{…,Array{Float64,2}},Tracker.TrackedReal{Float64}}},TrackedArray{…,Array{Float64,2}}}}}}
│   %9 = Base.materialize(%8)::Any
└──      return %9
```

torfjelde: > IMO it's still unfortunate since there would be no need for this type-based computation if Tracker would be fixed (I opened a PR: FluxML/Tracker.jl#85)

100% agree! And nice! 

> Apart from that, are the let blocks actually needed?

Nah, I just added them to make explicit that it's only needed for this particular statement, and not to clutter the names in the full function. Was nicer IMO, but ofc subjective.

test/bijectors/leaky_relu.jl
```diff
@@ -0,0 +1,86 @@
+using Test
+
+using Bijectors
+using Bijectors: LeakyReLU
+
+using LinearAlgebra
+using ForwardDiff
```
cpfiffer, line 7: I think my only real comment here is on testing other AD backends like ReverseDiff, but I'm not sure how important that is here.

cpfiffer (approved):
Looks good to me. Seems like David already did the hard work. I left a minor test comment, but I couldn't find anything in the actual code that seems incorrect or lacking.

test/bijectors/leaky_relu.jl
torfjelde: There are some tests for it in `test/bijectors/interface.jl`. But yeah, testing in Bijectors is honestly a bit of mess atm. In a couple of the other PRs I've added some functionality which makes it easier to use a "standardized" testing suite for a new `Bijector`, so the plan is to use that in the future :+1:

PR #82, opened 2020-02-18
Making SimplexBijector compatible with Tracker.jl

torfjelde:
This PR fixes the issues with Tracker.jl and SimplexBijector, using the by-hand implementations of the jacobian of `SimplexBijector` and its inverse (thanks to @mohamed82008!).

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -189,32 +189,7 @@ function link_jacobian(
     x::AbstractVector{T},
     ::Type{Val{proj}} = Val{true}
 ) where {T<:Real, proj}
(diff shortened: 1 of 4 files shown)


xukai92:
Is this PR mostly moving around codes?

torfjelde:
> Is this PR mostly moving around codes?

Moving around code and fixing the bug that `SimplexBijector` doesn't work with Tracker.jl, e.g. `SimplexBijector()(Tracker.param(randn(10))` will fail and similarily for its inverse.

mohamed82008:
Thanks @torfjelde !

Issue #85, opened 2020-03-07
TransformedDistributions should support maximum, minimum


arnavs:
Was testing these against the `UnivariateDistribution` fallback in `Expectations.jl`, but it fails because these methods aren't defined.

arnavs:
Same for `params()`

torfjelde:
I actually had a branch lying around with implementations for this:) But I didn't make it into a PR yet as it's slightly unclear to me whether or not `params` should return the parameters of the underlying distribution, i.e. `params(td.dist)`, or if it should also include parameters for the `Bijector`. If so, there's a fair bit of work and planning that needs to go into it. But, if users expect `params` to return only the parameters of the distribution, then this is no problem (see #86)

arnavs:
That's a good point about `params.` Why not both? Ideally in a "flat" way, e.g. `mu, sigma, transformationstuff...`

arnavs:
Closing as the PR was merged.

PR #86, opened 2020-03-08
Addressing #85

torfjelde:
This PR implements `StatsBase.params` for `TransformedDistribution`, and `Base.maximum, Base.minimum` for `UnivariateTransformedDistribution`.

## Up for discussion
- [x] `StatsBase.params` only returns the parameters of the underlying base distribution, i.e. `params(td.dist)`. In the case where in addition the `Bijector` has parameters, is it then expected that these are also included in the return-value of `params`? Uncertain what makes the most sense. If so, this does increase the complexity of this PR a fair bit as it's not 100% clear how we'd do that in a nice way.


arnavs:
I think just having the dist params is fine for now, but it would probably be ideal down the line to include the transform ones as well (though potentially in a different method.)

The reason I opened this issue (#85), by the way, was to check compatibility with `Expectations.jl`. This was per a JOSS reviewer comment from @oxinabox. I think it opens a broader question of whether we can add this sort of expectation functionality to Turing.

torfjelde:
Should be good to go :+1:

torfjelde:
Should be good now, as long as the tests pass :+1:

PR #91, opened 2020-03-17
ReverseDiff support

mohamed82008:
This PR adds support and tests for ReverseDiff.


src/compat/reversediff.jl
```diff
@@ -0,0 +1,78 @@
...
+end
+
+@noinline function special_forward_exec!(instruction::SpecialInstruction{typeof(replace_diag)})
+    output, input = instruction.output, instruction.input
+    out_value = replace_diag(ReverseDiff.value(input[1]), ReverseDiff.value(input[2]))
+    ReverseDiff.value!(output, out_value)
+    return nothing
+end
+
+# Work around to stop TrackedReal of Inf and -Inf from producing NaN in the derivative
```
torfjelde: Is this related to Bijectors.jl or should it be in DistributionsAD?

src/compat/reversediff.jl
```diff
@@ -0,0 +1,78 @@
...
+Base.:*(A::AbstractTriangular, B::RTV{<:Any, D}) where {D} = record_mul(A, B, D)
+Base.:*(A::AbstractTriangular, B::RTM{<:Any, D}) where {D} = record_mul(A, B, D)
+function Base.:*(A::Adjoint{T, <:AbstractTriangular{T}}, B::RTM{<:Any, D}) where {T, D}
+    return record_mul(A, B, D)
+end
+function Base.:*(A::Adjoint{T, <:AbstractTriangular{T}}, B::RTV{<:Any, D}) where {T, D}
+    return record_mul(A, B, D)
+end
+
+function replace_diag(::typeof(log), X::RTM{<:Any, D}) where {D}
```
torfjelde: May I ask why these impls of `replace_diag` are necessary?

src/compat/reversediff.jl
```diff
@@ -0,0 +1,78 @@
...
+    function back(∇)
+        g(i, j) = i == j ? ∇[i, j]*exp(X_value[i, j]) : ∇[i, j]
+        return g.(1:size(X_value, 1), (1:size(X_value, 2))')
+    end
+    out = ReverseDiff.track(out_value, D, tp)
+    ReverseDiff.record!(tp, ReverseDiff.SpecialInstruction, replace_diag, (exp, X), out, (back,))
+    return out
+end
+
+@noinline function ReverseDiff.special_reverse_exec!(instruction::SpecialInstruction{typeof(replace_diag)})
```
torfjelde: Same here, just curious why these impls are neceesary

src/compat/reversediff.jl
mohamed82008: Fair point, this could be in DistributionsAD.

src/compat/reversediff.jl
mohamed82008: I couldn't find a simple implementation that maintains a TrackedArray in Tracker and ReverseDiff. So I went with a custom adjoint for this function since it is simple enough to define. This ugly syntax is just a custom adjoint in ReverseDiff.

src/compat/reversediff.jl
mohamed82008: This is part of the custom adjoint.

src/compat/reversediff.jl
devmotion: Does indexing with `diagind` not work for TrackedArray and ReverseDiff? I thought it would be problematic only for Zygote due to the mutation.

mohamed82008: Mutation is not allowed with TrackedArrays too.

devmotion: I see :+1:

torfjelde: Ah, sorry; didn't realize `replace_diag` was a function from Bijectors.jl; makes sense :+1: 
Would be awesome with a comment stating why it's necessary:)

mohamed82008: Sure

PR #92, opened 2020-03-18
Fix import of ReverseDiff

devmotion:
This PR fixes errors such as
```julia
┌ Warning: Error requiring ReverseDiff from Bijectors:
│ LoadError: ArgumentError: Package Bijectors does not have ReverseDiff in its dependencies:
│ - If you have Bijectors checked out for development and have
│   added ReverseDiff as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with Bijectors
```
when loading, e.g., Turing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.6.3"
+version = "0.6.4"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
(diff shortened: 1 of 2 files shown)


devmotion:
It seems the Zygote tests took too long to run. However, Zygote support and integration is completely unrelated to this PR. My general impression is that it takes a lot of time to run the CI tests - I'm not sure if splitting some test sets could be helpful?

yebai:
@devmotion Would you like to fix the git conflicts before merging?

devmotion:
Sure, I'll fix the conflicts. It seems @mohamed82008 has already opened and merged another PR to address the problem, but it seems there are still some accesses of ReverseDiff at the top of the file - IMO one should add a `import .ReverseDiff` statement at the top (and probably move the other imports there as well).

devmotion:
See: https://github.com/TuringLang/Bijectors.jl/pull/97

PR #95, opened 2020-03-19
`NamedBijector`: a bijector for working with named tuples

torfjelde:
There's been a lot of interest in a way to use Bijectors.jl together with named tuples rather than just vectors, similar to the approach in TransformVariables.jl.

`NamedBijector` is an initial attempt at making something like a wrapper around `Bijector{N}`, thus inheriting a lot of the nice features from our previous work, but now taking named tuples as inputs rather than vectors.

It's functional but would like to get some thoughts from people before committing to this approach. What do you think?

## Example

``` julia
julia> using Bijectors
                                                                                                                                                                                                                                                                               
julia> using Bijectors: Exp, Log, NamedBijector
                                                                                                                                                                                                                                                                               
julia> b = NamedBijector((a = Exp(), b = Log()))
NamedBijector{NamedTuple{(:a, :b),Tuple{Exp{0},Log{0}}}}((a = Exp{0}(), b = Log{0}()))                                                                                                                                                                                         
                                                                                                                                                                                                                                                                               
julia> x = (a = 0.0, b = exp(1.0))
(a = 0.0, b = 2.718281828459045)                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                               
julia> b(x)
(a = 1.0, b = 1.0)                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                               
julia> inv(b)(b(x))
(a = 0.0, b = 2.718281828459045)                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                               
julia> logabsdetjac(b, x)
-1.0

julia> xs = (a = [0.0, 1.0], b = exp.([0.0, 1.0]))
(a = [0.0, 1.0], b = [1.0, 2.718281828459045])

julia> b(xs)
(a = [1.0, 2.718281828459045], b = [0.0, 1.0])

julia> inv(b)(b(xs))
(a = [0.0, 1.0], b = [1.0, 2.718281828459045])

julia> logabsdetjac(b, xs)
2-element Array{Float64,1}:
 0.0
 0.0

julia> (inv(b) ∘ b)(x) == x
true

julia> (b ∘ inv(b) ∘ b)(x)
(a = 1.0, b = 1.0)
```


cscherrer:
This looks great for the case where the transformation is diagonal, decomposing to a product of transformations that apply per-component. Any ideas for cases where there are dependencies?

Say, if you wanted to transform
```julia
a ~ Uniform(0,1)
b ~ Uniform(0,a)
```

to ℝ²

torfjelde:
So you could actually construct that (awkwardly) using a `CouplingLayer` from #48 , though it would require making a `CouplingLayer` for named tuples rather than vectors (which in the case of `CouplingLayer` significantly simplifies the implementation, actually). 

E.g. in the case where we have a 2D vector of your example, we can do
```julia
julia> using Bijectors

julia> using Bijectors: PartitionMask, CouplingLayer, Logit

julia> a = rand(Uniform())
0.9291003320904823

julia> b = rand(Uniform(0, a))
0.33723045188242684

julia> mask = PartitionMask(2, [2], [1]); # <= going to use `x[1]` to parameterize transformation of `x[2]`

julia> bij_b = CouplingLayer(a -> Logit(0., a[1]), mask, identity);

julia> bij_a = stack(Logit(0., 1.), Identity{0}());

julia> to_unconstrained = bij_a ∘ bij_b; # first transform `x[2]` given `x[1]` and then transform `x[1]`

julia> to_unconstrained([a, b])
2-element Array{Float64,1}:
  2.5729509834803506
 -0.5625202836554662
```
In the case of using vector-inputs we need to make use of masks for partitioning the vector into separate pieces, etc. In the case of using named tuples we don't have to do all that!
So we can definitively add such a bijector:) It would probably look something like
```julia
julia> using Bijectors

julia> using Bijectors: PartitionMask, CouplingLayer, Logit

julia> a = rand(Uniform())
0.9291003320904823

julia> b = rand(Uniform(0, a))
0.33723045188242684

julia> bij_b = NamedCouplingLayer(a -> Logit(0., a[1]), (b = (a:, )), identity); # specify that `b` depends on `a`

julia> bij_a = NamedBijector(a = Logit(0., 1.),  b = Identity{0}());

julia> to_unconstrained = bij_a ∘ bij_b; # first transform `b` given `a` and then transform `a`

julia> to_unconstrained((a = a, b = b))
  (a = 2.5729509834803506, b = -0.5625202836554662)
```

Note: the reason why the #48 hasn't been merged yet is because I'm still not _quite_ happy with the interface, which you can probably tell from above. But it's working so it's just a matter of "finishing" touches:)

cscherrer:
The `NamedBijector` inferface you've got is really slick, but yeah the Coupling stuff could maybe use some polishing. Not clear to me how it should look though.

Have you benchmarked `NamedBijector`s? This will be in inner loops, so hopefully at least comparable to alternatives :)

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,31 @@
+struct NamedBijector{Bs} <: AbstractBijector
```
devmotion: I just thought of the following three things.
1. What is the advantage of a `NamedBijector` over a named tuple of `Bijector`s?
2. In the function signatures you check that the names of the tuple of bijector and the tuple of inputs are identical - however, that is not a strict requirement for the implementation to work it seems, the names of the inputs could be just a (possibly permuted) subset of the names of the bijectors. Do you actually want to disallow this possibility?
3. If you actually want to dispatch on the names of the bijectors, maybe it would be helpful to define
   ```julia
    struct NamedBijector{names,B<:NamedTuple{names}}
        bijectors::B
    end
    ```
    Then one could dispatch on `names` only, e.g.,
    ```julia
    @generated function (b::NamedBijector{names})(x::NamedTuple{names}) where {names}
        return :(($([:($n = b.bs.$n(x.$n)) for n in names]...), ))
    end
    ```

torfjelde: 1. Fair point, though I think it's useful in case we might want to define special behaviors when using named tuples. Say we need something like the `NamedCouplingLayer` bijector described above, we can define something like a `AbstractNamedBijector{names}` and have `NamedBijector` and `NamedCouplingLayer` be concrete subtypes. Then we can define compositions and such between `AbstractNamedBijector`s rather than on a case-by-case basis.
2.  Good point! :+1: I'll change that for sure.
3. Good idea:) And I think it would be a good idea to put this in a `AbstractNamedBijector{names}` too, as mentioned in (1).

torfjelde:
> The NamedBijector inferface you've got is really slick, but yeah the Coupling stuff could maybe use some polishing. Not clear to me how it should look though.
Have you benchmarked NamedBijectors? This will be in inner loops, so hopefully at least comparable to alternatives :)

Haven't benchmarked yet, but we're using generated functions so for cases of `NamedTuple{names}` where `names` never changes, it should be just as fast as if you did it by hand:)

src/bijectors/named_bijector.jl
torfjelde: I think I've now addressed these comments :+1: Thanks!

torfjelde:
So the impl of the above is now:
```julia
julia> using Bijectors: NamedCoupling, Logit

julia> a = rand(Uniform())
0.3242840704140346

julia> b = rand(Uniform())
0.2923666685362576

julia> bij_a = NamedBijector((a = Logit(0., 1.), ));

julia> bij_b = NamedCoupling(:b, (:a, ), a -> Logit(0., a));

julia> to_unconstrained = bij_a ∘ bij_b; # first transform `b` given `a` and then transform `a`

julia> to_unconstrained((a = a, b = b))
(a = -0.7341528730193115, b = 2.2148573542215177)
```
And just to verify:
```julia
julia> using BenchmarkTools

julia> @btime to_unconstrained((a = a, b = b))
  260.597 ns (3 allocations: 96 bytes)
(a = -0.7341528730193115, b = 2.2148573542215177)

julia> @btime (a = $(Logit(0., 1.))(a), b = $(Logit(0, a))(b))
  284.699 ns (6 allocations: 160 bytes)
(a = -0.7341528730193115, b = 2.2148573542215177)
```

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,230 @@
...
+@generated function (b::NamedBijector{names})(x::NamedTuple) where {names}
+    return :(merge(x, ($([:($n = b.bs.$n(x.$n)) for n in names]...), )))
+end
+
+@generated function Base.inv(b::NamedBijector{names}) where {names}
+    return :(NamedBijector(($([:($n = inv(b.bs.$n)) for n in names]...), )))
+end
+
+@generated function logabsdetjac(b::NamedBijector{names}, x::NamedTuple) where {names}
+    return :(sum([$([:(logabsdetjac(b.bs.$n, x.$n)) for n in names]...), ]))
```
mohamed82008, line 43: I think we don't need to form a vector here, unnecessary allocations. Just use `+`.

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,230 @@
...
+### `NamedInverse` ###
+######################
+"""
+    NamedInverse <: AbstractNamedBijector
+
+Represents the inverse of a `AbstractNamedBijector`, similarily to `Inverse` for `Bijector`.
+
+See also: [`Inverse`](@ref)
+"""
+struct NamedInverse{B<:AbstractNamedBijector} <: AbstractNamedBijector
```
mohamed82008, line 57: Do we really need this? Can't we just define the inverse to be recursive?

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,230 @@
...
+        expr = :(cb.bs[$i]($expr))
+    end
+    return expr
+end
+
+function logabsdetjac(cb::NamedComposition, x)
+    y, logjac = forward(cb.bs[1], x)
+    for i = 2:length(cb.bs)
+        res = forward(cb.bs[i], y)
+        y = res.rv
```
mohamed82008, line 114: this line is not needed

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,230 @@
...
+
+NamedCoupling(target, deps, f::F) where {F} = NamedCoupling{target, deps, F}(f)
+function NamedCoupling(::Val{target}, ::Val{deps}, f::F) where {target, deps, F}
+    return NamedCoupling{target, deps, F}(f)
+end
+
+coupling(b::NamedCoupling) = b.f
+# For some reason trying to use the parameteric types doesn't always work
+# so we have to do this weird approach of extracting type and then index `parameters`.
+target(b::NamedCoupling) = typeof(b).parameters[1]
```
mohamed82008, line 206: This is not good. Please use type parameters.

mohamed82008 (commented):
Looks great other than the few comments I left. Thanks @torfjelde !

src/bijectors/named_bijector.jl
torfjelde, line 43:
```suggestion
    exprs = [:(logabsdetjac(b.bs.$n, x.$n)) for n in names]
    return :(+($(exprs...)))
```

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+
+@inline names_to_bijectors(b::NamedBijector) = b.bs
+
+@generated function (b::NamedBijector{names})(x::NamedTuple) where {names}
+    return :(merge(x, ($([:($n = b.bs.$n(x.$n)) for n in names]...), )))
+end
+
+@generated function Base.inv(b::NamedBijector{names}) where {names}
+    return :(NamedBijector(($([:($n = inv(b.bs.$n)) for n in names]...), )))
+end
```
torfjelde, line 40: See here @mohamed82008

src/bijectors/named_bijector.jl
torfjelde: Haha, well, we actually do that already, and this is just left-over that I forgot to remove!

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+See also: [`Inverse`](@ref)
+"""
+struct NamedInverse{B<:AbstractNamedBijector} <: AbstractNamedBijector
+    orig::B
+end
+Base.inv(nb::AbstractNamedBijector) = NamedInverse(nb)
+Base.inv(ni::NamedInverse) = ni.orig
+
+logabsdetjac(ni::NamedInverse, y::NamedTuple) = -logabsdetjac(inv(ni), ni(y))
+
```
torfjelde, line 65:
```suggestion
```

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+NamedCoupling(target, deps, f::F) where {F} = NamedCoupling{target, deps, F}(f)
+function NamedCoupling(::Val{target}, ::Val{deps}, f::F) where {target, deps, F}
+    return NamedCoupling{target, deps, F}(f)
+end
+
+coupling(b::NamedCoupling) = b.f
+# For some reason trying to use the parameteric types doesn't always work
+# so we have to do this weird approach of extracting type and then index `parameters`.
+target(b::NamedCoupling) = typeof(b).parameters[1]
```
torfjelde, line 207:
```suggestion
target(b::NamedCoupling{target}) where {target} = target
```

src/bijectors/named_bijector.jl
torfjelde: Done :+1:

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
+abstract type AbstractNamedBijector <: AbstractBijector end
+
+forward(b::AbstractBijector, x) = (rv = b(x), logabsdetjac = logabsdetjac(b, x))
+
```
devmotion, line 4: Isn't this the wrong file for such a generic definition? Or is it supposed to be 
```suggestion
forward(b::AbstractNamedBijector, x) = (rv = b(x), logabsdetjac = logabsdetjac(b, x))

```
?

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+julia> (a = 2 * x.a, b = exp(x.b), c = x.c)
+(a = 2.0, b = 1.0, c = 42.0)
+```
+"""
+struct NamedBijector{names, Bs<:NamedTuple{names}} <: AbstractNamedBijector
+    bs::Bs
+end
+
+@inline names_to_bijectors(b::NamedBijector) = b.bs
```
devmotion, line 32: I'm a bit surprised if `@inline` would be needed here? Seems like something that the compiler should inline anyway, I guess?

torfjelde: 100% :+1: Left-over from me with worse understanding of Julia:)

torfjelde:
```suggestion
names_to_bijectors(b::NamedBijector) = b.bs
```

src/bijectors/named_bijector.jl
devmotion, line 40: Is the generated function actually needed here? Would be interesting to benchmark if it is faster than just
```suggestion
Base.inv(b::NamedBijector) = NamedBijector(map(inv, b.bs))
```
or
```suggestion
function Base.inv(b::NamedBijector{names}) where {names}
	bijectors = map(inv, b.bs)
	return NamedBijector{names,typeof(bijectors)}(bijectors)
end
```

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+
+@generated function (b::NamedBijector{names})(x::NamedTuple) where {names}
+    return :(merge(x, ($([:($n = b.bs.$n(x.$n)) for n in names]...), )))
+end
```
devmotion, line 36: Maybe just fully generate the resulting named tuple based on the type of `x` if you are defining it in a generated function anyway?

src/bijectors/named_bijector.jl
torfjelde: That's one thing, but there's also the issue of type-stability. `map(inv, b.bs)` will fail if `bs` is large :confused:

src/bijectors/named_bijector.jl
torfjelde: Good suggestion :+1: Will do!

torfjelde: Diddit!

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+
+function (cb::NamedComposition{<:AbstractArray{<:AbstractNamedBijector}})(x)
+    @assert length(cb.bs) > 0
+    res = cb.bs[1](x)
+    for b ∈ Base.Iterators.drop(cb.bs, 1)
+        res = b(res)
+    end
+
+    return res
+end
```
devmotion, line 112: Isn't this just
```suggestion
(cb::NamedComposition{<:AbstractArray{<:AbstractNamedBijector}})(x) = foldl(|>, cb.bs; init=x)
```
?

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,231 @@
...
+
+@generated function (cb::NamedComposition{T})(x) where {T<:Tuple}
+    @assert length(T.parameters) > 0
+    expr = :(x)
+    for i in 1:length(T.parameters)
+        expr = :(cb.bs[$i]($expr))
+    end
+    return expr
+end
```
devmotion, line 121: I'm also wondering how much you actually gain by using a generated function here instead of also
```suggestion
(cb::NamedComposition{<:Tuple})(x) = fold(|>, cb.bs; init=x)
```

src/bijectors/named_bijector.jl
torfjelde: Type-stability issue if output is different type than input, e.g. if we're using `TrackedArray` for parameters of the bijectors but not the input.

src/bijectors/named_bijector.jl
torfjelde: Type-stability issue if output is different type than input, e.g. if we're using TrackedArray for parameters of the bijectors but not the input.

src/bijectors/named_bijector.jl
torfjelde: Btw, this is why this particular impl was introduced for `Composition`, and so I used the same code for `NamedComposition`.

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,243 @@
...
+coupling(b::NamedCoupling) = b.f
+# For some reason trying to use the parameteric types doesn't always work
+# so we have to do this weird approach of extracting type and then index `parameters`.
+target(b::NamedCoupling{Target}) where {Target} = Target
+deps(b::NamedCoupling{<:Any, Deps}) where {Deps} = Deps
+
+@generated function (nc::NamedCoupling{target, deps, F})(x::NamedTuple) where {target, deps, F}
+    return quote
+        b = nc.f($([:(x.$d) for d in deps]...))
+        return merge(x, ($target = b(x.$target), ))
```
devmotion, line 225: Similar as above, I guess you could get rid of `merge`.

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,243 @@
...
+    end
+end
+
+@generated function (ni::NamedInverse{<:NamedCoupling{target, deps, F}})(
+    x::NamedTuple
+) where {target, deps, F}
+    return quote
+        b = ni.orig.f($([:(x.$d) for d in deps]...))
+        return merge(x, ($target = inv(b)(x.$target), ))
```
devmotion, line 234: Same here.

src/bijectors/named_bijector.jl
devmotion: If there are type stability problems with TrackedArray, you get them for sure with the current implementation here as well, don't you? (Below you avoid the issue by building one long expression)

src/bijectors/named_bijector.jl
torfjelde: Actually, is there any point? `merge` is already implemented using generated functions.

src/bijectors/named_bijector.jl
torfjelde: See above response.

src/bijectors/named_bijector.jl
torfjelde: Second thought: might revert this change since `merge` is already implemented using `@generated`.

src/bijectors/named_bijector.jl
devmotion: I would assume that it's still slightly faster (but maybe negligible). Only benchmarking could tell though.

src/bijectors/named_bijector.jl
devmotion: For tuples (or at least small tuples) apparently the compiler can infer the output even if the output changes:
```julia
julia> f(x) = foldl(|>, x; init=1)

julia> @code_warntype f((x -> 2.0 * x, x -> x + 1))
Variables
  x::Core.Compiler.Const((var"#31#33"(), var"#32#34"()), false)

Body::Float64
1 ─ %1 = (:init,)::Core.Compiler.Const((:init,), false)
│   %2 = Core.apply_type(Core.NamedTuple, %1)::Core.Compiler.Const(NamedTuple{(:init,),T} where T<:Tuple, false)
│   %3 = Core.tuple(1)::Core.Compiler.Const((1,), false)
│   %4 = (%2)(%3)::Core.Compiler.Const((init = 1,), false)
│   %5 = Core.kwfunc(Main.foldl)::Core.Compiler.Const(Base.var"#foldl##kw"(), false)
│   %6 = (%5)(%4, Main.foldl, Main.:|>, x)::Core.Compiler.Const(3.0, false)
└──      return %6
```

devmotion: Though maybe the example is too simple since the compiler even figures out the result.

devmotion: More realistic maybe:
```julia
julia> g(x, fs) = foldl(|>, fs; init=x)

julia> @code_warntype g(1, (x -> 2.0 * x, x -> x + 1))
Variables
  #self#::Core.Compiler.Const(g, false)
  x::Int64
  fs::Core.Compiler.Const((var"#35#37"(), var"#36#38"()), false)

Body::Float64
1 ─ %1 = (:init,)::Core.Compiler.Const((:init,), false)
│   %2 = Core.apply_type(Core.NamedTuple, %1)::Core.Compiler.Const(NamedTuple{(:init,),T} where T<:Tuple, false)
│   %3 = Core.tuple(x)::Tuple{Int64}
│   %4 = (%2)(%3)::NamedTuple{(:init,),Tuple{Int64}}
│   %5 = Core.kwfunc(Main.foldl)::Core.Compiler.Const(Base.var"#foldl##kw"(), false)
│   %6 = (%5)(%4, Main.foldl, Main.:|>, fs)::Float64
└──      return %6
```

src/bijectors/named_bijector.jl
torfjelde:
```julia
julia> using Bijectors: NamedBijector, Exp, Log

julia> using BenchmarkTools

julia> b = NamedBijector((a = Exp(), b = Log()))
NamedBijector{(:a, :b),NamedTuple{(:a, :b),Tuple{Exp{0},Log{0}}}}((a = Exp{0}(), b = Log{0}()))

julia> x = (a = 0.0, b = exp(1.0))
(a = 0.0, b = 2.718281828459045)

julia> @generated function (b::NamedBijector{names})(x::NamedTuple) where {names}
           return :(merge(x, ($([:($n = b.bs.$n(x.$n)) for n in names]...), )))
       end

julia> @btime $b($x)
  7.862 ns (0 allocations: 0 bytes)
(a = 1.0, b = 1.0)

julia> @generated function (b::NamedBijector{names1})(
           x::NamedTuple{names2}
       ) where {names1, names2}
           exprs = []
           for n in names2
               if n in names1
                   # Use processed value
                   push!(exprs, :($n = b.bs.$n(x.$n)))
               else
                   # Use existing value
                   push!(exprs, :($n = x.$n))
               end
           end
           return :($(exprs...), )
       end

julia> @btime $b($x)
  8.765 ns (0 allocations: 0 bytes)
(a = 1.0, b = 1.0)
```

src/bijectors/named_bijector.jl
torfjelde: Only if the first bijector does not have tracked but later bijectors do, so I'd argue that the current impl covers more cases than the above, no?

src/bijectors/named_bijector.jl
torfjelde: Uhh, I can't make it _not_ work, so you might be right :flushed: 
If so, that's waaaay better! I'm down for changing it :+1:

src/bijectors/named_bijector.jl
devmotion: I guess one should be careful with benchmarks in the order of ns (usually that means most/everything just compiled away) but at least there's no strong argument for making the code more complicated and uglier anymore :+1:

src/bijectors/named_bijector.jl
devmotion: Yes, I guess it should be fine since the first call is handled separately. The repeated assignment of `res` could still be problematic depending on the output of the different bijectors, but I don't know if/how this could be fixed :shrug:

src/bijectors/named_bijector.jl
devmotion: Yeah let's not make the code more complicated than needed

src/bijectors/named_bijector.jl
torfjelde: Yeah, 100% agree; it was not meant as absolute evidence, but since it will make the code more complicated I'm okay with sticking with `merge` until the opposite is proven.

src/bijectors/named_bijector.jl
mohamed82008: Awesome!

src/bijectors/named_bijector.jl
```diff
@@ -0,0 +1,236 @@
...
+    @assert length(cb.bs) > 0
+    res = cb.bs[1](x)
+    for b ∈ Base.Iterators.drop(cb.bs, 1)
+        res = b(res)
+    end
+
+    return res
+end
+
+(cb::NamedComposition{<:Tuple})(x) = fold(|>, cb.bs; init=x)
```
devmotion, line 114: Of course, there was a typo...
```suggestion
(cb::NamedComposition{<:Tuple})(x) = foldl(|>, cb.bs; init=x)
```

torfjelde:
Are people happy with this or is there something to be fixed before approval?

devmotion (approved):
LGTM

PR #99, opened 2020-04-04
Making it a bit easier to override default `link` etc. using `bijector`

### LINKED ISSUE #93: Distributions.product_distribution - Bijectors.link & Bijectors.invlink transforms
Hi guys, 

I need to work with with a product distribution of uniforms . When transforming and inverse transforming the random variables, I dont get any error messages from Bijectors, but transforms and inverse transforms do not seem to work correctly. Minimum working example:

```
using Distributions, Bijectors #Bijectors v0.6.3, Distributions v0.22.6
x = [.5, .5]
distr = Distributions.product_distribution( [Uniform(-1,1) for i in 1:2] )
y = Bijectors.link(distr, x) #still [.5, .5]
Bijectors.invlink(distr, [150., 150]) #still [150., 150] ~ not in support
```

Apologies if I oversee something trivial.

torfjelde:
Making it a bit easier to override default `link` etc. using `bijector`, e.g. to fix #93 we just need
```julia
bijector(d::Distributions.Product) = stack(bijector.(d.vs)...) # or Stacked(bijector.(d.vs))
```
by replacing the default impls of `link`, `invlink` and `logpdf_with_trans` with impls using `bijector(d)` under the hood. 

Will help to stop issues like #100 occurring in the future :+1:

NB: At the moment some AD tests are failing! Trying to figure out why but help would be appreciated.

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -156,7 +156,14 @@ end
 ## Multivariate
 
 function logpdf_with_trans(
-    dist::Distributions.Product,
+    dist::Distributions.Product{Discrete},
+    x::AbstractVecOrMat{<:Real},
+    istrans::Bool,
+)
+    return logpdf(dist, x)
+end
+function logpdf_with_trans(
+    dist::Distributions.Product{Continuous},
     x::AbstractVector{<:Real},
     istrans::Bool,
 )
@@ -165,24 +172,25 @@ function logpdf_with_trans(
     end)
 end
 function logpdf_with_trans(
-    dist::Distributions.Product,
+    dist::Distributions.Product{Continuous},
     x::AbstractMatrix{<:Real},
     istrans::Bool,
 )
-    return map(eachcol(x)) do x
-        logpdf_with_trans(dist, x, istrans)
+    return map(1:size(x,2)) do i
+        c = x[:,i]
+        sum(maporbroadcast(dist.v, c) do d, x
+            logpdf_with_trans(d, x, istrans)
+        end)
     end
 end
 
-link(dist::Distributions.Product{Discrete}, x::AbstractVector{<:Real}) = copy(x)
+link(dist::Distributions.Product{Discrete}, x::AbstractVecOrMat{<:Real}) = copy(x)
 function link(
     dist::Distributions.Product{Continuous},
     x::AbstractVector{<:Real},
 )
     return maporbroadcast(link, dist.v, x)
 end
-
-link(dist::Distributions.Product{Discrete}, x::AbstractMatrix{<:Real}) = copy(x)
 function link(
     dist::Distributions.Product{Continuous},
     x::AbstractMatrix{<:Real},
@@ -192,15 +200,13 @@ function link(
     end
 end
 
-invlink(dist::Distributions.Product{Discrete}, x::AbstractVector{<:Real}) = copy(x)
+invlink(dist::Distributions.Product{Discrete}, x::AbstractVecOrMat{<:Real}) = copy(x)
 function invlink(
     dist::Distributions.Product{Continuous},
     x::AbstractVector{<:Real},
 )
     return maporbroadcast(invlink, dist.v, x)
 end
-
-invlink(dist::Distributions.Product{Discrete}, x::AbstractMatrix{<:Real}) = copy(x)
 function invlink(
     dist::Distributions.Product{Continuous},
     x::AbstractMatrix{<:Real},
@@ -213,9 +219,6 @@ end
 function maporbroadcast(f, dists::AbstractArray, x::AbstractArray)
     # Broadcasting here breaks Tracker for some reason
     return map(f, dists, x)
-end
-function maporbroadcast(f, dists::AbstractVector, x::AbstractMatrix)
-    return map(x -> sum(maporbroadcast(f, dists, x)), eachcol(x))
 end
 
 const SimplexDistribution = Union{Dirichlet}
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -257,7 +257,7 @@ end
 # Positive definite #
 #####################
 
-const PDMatDistribution = Union{InverseWishart, Wishart}
+const PDMatDistribution = Union{InverseWishart, Wishart, MatrixBeta}
 
 # link(d::PDMatDistribution, X::AbstractMatrix{<:Real}) = PDBijector()(X)
 # invlink(d::PDMatDistribution, Y::AbstractMatrix{<:Real}) = inv(PDBijector())(Y)
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -1,5 +1,33 @@
 module Bijectors
 
+#=
+  NOTE: Codes below are adapted from
+  https://github.com/brian-j-smith/Mamba.jl/blob/master/src/distributions/transformdistribution.jl
+  The Mamba.jl package is licensed under the MIT License:
+  > Copyright (c) 2014: Brian J Smith and other contributors:
+  >
+  > https://github.com/brian-j-smith/Mamba.jl/contributors
+  >
+  > Permission is hereby granted, free of charge, to any person obtaining
+  > a copy of this software and associated documentation files (the
+  > "Software"), to deal in the Software without restriction, including
+  > without limitation the rights to use, copy, modify, merge, publish,
+  > distribute, sublicense, and/or sell copies of the Software, and to
+  > permit persons to whom the Software is furnished to do so, subject to
+  > the following conditions:
+  >
+  > The above copyright notice and this permission notice shall be
+  > included in all copies or substantial portions of the Software.
+  >
+  > THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+  > EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+  > MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+  > IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+  > CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+  > TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+  > SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+=#
+
 using Reexport, Requires
 @reexport using Distributions
 using StatsFuns
@@ -33,7 +61,6 @@ export  TransformDistribution,
         DistributionBijector,
         bijector,
         transformed,
-        TransformedDistribution,
         UnivariateTransformed,
         MultivariateTransformed,
         logpdf_with_jac,
@@ -52,18 +79,16 @@ _debug(str) = @debug str
 _eps(::Type{T}) where {T} = T(eps(T))
 _eps(::Type{Real}) = eps(Float64)
 _eps(::Type{<:Integer}) = eps(Float64)
-_istracked(::Any) = false
+
+function _clamp(x, a, b)
+    T = promote_type(typeof(x), typeof(a), typeof(b))
+    ϵ = _eps(T)
+    clamped_x = ifelse(x < a, convert(T, a), ifelse(x > b, convert(T, b), x))
+    DEBUG && _debug("x = $x, bounds = $((a, b)), clamped_x = $clamped_x")
+    return clamped_x
+end
 
 function mapvcat(f, args...)
-    out = map(f, args...)
-    if _istracked(out)
-        init = vcat(out[1])
-        return reshape(reduce(vcat, drop(out, 1); init = init), size(out))
-    else
-        return out
-    end
-end
-function mapvcat2(f, args...)
     out = map(f, args...)
     init = vcat(out[1])
     return reshape(reduce(vcat, drop(out, 1); init = init), size(out))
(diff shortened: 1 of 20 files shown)


src/transformed_distribution.jl
```diff
@@ -33,8 +33,9 @@ transformed(d) = transformed(d, bijector(d))
 
 Returns the constrained-to-unconstrained bijector for distribution `d`.
 """
-bijector(d::Distribution) = DistributionBijector(d)
-bijector(d::UnivariateDistribution) = TruncatedBijector(minimum(d), maximum(d))
+bijector(d::UnivariateDistribution) = Identity{0}()   # default to identity
```
mohamed82008, line 36: I suggest we add an assertion here that the domain is unbounded. This will avoid a silent wrong result when the domain is bounded.

mohamed82008: I suggest we throw as a fallback for multi and matrix variate distributions to avoid a silent wrong result.

torfjelde: > I suggest we add an assertion here that the domain is unbounded. This will avoid a silent wrong result when the domain is bounded.

In the sense that we return identity iff it's unbounded, otherwise throw an error?
I think then we should just do something like
```julia
bijector(d::UnivariateDiscreteDistribution) = Identity{0}()
bijector(d::UnivariateContinuousDistribution) = TruncatedBijector(minimum(d), maximum(d))
```
and remove the impl for `TransformDistributions`. This way we'll get the expected behaviour if we have a well-defined `minimum` and `maximum` (irregardless of whether or not it's bounded), but will throw an error if this information isn't available.

> I suggest we throw as a fallback for multi and matrix variate distributions to avoid a silent wrong result.

So you're suggest _not_ implementing it for multivariate and matrix variate? And just letting a `unknown method` error be thrown? If so, I'm down for this. We just need to document this in the Turing documentation so that people are aware that if they want to create custom distributions they also need to define `bijector` for it if they're using AD-based samplers :+1:

src/interface.jl
```diff
@@ -105,7 +105,7 @@ logabsdetjacinv(b::Bijector, y) = logabsdetjac(inv(b), y)
 ##############################
 
 struct Identity{N} <: Bijector{N} end
-(::Identity)(x) = x
+(::Identity)(x) = copy(x)
```
torfjelde, line 108: Should this be copy or not? @mohamed82008 I noticed you had implemented some using `copy`, and I tried changing it to see if helped with the Zygote error (though it didn't seem to).

src/transformed_distribution.jl
mohamed82008: >So you're suggest not implementing it for multivariate and matrix variate? And just letting a unknown method error be thrown? If so, I'm down for this. We just need to document this in the Turing documentation so that people are aware that if they want to create custom distributions they also need to define bijector for it if they're using AD-based samplers 👍

I think this makes sense. We need to document how to write a custom distribution. Perhaps we can have a `domain` function that returns `Unbounded()`, `Simplex()`, `PD()`, `Positive()`, `Interval(lower, upper)`, or `Unknown`? Then if the user defines this function for the multivariate or matrix-variate distribution, we use the corresponding bijector accordingly. If `Unknown` we throw an unknown method error.

src/interface.jl
mohamed82008: Hmm not sure, if it worked previously then the `copy` is not necessary I suppose.

src/transformed_distribution.jl
torfjelde: I like that :+1: Could do some something like
```julia
domain(d::Distribution)
bijector(c::Constraint)

# Default impl for `Distribution`
bijector(d::Distribution) = bijector(domain(d))

# e.g. the following could replace `bijector(d::Exponential)`
domain(d::Exponential) = Positive()
bijector(::Positive) = Log{0}()
```

Though wouldn't it be better to _not_ have `Unknown`, and just let `domain` throw a method-not-implemented error?

mohamed82008: Ya sure `Unknown` is not necessary.

torfjelde: So one thing that annoys me a bit is that it _seems_ like it should be possible to define a really nice interface for working with constraints/domains, at which point you could actually encode the input- and output-space into the Bijector definition itself. And you should be able to do stuff like create product-spaces from different spaces, etc. 

But I guess we have to start somewhere, and just getting something that makes it easier for the user is I guess useful enough (in constrast to having the user figure out which Bijector they need).

torfjelde: Second thought; maybe we should delay the introduction of constraints to a different PR. I've done some work locally, but I run into issues where it's kind of unclear what `Bijector` to return, e.g. for a `VectorOfMultivariate`. The work here doesn't rely on introduction of constraints, so IMO we should just not do it here.

mohamed82008: Fair enough. Let me review it first.

src/transformed_distribution.jl
```diff
@@ -33,8 +33,10 @@ transformed(d) = transformed(d, bijector(d))
 
 Returns the constrained-to-unconstrained bijector for distribution `d`.
 """
-bijector(d::Distribution) = DistributionBijector(d)
-bijector(d::UnivariateDistribution) = TruncatedBijector(minimum(d), maximum(d))
+bijector(d::DiscreteUnivariateDistribution) = Identity{0}()
+bijector(d::ContinuousUnivariateDistribution) = TruncatedBijector(minimum(d), maximum(d))
+bijector(d::Distributions.Product) = stack(bijector.(d.v)...)
```
mohamed82008, line 38: This splatting is not good for large arrays commonly used with `Product` in Turing.

torfjelde: I'm fine with not splatting, i.e. doing `Stacked(bijector.(d.v))`, but to make things type-stable it seems like we have to splat, right? Or equivalently do `Stacked(tuple(bijector.(d.v)))` but that should be the same?

mohamed82008: I don't think type stability is a concern here since all of the elements of `d.v` are usually of the same type. `tuple` should be equally bad ya.

mohamed82008: Actually, `tuple` doesn't convert, `Tuple` does.

torfjelde:
So I've now fixed some of the failed tests and improved code re-use. But now I'm running into an issue with Zygote. The following tests fails:
```julia
┌ Info: Testing: (var"##x#698", var"##x#699")->begin
│         #= /home/tor/.julia/dev/Bijectors/test/ad/distributions.jl:23 =#
│         filldist(MvNormal(var"##x#698", var"##x#699"), 2)
└     end

signal (11): Segmentation fault
```

Also, it's not quite clear to me exactly what is failing in these tests due to the use of `DistSpec`. Is the use of `Distspec` "documented" somewhere?

mohamed82008:
No it is not documented. This test basically tests `filldist(MvNormal(m, v), 2)`. Try isolating the test and check which combination of types fails.

torfjelde:
> This test basically tests filldist(MvNormal(m, v), 2). Try isolating the test and check which combination of types fails.

I got that, but it's unclear exactly what methods are being tested. Will this test `logpdf` gradients wrt. parameters of distribution, gradient wrt. input, etc.

src/transformed_distribution.jl
torfjelde: > I don't think type stability is a concern here since all of the elements of d.v are usually of the same type.

Ah, if that's the case then it should be fine :+1:

mohamed82008:
Yes but for `logpdf_with_trans(d, invlink(d, y), true)` for input parameters and `y` not `logpdf`. All possible combinations and methods are tested.

torfjelde:
Uhmmm _what_?!
```julia
julia> using DistributionsAD, Bijectors, Tracker, ForwardDiff

julia> function f1(vs)
           n1 = vs[1]
           x = [reshape(vs[2:10], (3, 3)), reshape(vs[11:end], (3, 3))]

           dist = MatrixBeta(3, n1, 3.)
           b = bijector(dist)

           return logabsdetjac(b, x) # <= this line changes between impl
       end
f1 (generic function with 1 method)

julia> function f2(vs)
           n1 = vs[1]
           x = [reshape(vs[2:10], (3, 3)), reshape(vs[11:end], (3, 3))]

           dist = MatrixBeta(3, n1, 3.)
           b = bijector(dist)

           return logpdf(dist, x) # <= this line changes between impl
       end
f2 (generic function with 1 method)

julia> function f(vs)
           n1 = vs[1]
           x = [reshape(vs[2:10], (3, 3)), reshape(vs[11:end], (3, 3))]

           dist = MatrixBeta(3, n1, 3.)
           b = bijector(dist)

           return logpdf(dist, x) - logabsdetjac(b, x) # <= this line changes between impl
       end

julia> # offending input
       args = [3.0, 0.3239552320287513, 0.07451752964111555, -0.010516169325024586, 0.07451752964111555, 0.8201186717610907, -0.16101202850282761, -0.010516169325024586, -0.16101202850282761, 0.8397115540645331, 0.3239552320287513, 0.07451752964111555, -0.010516169325024586, 0.07451752964111555, 0.8201186717610907, -0.16101202850282761, -0.010516169325024586, -0.16101202850282761, 0.8397115540645331];

julia> grad_tracker_batch = Tracker.gradient(sum ∘ f1, args)[1];

julia> grad_forwarddiff_batch = ForwardDiff.gradient(sum ∘ f1, args);

julia> @assert grad_tracker_batch ≈ grad_forwarddiff_batch

julia> grad_tracker_batch = Tracker.gradient(sum ∘ f2, args)[1];

julia> grad_forwarddiff_batch = ForwardDiff.gradient(sum ∘ f2, args);

julia> @assert grad_tracker_batch ≈ grad_forwarddiff_batch

julia> grad_tracker_batch = Tracker.gradient(sum ∘ f, args)[1];

julia> grad_forwarddiff_batch = ForwardDiff.gradient(sum ∘ f, args);

julia> @assert grad_tracker_batch ≈ grad_forwarddiff_batch
ERROR: AssertionError: grad_tracker_batch ≈ grad_forwarddiff_batch
Stacktrace:
 [1] top-level scope at REPL[54]:

julia> grad_tracker_batch
Tracked 19-element Array{Float64,1}:
   2.993247962339498 
   0.2327826077125026
   8.62954023858255  
  -8.633701633695726 
   8.62954023858255  
  66.70296743738268  
 -68.33269711699127  
  -8.633701633695726 
 -68.33269711699127  
  71.58596958539151  
   0.2327826077125026
   8.62954023858255  
  -8.633701633695726 
   8.62954023858255  
  66.70296743738268  
 -68.33269711699127  
  -8.633701633695726 
 -68.33269711699127  
  71.58596958539151  

julia> grad_forwarddiff_batch
19-element Array{Float64,1}:
   2.993247962339498
   6.505523736444465
   8.629540238582548
  -8.633701633695727
   7.763808497667187
  68.6192892602518  
 -68.33269711699127 
  -8.665802204271673
 -67.84384900953366 
  72.82352124245504 
   6.505523736444465
   8.629540238582548
  -8.633701633695727
   7.763808497667187
  68.6192892602518  
 -68.33269711699127 
  -8.665802204271673
 -67.84384900953366 
  72.82352124245504
```
Notice the inputs are very different → not just a numerical error!

torfjelde:
Found the mistake! It was because
```julia
logabsdetjac(b::PDBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = map(X) do x
    logabsdetjac(b, x)
end
```
lead to `logabsdetjac` returning `Array{<:TrackedReal}` while `logpdf(dist, x)` return `TrackedArray`. Subtraction then lead to `TrackedArray{<:TrackedReal}` and this somehow screwed things up.

Replaced the impl by:
```julia
logabsdetjac(b::PDBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = mapvcat(X) do x
    logabsdetjac(b, x)
end
```
and it works now.

This also raises a question as to whether or not #102 is really a good idea :confused: It might lead to similar issues when some operations return tracked arrays while others return arrays with tracked real.

mohamed82008:
I think this is a bug in Tracker when subtracting a TrackedArray from an array of TrackedReals. We can define this correctly to return an array of TrackedReal. How is ReverseDiff doing here?

torfjelde:
Yeah, but it's a question of doing this for _all_ operations: https://github.com/FluxML/Tracker.jl/issues/4

Though _maybe_ for us it would be sufficient to only do so for addition? I'm afraid this will come back and bite us down the road though, when someone eventually decides to do something other than addition and `Tracker.grad` _silently_ returns the wrong result.

mohamed82008:
Ya this is a bad one. Perhaps `logpdf` here should also return an array of TrackedReals?

torfjelde:
I don't particularly like that either, e.g. a user wants to make a custom distributon, say, a normalizing flow, involving a lot of lin alg so they use Tracker with `TrackedArray` :confused: 
Difficult to see a way around it other than (1) addressing the short-coming of Tracker.jl, (2) keep it as it is now, despite perf issues, or (3) drop Tracker support. 

Also, the perf-issues are mainly _just_ for Tracker.jl, right? E.g. `mapvcat` which does nothing if not tracked.

mohamed82008:
I think we can keep Tracker but advise against it. For code that can easily fallback to arrays of tracked reals, we can advise the users to try ReverseDiff or Zygote. Tracker still has a large number of custom adjoints that make it useful sometimes to compare against. Soon enough ReverseDiff will catch up (I am a maintainer there now) and so we will have less and less need for Tracker. For now let's try to do the right thing inside Bijectors and not worry too much about what users write. But this is another argument to make ReverseDiff the default one if it doesn't have this problem. CC: @yebai

mohamed82008:
I think the problem here can be worked around if we also return an array of tracked reals from `logpdf` by removing the `mapvcat` there as well.

src/bijectors/distribution_bijector.jl
```diff
@@ -1,4 +1,3 @@
-# TODO: Deprecate?
```
torfjelde, line 1: At this point, `DistributionBijector` should really be deprecated. The idea was to use it as a placeholder while we implemented `Bijector` for all the different distributions. 

In this PR the goal is to basically make `link` and `invlink` nothing but wrappers around the `Bijector` interface. This then makes `DistributionBijector` completely unnecessary:)

src/Bijectors.jl
```diff
@@ -84,213 +109,163 @@ function eachcolmaphcat(f, x)
...
+function link(
+    dist::Distributions.Product,
+    x::AbstractMatrix{<:Real},
+)
+    return map(eachcol(x)) do x
+        link(dist, x)
+    end
 end
 
+function invlink(
```
torfjelde, line 192: These impls shouldn't be necessary, but instead be handled by the batch-version for `Stacked`, no?

src/bijectors/distribution_bijector.jl
mohamed82008: Well for `(Fill)(Vector|Matrix)Of(Univariate|Multivariate)` distributions, somewhat efficient `logpdf_with_trans`, `link` and `invlink` methods are there which are better than falling back on `Stacked`. `Stacked` wasn't really designed for this, think 1000s of distributions of the same type and all are univariate. The whole `ranges` field is unnecessary, not to mention it is a tuple, so the length is part of the type. We can make a different bijector for these distributions.

src/Bijectors.jl
mohamed82008: I am making these distributions' bijectors the `DistributionBijector` simply because I was too lazy to define a new bijector type but that's the correct approach.  See the above comment for details.

src/bijectors/distribution_bijector.jl
torfjelde: Can't the `ranges` field just be a single-element then, i.e. `1:end`?
And combine this with a specific impl when `dists` are `Fill`, we should obtain the same perf, right?

mohamed82008: Hmm I may have misinterpreted the `ranges` field! Let me take a look again in a bit.

mohamed82008:
This PR should now pass tests but the (Fill)(Vector|Matrix)Of(Univariate|Multivariate) distributions are not pretty. I have a plan to sort it out without introducing new bijectors, but this branch is usable now.

torfjelde:
It's still failing :confused:

yebai:
> EDIT: Granted, it's only for the logabsdetjac for PDBijector. Everything else works

@mohamed82008 Could you take a look at `PDBijector`? we'd like to get this PR merged quickly if possible. Thanks!

mohamed82008:
It should work with the `mt/neurips` branch of DistributionsAD and the `mt/array_funcs` branch of ReverseDiff. The RD PR will take a while to be reviewed so I will make a quick DAD PR (using the current RD) and release it first then this PR should pass.

mohamed82008:
https://github.com/TuringLang/DistributionsAD.jl/pull/70

mohamed82008:
I am working on removing `DistributionBijector` even for the special distributions in DAD. If tests pass here, we can merge and release and I will get rid of `DistributionBijector` in another PR.

torfjelde:
Sounds good :+1: And thanks you so much for fixing this mate!

Issue #100, opened 2020-04-04
Weird behavior from `link` and `invlink` for `Distributions.Product` with DistributionsAD


torfjelde:
Seems to be an issue with Bijectors.jl's `compat/distributionsad.jl`:
```julia
julia> using DistributionsAD, Bijectors

julia> dists = [Gamma(2, 3) for i = 1:2]
2-element Array{Gamma{Float64},1}:
 Gamma{Float64}(α=2.0, θ=3.0)
 Gamma{Float64}(α=2.0, θ=3.0)

julia> dist = product_distribution(dists)
Product{Continuous,Gamma{Float64},Array{Gamma{Float64},1}}(v=Gamma{Float64}[Gamma{Float64}(α=2.0, θ=3.0), Gamma{Float64}(α=2.0, θ=3.0)])

julia> x = rand(dist)
2-element Array{Float64,1}:
 3.3962613472010568
 4.504150811063049

julia> Bijectors.link(dist, x)
2-element Array{Float64,1}:
 1.222675222849948
 1.5049993740830023

julia> Bijectors.invlink(dist, Bijectors.link(dist, x))
2-element Array{Float64,1}:
 1.222675222849948
 1.5049993740830023
```
while on a project with Bijectors.jl only:
```julia
julia> using Bijectors

julia> dists = [Gamma(2, 3) for i = 1:2]
2-element Array{Gamma{Float64},1}:
 Gamma{Float64}(α=2.0, θ=3.0)
 Gamma{Float64}(α=2.0, θ=3.0)

julia> dist = product_distribution(dists)
Product{Continuous,Gamma{Float64},Array{Gamma{Float64},1}}(v=Gamma{Float64}[Gamma{Float64}(α=2.0, θ=3.0), Gamma{Float64}(α=2.0, θ=3.0)])

julia> x = rand(dist)

2-element Array{Float64,1}:
 11.597241830334823
 19.19914746393061

julia> Bijectors.link(dist, x)
2-element Array{Float64,1}:
 11.597241830334823
 19.19914746393061

julia> Bijectors.invlink(dist, Bijectors.link(dist, x))
2-element Array{Float64,1}:
 11.597241830334823
 19.19914746393061

(TestingBijectors) pkg> st
Project TestingBijectors v0.1.0
Status `/tmp/TestingBijectors/Project.toml`
  [76274a88] Bijectors v0.6.6
```

Behavior is caused by the following when used with DistributionsAD.jl:
```julia
[...]

PR #101, opened 2020-04-06
Quick-fix for #100

torfjelde:
While I'm finishing up #99, this works as a quick-fix for #100.

### DIFF
--- a/src/compat/distributionsad.jl
+++ b/src/compat/distributionsad.jl
@@ -254,7 +254,7 @@ function link(
     dist::VectorOfUnivariate,
     x::AbstractMatrix{<:Real},
 )
-    return mapvcat(eachcol(x)) do x
+    return eachcolmaphcat(x) do x
         link(dist, x)
     end
 end
@@ -271,7 +271,7 @@ function invlink(
     dist::VectorOfUnivariate,
     x::AbstractMatrix{<:Real},
 )
-    return mapvcat(eachcol(x)) do x
+    return eachcolmaphcat(x) do x
         invlink(dist, x)
     end
 end
--- a/src/compat/distributionsad.jl
+++ b/src/compat/distributionsad.jl
@@ -259,6 +259,23 @@ function link(
     end
 end
 
+function invlink(
+    dist::VectorOfUnivariate,
+    x::AbstractVector{<:Real},
+)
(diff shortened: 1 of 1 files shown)


mohamed82008:
As soon as tests pass, let's merge this.

torfjelde:
Test fails:
```julia
MethodError: no method matching logpdf_with_trans(::Product{Discrete,Bernoulli{Tracker.TrackedReal{Float64}},Array{Bernoulli{Tracker.TrackedReal{Float64}},1}}, ::Array{Array{Int64,1},1}, ::Bool)
```
But should this even be tested? Are we supporting `logpdf_with_trans(d::MultivariateDistribution, x::Vector{Vector{T}}, ::Bool)` now?

torfjelde:
So I'm quite confused.. Locally I've commented out the lines
https://github.com/TuringLang/Bijectors.jl/blob/8f53007ffd3b3eb9b43567ff17895102802846dd/test/ad/distributions.jl#L89-L92
but when I do `] test` I _still_ get the error
```julia
Multivariate discrete distributions: Error During Test at /home/tor/.julia/dev/Bijectors/test/ad/distributions.jl:521
  Got exception outside of a @test
  MethodError: no method matching logpdf_with_trans(::Product{Discrete,BetaBinomial{Tracker.TrackedReal{Float64}},Array{BetaBinomial{Tracker.TrackedReal{Float64}},1}}, ::Array{Array{Int64,1},1}, ::Bool)
```
:confused:

mohamed82008:
This is weird, I will take a look.

src/compat/distributionsad.jl
```diff
@@ -259,6 +259,23 @@ function link(
...
+)
+    return mapvcat(dist.v, x) do dist, x
+        invlink(dist, x)
+    end
+end
+function invlink(
+    dist::VectorOfUnivariate,
+    x::AbstractMatrix{<:Real},
+)
+    return mapvcat(eachcol(x)) do x
```
mohamed82008, line 274: This should be `maphcat`.

mohamed82008: Same for `link`

mohamed82008:
Fixed. Let's wait for tests to pass then merge.

PR #102, opened 2020-04-10
AD perf fixes

mohamed82008:
This PR does something similar to https://github.com/TuringLang/DistributionsAD.jl/pull/63 by not concatenating arrays of `TrackedReal`s to `TrackedArrays` all over the place. This PR also does the same flattening trick in DistributionsAD for `filldist` to enable efficient broadcasting in the `logpdf` of `filldist` returning `TrackedArray`s.

This PR also fixes the `invlink` issue in `arraydist` where `invlink` was the identity for `arraydist` of univariate distributions. In this PR, I also define a number of ReverseDiff custom adjoints using the `@grad` macro in https://github.com/JuliaDiff/ReverseDiff.jl/pull/123. Some type piracy code for AD was also removed from here and added to DistributionsAD in https://github.com/TuringLang/DistributionsAD.jl/pull/63. So this PR requires both the ReverseDiff and DistributionsAD PRs to be merged and released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -536,13 +536,13 @@ struct Logit{T<:Real} <: Bijector{0}
 end
 
 (b::Logit)(x::Real) = logit((x - b.a) / (b.b - b.a))
-(b::Logit)(x) = mapvcat(b, x)
+(b::Logit)(x) = map(b, x)
 # `orig` contains the `Bijector` which was inverted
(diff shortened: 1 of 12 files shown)


torfjelde:
A bit uncertain whether or not dropping `vcat` is the best idea atm :confused: Ref: https://github.com/TuringLang/Bijectors.jl/pull/99#issuecomment-612398170

mohamed82008:
Closing in favor of #99.

PR #104, opened 2020-05-01
Get rid of DistributionBijector

mohamed82008:
This PR cleans up Bijectors some more following #99. Tests should probably pass but there is a 2x slow down in the `lda` and `hmm_semisup` benchmarks in TuringExamples so I will keep looking into it before merging. On the plus side, `naive_bayes` is 3x faster with this branch. 

In this PR, I get rid of DistributionBijector as the "default" bijector entirely. It is now renamed to `ADBijector` and used purely for some AD tests. But I am not sure if ADBijector itself is useful or not. @torfjelde ?


torfjelde:
Sorry, _renamed_ to `ADBijector`; don't we already have one of those?

mohamed82008:
Ya the abstract `ADBijector` is now `AbstractADBijector`.

torfjelde:
But the issue with `DistributionBijector` and now `ADBijector` is that the `link` and `invlink` be using `bijector` internally from now on, no? So I don't quite understand why we want it at all.

mohamed82008:
>But the issue with DistributionBijector and now ADBijector is that the link and invlink be using bijector internally from now on, no? So I don't quite understand why we want it at all.

mohamed82008:
And yes `link`, `invlink` and `logpdf_with_trans` use `bijector` now. `ADBijector` calls `link` and `invlink` which would call `bijector`.

torfjelde:
> I agree but I left it because there were quite a few AD-related tests I wasn't comfortable removing.

Gotcha :+1: But yeah I think we should just remove it completely (though keep the `abstract type ADBijector`)

> And yes link, invlink and logpdf_with_trans use bijector now. ADBijector calls link and invlink which would call bijector.

Exactly! So it's full circle if there's no prior impl for `bijector` :)

mohamed82008:
@torfjelde what do you say we make ADBijector wrap other bijectors instead of distributions? That way if I am defining my own bijector, I don't need to define `logabsdetjac` and I can just wrap it in `ADBijector` making use of AD for that.

mohamed82008:
Also if I am defining a bijector for my own distribution, I can do `bijector(d::MyDist) = ADBijector(mybijector())` where `mybijector` only defines the the forward and inverse transformations. That way `link`, `invlink` and `logpdf_with_trans` will also work for that distribution.

torfjelde:
> @torfjelde what do you say we make ADBijector wrap other bijectors instead of distributions? That way if I am defining my own bijector, I don't need to define logabsdetjac and I can just wrap it in ADBijector making use of AD for that.

Hmm, that might be a good idea. So currently you would have to subclass `ADBijector`, but you're proposing making it a wrapper instead, i.e.
```julia
struct MyBijector{AD} <: ADBijector{AD}
# vs
struct MyBijector <: Bijector
ADBijector(MyBijector())
```
Pfft, I dunno. One thing that might be annoying dispatching on `MyBijector` because you never know wheter or not it's going to be wrapped in a `ADBijector`?

> Also if I am defining a bijector for my own distribution, I can do bijector(d::MyDist) = ADBijector(mybijector()) where mybijector only defines the the forward and inverse transformations. That way link, invlink and logpdf_with_trans will also work for that distribution.

Yeah! That should work now:)

mohamed82008:
Hmm on second thought, I will stick with the subtype. No need to change the status quo if there is no clear advantage.

mohamed82008:
Once tests pass, this will be ready.

PR #106, opened 2020-05-13
Define == for Bijector

mohamed82008:
This PR defines == for bijectors. This is needed to check that the support of a distribution hasn't changed in DynamicPPL.

### DIFF
--- a/src/bijectors/logit.jl
+++ b/src/bijectors/logit.jl
@@ -12,6 +12,8 @@ function Logit(a, b)
     Logit{0, T}(a, b)
 end
 up1(b::Logit{N, T}) where {N, T} = Logit{N + 1, T}(b.a, b.b)
+# For equality of Logit with Float64 fields to one with Duals
+Base.(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
 
 (b::Logit{0})(x::Real) = _logit(x, b.a, b.b)
 (b::Logit{0})(x) = _logit.(x, b.a, b.b)
--- a/src/bijectors/logit.jl
+++ b/src/bijectors/logit.jl
@@ -13,7 +13,7 @@ function Logit(a, b)
 end
 up1(b::Logit{N, T}) where {N, T} = Logit{N + 1, T}(b.a, b.b)
 # For equality of Logit with Float64 fields to one with Duals
-Base.(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
+Base.:(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
 
 (b::Logit{0})(x::Real) = _logit(x, b.a, b.b)
 (b::Logit{0})(x) = _logit.(x, b.a, b.b)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -70,6 +70,7 @@ up1(b::Inverse) = Inverse(up1(b.orig))
 
 inv(b::Bijector) = Inverse(b)
 inv(ib::Inverse{<:Bijector}) = ib.orig
+Base.:(==)(b1::Inverse{<:Bijector}, b2::Inverse{<:Bijector}) = b1.orig == b2.orig
 
 """
     logabsdetjac(b::Bijector, x)
--- a/src/bijectors/composed.jl
+++ b/src/bijectors/composed.jl
@@ -94,6 +94,13 @@ end
 
 isclosedform(b::Composed) = all(isclosedform, b.ts)
 up1(b::Composed) = Composed(up1.(b.ts))
+function Base.:(==)(b1::Composed{<:Any, N}, b2::Composed{<:Any, N}) where {N}
+    ts1, ts2 = b1.ts, b2.ts
+    if !(ts1 isa Tuple && ts2 isa Tuple || ts1 isa Vector && ts2 isa Vector)
+        return false
+    end
+    return all(ts1 .== ts2)
+end
 
 """
     composel(ts::Bijector...)::Composed{<:Tuple}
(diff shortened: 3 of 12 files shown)


torfjelde:
Uhhh a heads up: most of these should _just work_ because Julia, e.g.
```julia
julia> using Bijectors: Exp, Log, Logit

julia> Exp() == Exp()
true

julia> Logit(0., 1.) == Logit(1., 2.)
false

julia> Logit(0., 1.) == Logit(0., 1.)
true
```

Are there any particular ones for which it doesn't work?

torfjelde:
Probably only needed for mutable ones, e.g. `PlanarLayer`, `RadialLayer` and `Normalize` (I'm not entirely sure why we need these to be mutable though; could just make the mutations be performed elementwise using `.=` since the fields are all arrays anyways)

src/interface.jl
```diff
@@ -1,4 +1,4 @@
-import Base: inv, ∘
+import Base: inv, ∘, ==
```
devmotion, line 1: IMO one shouldn't import `==` (or `inv`...) when overloading these methods in multiple source files. In general, I think it is much easier to see which methods are internal and which overload definitions in other packages (and also in which package) if one just defines them explicitly, such as `Base.:(==)(x, y)` in this case.

torfjelde: Not opposed to that :+1:

mohamed82008:
>Uhhh a heads up: most of these should just work because Julia, e.g.
>Are there any particular ones for which it doesn't work?

Hmm let me see.

torfjelde:
Related: https://github.com/JuliaLang/julia/issues/4648

Immutable structs are compared by recursively applying `==`, but this is not the case for `mutable` (because issues: https://github.com/JuliaLang/julia/issues/4648#issuecomment-360634424)

mohamed82008:
`Permute` seems to need an `==` method too.

torfjelde:
Curious. So from the issue it seems like `==` will recursively call `isequal` and, lo and behold,
```julia
julia> b1 = Permute(2, 1 => 2, 2 => 1)
Permute{SparseArrays.SparseMatrixCSC{Float64,Int64}}(
  [2, 1]  =  1.0
  [1, 2]  =  1.0)

julia> b2 = Permute(2, 1 => 2, 2 => 1)
Permute{SparseArrays.SparseMatrixCSC{Float64,Int64}}(
  [2, 1]  =  1.0
  [1, 2]  =  1.0)

julia> b1 == b2
false

julia> b1.A == b2.A
true

julia> isequal(b1.A) == isequal(b2.A)
false
```
Not clear to me why this happens. _Might_ be an issue with `SparseArrays.jl`, because there's nothing in the docstring of `isequal` that tells me they shouldn't be equal.

mohamed82008:
I think the recursive "rule" doesn't hold for arrays in general.
```julia
julia> struct T
           a
       end

julia> T(ones(2)) == T(ones(2))
false
```

mohamed82008:
So we need `==` methods for any bijector that can have an array field.

torfjelde:
Ah, yeah seems like you're right :+1:

mohamed82008:
@torfjelde I think we need `==` for all bijectors with numeric fields for this to return true:
```julia
julia> using Bijectors, ReverseDiff

julia> Bijectors.Logit(0, 1) == Bijectors.Logit(ReverseDiff.track(0), ReverseDiff.track(1))
false
```
This is not super important as the AD number types showing up in the bijector is already a sign that the support of the distribution is changing so we might as well get `false` here. But I would still like to see `true` in the above case.

src/bijectors/logit.jl
```diff
@@ -12,6 +12,8 @@ function Logit(a, b)
     Logit{0, T}(a, b)
 end
 up1(b::Logit{N, T}) where {N, T} = Logit{N + 1, T}(b.a, b.b)
+# For equality of Logit with Float64 fields to one with Duals
+Base.(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
```
devmotion, line 16:
```suggestion
Base.:(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
```

mohamed82008: Oops, thanks!

src/bijectors/logit.jl
```diff
@@ -38,4 +38,4 @@ logabsdetjac(b::Logit{2}, x::AbstractMatrix) = sum(logit_logabsdetjac.(x, b.a, b
 logabsdetjac(b::Logit{2}, x::AbstractArray{<:AbstractMatrix}) = map(x) do x
     logabsdetjac(b, x)
 end
-logit_logabsdetjac(x, a, b) = -log((x - a) * (b - x) / (b - a))
\ No newline at end of file
+logit_logabsdetjac(x, a, b) = -log((x - a) * (b - x) / (b - a))
```
devmotion, line 41: Since this showed up in the commits, is there a reason for not writing
```suggestion
logit_logabsdetjac(x, a, b) = log(b - a) - log(x - a) - log(b - x)
```
?

mohamed82008: Hmm on second thought, I think your version has 3 log calls so it will be slower than the first one which has a single log. `*` and `/` are generally faster than `log`. For example:
```julia
julia> using BenchmarkTools

julia> f(x, a, b) = -log((x - a) * (b - x) / (b - a))

julia> g(x, a, b) = -log(x - a) - log(b - x) + log(b - a)

julia> fb(x, a, b) = f.(x, a, b)
fb (generic function with 1 method)

julia> gb(x, a, b) = g.(x, a, b)
gb (generic function with 1 method)

julia> @btime fb($(rand(1000)), 0.0, 1.0);
  8.866 μs (1 allocation: 7.94 KiB)

julia> @btime gb($(rand(1000)), 0.0, 1.0);
  18.699 μs (1 allocation: 7.94 KiB)
```
In this case, we are only multiplying 2 numbers so underflow is unlikely. Hence there is no need for the log sum trick.

PR #107, opened 2020-05-18
Fixed bug in logabsdetjac for Composed using Arrays

torfjelde:
There's a bug in `logabsdetjac` for `Composed{<:AbstractArray}` where it returns `(rv = ..., logabsdetjac = ...)`, i.e. same as `forward`. This PR fixes that.


torfjelde:
I'm just going to merge this as soon as tests pass, since this is a breaking bug.

PR #110, opened 2020-05-31
More general equality check for `Composed`

devmotion:


torfjelde:
I guess I'm a bit too late to the party, but this PR made it so that:
```julia
b1 = Composed((Exp{0}(), Log{0}())) # using tuple
b2 = Composed([Exp{0}(), Log{0}()]) # using array

b1 == b2
```
results in `true`. I mean, I'm guessing this is the reason why this PR got merged in the first place. I'm just wondering, is this something we want? I guess it's related to the reason given in #106: "This is needed to check that the support of a distribution hasn't changed in DynamicPPL.". Hmm, not sure I like this. I'm generally not a fan of these sorts of solutions, i.e. the main motivation seems is to provide functionality to a different package (DynamicPPL in this case). The correct solution is introducing constraints, as we've spoken about before (but as we've noted then, it's requires a fair bit of work). So I'm guessing we did it in this particular case because we needed that fix, but it's a slippery slope imo.

Also, it seems like several PRs have been merged with failing tests for no apparent reason (currently trying to fix stuff, and it all seems directly related to the changes made in the PRs). So, yah know, let's try _not_ to do that in the future :upside_down_face:

PR #111, opened 2020-06-15
Update DistributionsAD and Distributions compatibilities

devmotion:
This PR updates Bijectors to the latest release of DistributionsAD.

BTW I'm a bit surprised that it's necessary to reimplement the log probabilities of the (inverse) Wishart distributions here, somehow that feels wrong.


devmotion:
I assume that tests will fail since all tests fail on the master branch it seems.

torfjelde:
Uhmm, this is strange :confused:  I'll try to fix the tests and make a PR.

PR #112, opened 2020-06-15
Fixes for failing tests that have accumulated

torfjelde:


test/interface.jl
```diff
@@ -564,17 +580,17 @@ end
...
 
     sb2 = Stacked([b, b, inv(b), inv(b)])        # <= Array
     res2 = forward(sb2, [x, x, y, y])
     @test sb2(param([x, x, y, y])) isa TrackedArray
 
-    @test sb2([x, x, y, y]) == res2.rv
-    @test logabsdetjac(sb2, [x, x, y, y]) ≈ 0.0
-    @test res2.logabsdetjac ≈ 0.0
+    @test sb2([x, x, y, y]) ≈ res2.rv
+    @test isapprox(logabsdetjac(sb2, [x, x, y, y]), 0.0, atol = 1e-12)
```
devmotion, line 592: Yeah we should avoid checks of the form `x ≈ 0.0` since by default `atol = 0`, hence they will always fail if `x` is not exactly zero (there's a warning in the doc string). IIRC you could also write `@test x ≈ 0.0 atol=1e-12` if you think that's visually more appealing.

torfjelde: > Yeah we should avoid checks of the form x ≈ 0.0 since by default atol = 0

Ah, didn't know that; thanks!

> IIRC you could also write @test x ≈ 0.0 atol=1e-12 if you think that's visually more appealing.

I do! Will change those:)

devmotion:
It seems the test failures are not "real" but due to Github actions being cancelled or not started?

torfjelde:
Hmm, not entirely sure. As far as I can tell, Travis doesn't test Zygote so there might just be a bug with Zygote. Also, I have no idea why Github Actions isn't allowing me to view the logs...

devmotion:
There's more information available if you download the log archive. According to the logs therein the Zygote tests on Julia 1 and 1.3 (ubuntu-latest) failed without Julia-related error messages since "The operation was canceled". It is unclear to me why that happened - maybe the time limit was exceeded?

torfjelde:
Ah, yeah then that might be the case. Uhhm, should we try to rerun or?

devmotion:
Yeah, maybe try it again to check if the problem persists.

devmotion:
Seems to work better this time, Zygote tests on MacOS were successful. Maybe it was really just some weird non-Julia related problem with Github Actions before.

PR #114, opened 2020-06-17
Some fixes

devmotion:
Some test issues in https://github.com/TuringLang/Bijectors.jl/pull/113 were not fixed. It seems the problem is rather running the tests on Windows than on 32bit (or maybe both), and tests fail on Julia 1.0 since `eachcol` is not defined in the optional AD backends.

### DIFF
--- a/.github/workflows/ForwardDiff_Tracker.yml
+++ b/.github/workflows/ForwardDiff_Tracker.yml
@@ -19,11 +19,7 @@ jobs:
           - ubuntu-latest
           - macOS-latest
         arch:
-          - x86
           - x64
-        exclude:
-          - os: macOS-latest
-            arch: x86
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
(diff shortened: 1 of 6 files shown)


torfjelde:
Sweet! But are now just disabling testing on Windows?

devmotion:
> But are now just disabling testing on Windows?

I tried to only disable Windows (it was only added in the last PR and not tested before) but there were still memory issues in one of the 32bit tests (e.g., https://github.com/TuringLang/Bijectors.jl/runs/780512942), so I disabled Windows and 32bit for now. I'm still hoping that (at least parts of) the problem can be resolved by not using `eval`.

devmotion:
Finally all tests passed :tada: Can we merge this and make a release, so that we can fix the issues with Distributions >= 0.23.3 in Turing and DynamicPPL? Since the `c0` fixes are breaking, I assume the next release should be 0.8.0.

yebai:
> Finally all tests passed 🎉 Can we merge this and make a release, so that we can fix the issues with Distributions >= 0.23.3 in Turing and DynamicPPL? Since the c0 fixes are breaking, I assume the next release should be 0.8.0.

Sounds good!

PR #116, opened 2020-06-21
Remove use of `eval` in tests

devmotion:
This PR removes the use of `eval` in the tests, simplifies their structure, and avoids some symmetries and "nice" specially chosen values in the AD tests by using random parameters and samples.


devmotion:
With this PR Zygote tests finish in around 25 minutes instead of > 3h on master. It seems that Julia 1.0 requires some additional adjoints for Tracker and ReverseDiff though.

devmotion:
Travis tests fail due some change introduced in Tracker 0.2.7 (the previous commit passed on Tracker 0.2.6, and the last commit did not change anything that affects the Travis tests).

devmotion:
@torfjelde Any idea what change in Tracker (or some other dependency) might cause the issues on Travis?

Otherwise the tests pass and are extremely fast now :tada: I guess, the tests in DistributionsAD should be changed in the same/similar way.

devmotion:
BTW if someone enables codecov for this repo, then the codecov comments in the PRs provide some more meaningful information about coverage changes.

yebai:
>BTW if someone enables codecov for this repo, then the codecov comments in the PRs provide some more meaningful information about coverage changes.

Isn't Codecov already enabled?

devmotion:
Hmm maybe it is actually. I was confused that the report states that the coverage diff is "n/a" but maybe that's not relevant.

yebai:
> Hmm maybe it is actually. I was confused that the report states that the coverage diff is "n/a" but maybe that's not relevant.


It's indeed strange. Codecov page says there aren't sufficient commits for this repo yet. 

@torfjelde Please add a code review, or proceed with merging if everything looks good.

devmotion:
I assume the issues on Travis are the same as in https://github.com/FluxML/Tracker.jl/pull/71. It seems the latest Tracker release is declared to be compatible with Adapt 2 but was only tested with Adapt 1.

devmotion:
Tests on master fail due to the same problem.

test/ad/distributions.jl
```diff
@@ -1,586 +1,621 @@
...
+        #DistSpec(LogitNormal, (), [0.5]),
+        #DistSpec(LogitNormal, (1.0,), [0.5]),
+        #DistSpec(LogitNormal, (1.0, 2.0), [0.5]),
+
+        DistSpec(LogNormal, (), 0.5),
+        DistSpec(LogNormal, (1.0,), 0.5),
+        DistSpec(LogNormal, (1.0, 2.0), 0.5),
+        #DistSpec(LogNormal, (), [0.5]),
+        #DistSpec(LogNormal, (1.0,), [0.5]),
+        #DistSpec(LogNormal, (1.0, 2.0), [0.5]),
```
torfjelde, line 215: Is there a reason why we have all these comments? Unless I'm completely mistaken, these are always identical to the lines above, right?

test/ad/distributions.jl
```diff
@@ -1,586 +1,621 @@
...
+            # Create `arraydist` distribution
+            f_arraydist = (θ...,) -> arraydist([d.f(θ...) for _ in 1:n])
+            d_arraydist = f_arraydist(d.θ...)
+
+            for sz in ((n,), (n, 2))
+                # Matrix case doesn't work for continuous distributions for some reason
+                # now but not too important (?!)
+                if length(sz) == 2 && Distributions.value_support(typeof(d)) === Continuous
+                    continue
+                end
```
torfjelde, line 493: Do you know why? I.e. what's the error?

Project.toml
```diff
@@ -18,30 +18,28 @@ Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
...
 Requires = "0.5, 1"
-Roots = "0.8.4, 1.0"
+Roots = "0.8.4, 1"
 StatsFuns = "0.8, 0.9.3"
-Tracker = "0.2.3"
-Zygote = "0.4.10"
 julia = "1"
 
 [extras]
+Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
```
torfjelde, line 34: Where is `Adapt` used?

test/ad/distributions.jl
devmotion: I don't know why they exist (I just copied them since I thought maybe there's some reason). They are not identical (the difference seems to be that they use vector-valued samples/locations), but since we test the gradient with respect to vector-valued inputs in the end even the scalar samples end up as part of a vector. IMO the comments could just be removed. Would that be fine with you?

test/ad/distributions.jl
devmotion: No, I can check it. That's why I added "(?!)" to the comment :smile:

Project.toml
devmotion: It's a dependency of Tracker, it's never used explicitly. It's just a bugfix for the latest release (Tracker 0.2.7) which incorrectly specifies that it is compatible with Adapt 2 whereas actually it isn't (it was added but never tested since the checked-in Manifest.toml of Tracker was not updated...). There's some discussion in an open PR to the Tracker repo, I even suggested fixing the bounds in the registry (since it's just a patch release it's not possible to prevent the use of this latest release if the version is not specified explicitly, i.e., by `Tracker = "= 0.2.6"` etc.

The fix should be removed as soon as the problems in Tracker/Adapt are fixed.

test/ad/distributions.jl
devmotion: BTW I think it's pretty bad to just state arrays of distributions for which tests are broken because we won't notice if some change (in Bijectors or some other package) fixes a broken test. IMO we should explicitly test whatever behaviour is broken. I just didn't want to put too many changes in this PR and wanted to check if it actually helps to reduce testing times first.

test/ad/distributions.jl
torfjelde: Ah, sorry; my slightly sleep-deprived mind could not comprehend `[` and `]`. But yeah, that sounds good to me :+1:

Project.toml
torfjelde: Ah, aight; sounds good! :+1:

test/ad/distributions.jl
devmotion: It seems it's mainly an issue with Tracker (yields `MethodError: Cannot `convert` an object of type Array{Float64,1} to an object of type SubArray{Float64,1,Array{Float64,2},Tuple{Base.Slice{Base.OneTo{Int64}},Int64},true}` already for the first distribution), although Zygote also yields errors for some distributions such as `arraydist` with `GeneralizedExtremeValue` (but works fine for all distributions above, i.e., `Arcsine` to `Gamma`).

torfjelde: Sounds like a good idea, but how would we do this in practice when checking PRs and stuff? Should have a one test-suite which tests broken tests and is allowed to fail? Doesn't seem like that will helps us discover if tests are fixed by upstream packages since there will be too much noise.

devmotion: I guess we would have to state explicitly for which AD backends tests don't pass, and then use `@test_broken` instead of `@test` for these inside of `test_ad`. Probably would need an additional argument to `DistSpec`, e.g., a field `broken` with default value `()` that could be set to `("Zygote",)`, `("Zygote", "ReverseDiff")` etc. Would be tedious to set up initially, but should work quite reliably I assume.

torfjelde: Aaaah, pfft I haven't read the docs of `@test_broken` before; that makes complete sense. But should we maybe just get this PR merged now and then we can add this later? Unless you're _really_ keen on just getting it done immediately, in which case I won't stop you:)

devmotion: No, actually I'm not very keen on going through the lists and figuring out why these tests are broken right now :stuck_out_tongue: So I'm fine with adding this later.

torfjelde:
Merged; thanks @devmotion !:)

Issue #121, opened 2020-07-06
PlanarLayer parameter b is not Tracked


cmerkatas:
Hi, it seems that parameter b in PlanarLayer is not Tracked. I have tried to reproduce the example in documentation and I get
<pre><code>
using Bijectors
using Tracker
d = MvNormal(zeros(2), ones(2));
b = PlanarLayer(2, param);
flow = transformed(d, b)
Bijectors.TransformedDistribution{MvNormal{Float64,PDMats.PDiagMat{Float64,Array{Float64,1}},Array{Float64,1}},PlanarLayer{TrackedArray{…,Array{Float64,1}},Float64},Multivariate}(
dist: DiagNormal(
dim: 2
μ: [0.0, 0.0]
Σ: [1.0 0.0; 0.0 1.0]
)
transform: PlanarLayer{TrackedArray{…,Array{Float64,1}},Float64}([1.1142736114526957, -1.8696189914068417] (tracked), [0.6892713124675772, 0.53047763560658] (tracked), -1.0403736727795068)
)
</code></pre>

Bijectors is v0.8.1

torfjelde:
Ah, thanks for pointing this out! This was introduced not too long ago in a PR and the change sort of slipped under the radar.

For the moment, I'd manually instantiating `PlanarLayer`. Under the hood it's literally just doing `PlanarLayer(randn(2), randn(2), randn())` and it was supposed to wrap the arguments in `param`.

torfjelde:
Fix is one the way!

cmerkatas:
Thank you @torfjelde. It will be useful to fit flows I believe. I haven't managed to "track" it even manually but it was a quick test. I will try again later.

torfjelde:
Should be fixed now, i.e. you can do `PlanarLayer(2, Tracker.param)` :+1:

torfjelde:
Fixed in #122

cmerkatas:
Thank you, that fix really does Track the components. However, there is still some problem with the examples. For example

<pre><code>
using Bijectors
using Flux, Tracker

d = MvNormal(zeros(2), ones(2));
b = PlanarLayer(2, Tracker.param);
flow = transformed(d,b)
# after this point b is not tracked anymore 

@Flux.treelike Composed
@Flux.treelike Bijectors.TransformedDistribution
@Flux.treelike PlanarLayer
ps = Flux.params(flow)

[...]

PR #122, opened 2020-07-06
Improved default constructors for `RadialLayer` and `PlanarLayer`

torfjelde:
Fixes issue #121.


torfjelde:
Irrelevant to the change you mean? Yeah. So I guess we can just merge it.

But there's some weird stuff going on after the most recent update to ChainRules.jl.

Issue #124, opened 2020-08-06
Unit tests for PlanarLayer failed in Julia 1.5


yiyuezhuo:
```julia
] test
```

raise:

```
PlanarLayer: Test Failed at /home/yiyuezhuo/.julia/packages/Bijectors/66bcj/test/norm_flows.jl:30
  Expression: ≈((inv(flow))(flow(z)), z, rtol = 0.2)
   Evaluated: [2.2075135426840555 -0.18182425891670018 … 1.3994120484164432 0.9248803865772497; -1.5719067963233282 -0.1836889984451685 … 0.7022959819078269 0.27282017311337303] ≈ [1.9716204287413468 -0.15707698104964826 … 1.3606109850208474 0.4787377009574211; -1.3961992590989196 -0.20212227601885754 … 0.7311973736313007 0.6051343777576946] (rtol=0.2)
Stacktrace:
 [1] top-level scope at /home/yiyuezhuo/.julia/packages/Bijectors/66bcj/test/norm_flows.jl:30
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/yiyuezhuo/.julia/packages/Bijectors/66bcj/test/norm_flows.jl:23
PlanarLayer: Test Failed at /home/yiyuezhuo/.julia/packages/Bijectors/66bcj/test/norm_flows.jl:31
  Expression: ≈((inv(flow) ∘ flow)(z), z, rtol = 0.2)
   Evaluated: [2.2075135426840555 -0.18182425891670018 … 1.3994120484164432 0.9248803865772497; -1.5719067963233282 -0.1836889984451685 … 0.7022959819078269 0.27282017311337303] ≈ [1.9716204287413468 -0.15707698104964826 … 1.3606109850208474 0.4787377009574211; -1.3961992590989196 -0.20212227601885754 … 0.7311973736313007 0.6051343777576946] (rtol=0.2)
Stacktrace:
 [1] top-level scope at /home/yiyuezhuo/.julia/packages/Bijectors/66bcj/test/norm_flows.jl:31
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/yiyuezhuo/.julia/packages/Bijectors/66bcj/test/norm_flows.jl:23
[...]

PR #125, opened 2020-08-06
Add LKJ bijector

### LINKED ISSUE #108: Add LKJ Matrix Distribution
I'm currently trying to build a model in Turing that utilizes the [newly](https://github.com/JuliaStats/Distributions.jl/pull/1066) added LKJ distribution from Distributions.jl. I ran into issues when I attempted to sample the model:
```
ERROR: MethodError: no method matching bijector(::LKJ{Float64,Int64})
Closest candidates are:
  bijector(::KSOneSided) at /Users/joshualeond/.julia/packages/Bijectors/bHaf6/src/transformed_distribution.jl:58
  bijector(::Product{Discrete,T,V} where V<:AbstractArray{T,1} where T<:Distribution{Univariate,Discrete}) at /Users/joshualeond/.julia/packages/Bijectors/bHaf6/src/transformed_distribution.jl:39
  bijector(::Product{Continuous,T,Tdists} where Tdists<:(FillArrays.Fill{T,1,Axes} where Axes) where T<:Distribution{Univariate,Continuous}) at /Users/joshualeond/.julia/packages/Bijectors/bHaf6/src/compat/distributionsad.jl:16
```

It appears that the LKJ distribution maybe needs to be added here in Bijectors: https://github.com/TuringLang/Bijectors.jl/blob/1f3b581afe04f690bd93fba9edd88735cc1fc140/src/Bijectors.jl#L202 

However, after I simply added LKJ to this line I hit the following error:

```
ERROR: MethodError: no method matching getlogp(::LKJ{Float64,Int64}, ::Cholesky{Float64,Array{Float64,2}}, ::Array{Float64,2})
Closest candidates are:
  getlogp(::MatrixBeta, ::Any, ::Any) at C:\Users\dunjos0\.julia\dev\Bijectors\src\Bijectors.jl:231
  getlogp(::Wishart, ::Any, ::Any) at C:\Users\dunjos0\.julia\dev\Bijectors\src\Bijectors.jl:236
  getlogp(::InverseWishart, ::Any, ::Any) at C:\Users\dunjos0\.julia\dev\Bijectors\src\Bijectors.jl:239
```
So it looks like there needs to be a new `getlogp` method in Bijectors for the LKJ. I [see](https://github.com/JuliaStats/Distributions.jl/blob/2d7f203deda5070cad8ef7d69d43cc054bb60457/src/matrix/lkj.jl#L110) that these `getlogp` methods follow closely to the `logkernel` methods in Distributions.jl but am unsure what needs to be adjusted to make sure it works nicely with Turing.

yiyuezhuo:
Fix #108 

Implement LKJ bijector (transform), following Stan's [spec](https://mc-stan.org/docs/2_24/reference-manual/correlation-matrix-transform-section.html).

Since #124, I guess the CI will not pass. Though all PR related tests should pass.

Implement #108 example:

Setup:

```julia
using Zygote
using ForwardDiff
using ReverseDiff
using Tracker

using Distributions
using DistributionsAD

using Turing

using Bijectors
using LinearAlgebra
using PDMats
using Serialization

sigma = [1,2,3]
Omega = [1 0.3 0.2;
        0.3 1 0.1;
        0.2 0.1 1]

Sigma = Diagonal(sigma) * Omega * Diagonal(sigma)
N = 100
J = 3
# y = rand(MvNormal(zeros(J), Sigma), N)'
y = deserialize("y.ser") # Same value is compared with CmdStan
```

Model definition:

```julia


@model correlation(J, N, y, Zero0) = begin
    sigma ~ filldist(truncated(Cauchy(0., 5.), 0., Inf), J) # prior on the standard deviations
    Omega ~ LKJ(J, 1) # LKJ prior on the correlation matrix

    _Sigma = Symmetric(Diagonal(sigma) * Omega * Diagonal(sigma))
    # Sigma = PDMat(_Sigma) # PDMat requires extra AD support
    Sigma = _Sigma

    for i in 1:N
        y[i,:] ~ MvNormal(Zero0, Sigma) # sampling distribution of the observations
    end
    return Sigma
end

```

Since `truncated(Cauchy(0., 5.), 0., Inf)` is a weakly informative prior, which sometimes gives a too large value which breaks sampling, specifying initial value is useful.

```julia
model = correlation(J, N, y, zeros(J))
varinfo = Turing.VarInfo(model)
model(varinfo, Turing.SampleFromPrior(), Turing.PriorContext((
    sigma=ones(3),
    Omega=Float64.(collect(I(3))) # unit matrix
)))
init_theta = varinfo[Turing.SampleFromPrior()]
```

```julia
setadbackend(:forwarddiff)
sample(correlation(J, N, y, zeros(J)), HMC(0.01, 5), 100, init_theta = init_theta)

#=
┌ Info: Using passed-in initial variable values
│   init_theta = [1.0, 1.0, 1.0, 1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0]
└ @ Turing.Inference /home/yiyuezhuo/.julia/packages/Turing/p8FFd/src/inference/Inference.jl:282
Sampling: 100%|█████████████████████████████████████████| Time: 0:00:00
Chains MCMC chain (100×21×1 Array{Float64,3}):

Iterations        = 1:100
Thinning interval = 1
Chains            = 1
Samples per chain = 100
parameters        = Omega[1,1], Omega[1,2], Omega[1,3], Omega[2,1], Omega[2,2], Omega[2,3], Omega[3,1], Omega[3,2], Omega[3,3], sigma[1], sigma[2], sigma[3]
internals         = acceptance_rate, hamiltonian_energy, hamiltonian_energy_error, is_accept, log_density, lp, n_steps, nom_step_size, step_size

Summary Statistics
  parameters      mean       std   naive_se      mcse       ess      rhat  
      Symbol   Float64   Float64    Float64   Missing   Float64   Float64  
                                                                           
  Omega[1,1]    1.0000    0.0000     0.0000   missing       NaN       NaN  
  Omega[1,2]    0.2779    0.0964     0.0096   missing   10.4459    0.9932  
  Omega[1,3]    0.0572    0.0809     0.0081   missing   14.2873    0.9924  
  Omega[2,1]    0.2779    0.0964     0.0096   missing   10.4459    0.9932  
  Omega[2,2]    1.0000    0.0000     0.0000   missing   87.0968    0.9899  
  Omega[2,3]    0.1385    0.0858     0.0086   missing   13.3596    1.0244  
  Omega[3,1]    0.0572    0.0809     0.0081   missing   14.2873    0.9924  
  Omega[3,2]    0.1385    0.0858     0.0086   missing   13.3596    1.0244  
  Omega[3,3]    1.0000    0.0000     0.0000   missing   75.9334    0.9899  
    sigma[1]    1.0294    0.0934     0.0093   missing   10.0810    0.9911  
    sigma[2]    2.1043    0.1837     0.0184   missing   14.3855    1.0325  
    sigma[3]    2.8150    0.2709     0.0271   missing    9.9845    0.9955  

Quantiles
  parameters      2.5%     25.0%     50.0%     75.0%     97.5%  
      Symbol   Float64   Float64   Float64   Float64   Float64  
                                                                
  Omega[1,1]    1.0000    1.0000    1.0000    1.0000    1.0000  
  Omega[1,2]    0.0563    0.2390    0.2869    0.3380    0.4501  
  Omega[1,3]   -0.0743    0.0020    0.0507    0.1134    0.2157  
  Omega[2,1]    0.0563    0.2390    0.2869    0.3380    0.4501  
  Omega[2,2]    1.0000    1.0000    1.0000    1.0000    1.0000  
  Omega[2,3]   -0.0018    0.0721    0.1273    0.2092    0.2812  
  Omega[3,1]   -0.0743    0.0020    0.0507    0.1134    0.2157  
  Omega[3,2]   -0.0018    0.0721    0.1273    0.2092    0.2812  
  Omega[3,3]    1.0000    1.0000    1.0000    1.0000    1.0000  
    sigma[1]    0.8667    0.9653    1.0302    1.0917    1.1970  
    sigma[2]    1.7321    1.9863    2.0986    2.2369    2.4306  
    sigma[3]    2.3502    2.6356    2.8207    2.9906    3.2872  
=#
```

While tests passed for `ForwardDiff, ReverseDiff, Zygote, Tracker`. Above model will not work on `Tracker` backend.

NUTS sampler is not that robust for this model, because errors such as non-positive-definite will break whole sampling rather than just reject current step like Stan, so following adaption is proposed, according to https://github.com/TuringLang/Turing.jl/issues/702 .

```julia
@model correlation(J, N, y, Zero) = begin
    
    if :Omega in fieldnames(typeof(_varinfo.metadata))
        if :vals in fieldnames(typeof(_varinfo.metadata.Omega))
            if :value in fieldnames(typeof(_varinfo.metadata.Omega.vals[1]))
                if det(Bijectors.inv_link_lkj(reshape(map(x->x.value, _varinfo.metadata.Omega.vals), 3, 3)))<=0
                    println("reject 1!")
                    Turing.acclogp!(_varinfo, -Inf)
                    return
                end
            end
        end
    end
    
    sigma ~ filldist(truncated(Cauchy(0., 5.), 0., Inf), J) # prior on the standard deviations
    Omega ~ LKJ(J, 1) # LKJ prior on the correlation matrix

    _Sigma = Symmetric(Diagonal(sigma) * Omega * Diagonal(sigma))
    if !isposdef(_Sigma)
        println("reject 2!")
        Turing.acclogp!(_varinfo, -Inf)
        return
    end
    Sigma = PDMat(_Sigma)

    for i in 1:N
        y[i,:] ~ MvNormal(Zero, Sigma) # sampling distribution of the observations
    end
    return Sigma
end
```

Above code will only work for `ForwardDiff, ReverseDiff` backends.


src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
+# See stan doc for parametrization method:
+# https://mc-stan.org/docs/2_23/reference-manual/correlation-matrix-transform-section.html
+# (7/30/2020) their "manageable expression" is wrong...
+
+function upper1(AT, A)
+    AU = zero(AT)
+    for i=1:size(A,1), j=(i+1):size(A,2)
+        AU[i,j] = A[i,j]
+    end
+    AU
```
torfjelde, line 10:
```suggestion
    return AU
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+
+logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real}) = log_abs_det_jac_lkj(y)
+function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
+    
+    if !LinearAlgebra.isposdef(X)
+        println("!isposdef(X)")
+        return NaN # prevent Cholesky decomposition to break inference
+    end
+    
+    -log_abs_det_jac_lkj(b(X)) # It may be more efficient if we can use un-contraint value to prevent call of b
```
torfjelde, line 30:
```suggestion
    return -log_abs_det_jac_lkj(b(X)) # It may be more efficient if we can use un-contraint value to prevent call of b
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+(b::CorrBijector)(X::AbstractMatrix{<:Real}) = link_lkj(X)
+(b::CorrBijector)(X::AbstractArray{<:AbstractMatrix{<:Real}}) = map(b, X)
+
+(ib::Inverse{<:CorrBijector})(Y::AbstractMatrix{<:Real}) = inv_link_lkj(Y)
+(ib::Inverse{<:CorrBijector})(Y::AbstractArray{<:AbstractMatrix{<:Real}}) = map(ib, Y)
+
+
+logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real}) = log_abs_det_jac_lkj(y)
+function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
+
```
torfjelde, line 24:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    
+    -log_abs_det_jac_lkj(b(X)) # It may be more efficient if we can use un-contraint value to prevent call of b
+end
+logabsdetjac(b::CorrBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = mapvcat(X) do x
+    logabsdetjac(b, x)
+end
+
+
+function log_abs_det_jac_lkj(y)
+    # println("log_abs_det_jac_lkj $(typeof(y) == Matrix{Float64})")
```
torfjelde, line 38:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    # it's defined on inverse mapping
```
torfjelde, line 39: Hmm, not entirely sure what this comment refers to :confused:

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    right = 0
+    for i = 1:(K-1), j = (i+1):K
+        right += log(cosh(y[i, j])^2)
+    end
+    
+    return  (0.5 * left - right)
+end
+
+function inv_link_w_lkj(y)
+    # println("inv_link_w_lkj $(typeof(y) == Matrix{Float64})")
```
torfjelde, line 57:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    return w
+end
+
+function inv_link_lkj(y)
+    w = inv_link_w_lkj(y)
+    return w' * w
+end
+
+function link_w_lkj(w)
+    # println("link_w_lkj $(typeof(w) == Matrix{Float64})")
```
torfjelde, line 92:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    K = size(w, 1)
+
+    # z = zero(w) # `zero` isn't compatible with ReverseDiff
```
torfjelde, line 95:
```suggestion
    # `zero` isn't compatible with ReverseDiff, so we use `similar`
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    for i=1:K, j=1:K
+        z[i,j] = 0
+    end
+    
+    for j=2:K
+        z[1, j] = w[1, j]
+    end
+
+    #=
+    # This implementation will not works when w[i-1, j] = 0.
```
torfjelde, line 106:
```suggestion
    # This implementation will not work when w[i-1, j] = 0.
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    # Though it is a zero measure set, unit matrix initialization will not works.
```
torfjelde, line 107:
```suggestion
    # Though it is a zero measure set, unit matrix initialization will not work.
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+
+    for i=2:K, j=(i+1):K
+        z[i, j] = w[i, j] / w[i-1, j] * z[i-1, j] / sqrt(1 - z[i-1, j]^2)
+    end
```
torfjelde, line 111: Are you saying that this commented out code won't work with a unit-matrix, but the below impl will?

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+        end
+        z[i,j] = p
+    end
+    
+    y = atanh.(z)
+    return y
+end
+
+function link_lkj(x)
+    # println("link_lkj $(typeof(x) == Matrix{Float64})")
```
torfjelde, line 126:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    w = cholesky(x).U
+    # w = collect(cholesky(x).U)
+    # w = convert(typeof(x), cholesky(x).U) # ? test requires it, such quirk
+    # w = upper(parent(cholesky(x).U))
+    # return link_w_lkj(w)
```
torfjelde, line 131: Hmm, not entirely sure what you mean here :confused: Could you explain a bit?

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    r = link_w_lkj(w) 
+    # return r - lower(parent(r)) # test requires it, such quirk
+    upper1(x, r)
```
torfjelde, line 134:
```suggestion
    return upper1(x, r)
```

src/compat/reversediff.jl
```diff
@@ -180,4 +180,94 @@ lower(A::TrackedMatrix) = track(lower, A)
     return lower(Ad), Δ -> (lower(Δ),)
 end
 
+#=
```
torfjelde, line 183:
```suggestion
```

src/compat/reversediff.jl
```diff
@@ -180,4 +180,94 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+        end
+        
+        for j=2:K
+            Δw[1, j] += Δz[1, j]
+        end
+        
+        return (Δw,)
+    end
+end
+=#
```
torfjelde, line 271:
```suggestion
```

src/compat/tracker.jl
```diff
@@ -440,3 +442,148 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+@grad function upper1(AT_tracked, A_tracked)
+    AT = data(AT_tracked)
+    A = data(A_tracked)
+    return upper1(AT, A), Δ -> (nothing, upper1(AT, Δ))
+end
+
+# Workaround for Tracker ambiguous bug. See: https://github.com/FluxML/Tracker.jl/issues/74
+# (*)(X::Diagonal, Y::TrackedArray{T,2,A} where A where T) = collect(X) * Y
+x::Diagonal * y::TrackedMatrix  = track(*, x, y)
+x::TrackedMatrix * y::Diagonal  = track(*, x, y)
```
torfjelde, line 576: I'm not sure this is the best place to put it. Might be better suited for DistributionsAD.jl.
But I'm fine with leaving here for now, and we can just create an issue and link it to the Tracker.jl-issue. Once the Tracker.jl issue is resolved, we remove it and bump the version.

src/compat/tracker.jl
```diff
@@ -440,3 +442,148 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+
+function LinearAlgebra.isposdef(w_tracked::TrackedMatrix)
+    # w = data(w_tracked)
+    w = w_tracked.data
+    LinearAlgebra.isposdef(w)
```
torfjelde, line 581:
```suggestion
    return LinearAlgebra.isposdef(w)
```

src/compat/tracker.jl
```diff
@@ -440,3 +442,148 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+    return upper1(AT, A), Δ -> (nothing, upper1(AT, Δ))
+end
+
+# Workaround for Tracker ambiguous bug. See: https://github.com/FluxML/Tracker.jl/issues/74
+# (*)(X::Diagonal, Y::TrackedArray{T,2,A} where A where T) = collect(X) * Y
+x::Diagonal * y::TrackedMatrix  = track(*, x, y)
+x::TrackedMatrix * y::Diagonal  = track(*, x, y)
+
+function LinearAlgebra.isposdef(w_tracked::TrackedMatrix)
+    # w = data(w_tracked)
```
torfjelde, line 579: Is there a reason why we use `.data` instead of `data(...)`?

src/compat/tracker.jl
```diff
@@ -440,3 +442,148 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+    w = w_tracked.data
+    LinearAlgebra.isposdef(w)
+end
+
+function LinearAlgebra.isposdef(w_tracked::Symmetric{<:Any, <:TrackedMatrix})
+    # w = data(w_tracked)
+    w = w_tracked.data
+    LinearAlgebra.isposdef(w)
```
torfjelde, line 587:
```suggestion
    return LinearAlgebra.isposdef(w)
```

src/compat/zygote.jl
```diff
@@ -193,3 +193,126 @@ end
...
+            Δz[i-1, j] += Δw1[i, j] * w1[i-1, j] * 0.5 / tz * (-2 * z[i-1, j])
+        end
+
+        Δy = Δz .* (1 ./ cosh.(y).^2)
+        return (Δy,)
+    end
+end
+
+@adjoint function link_w_lkj(w)
+    # println("link_w_lkj $(typeof(w) == Matrix{Float64})")
```
torfjelde, line 248:
```suggestion
```

src/compat/zygote.jl
```diff
@@ -193,3 +193,126 @@ end
...
+    
+    for j=2:K
+        z[1, j] = w[1, j]
+    end
+
+    #=
+    for i=2:K, j=(i+1):K
+        z[i, j] = w[i, j] / w[i-1, j] * z[i-1, j] / sqrt(1 - z[i-1, j]^2)
+    end
+    =#
```
torfjelde, line 260: I guess this is the same as before? The code below is an "alternative" impl of this commented out section?

src/compat/zygote.jl
```diff
@@ -193,3 +193,126 @@ end
...
+end
+
+@adjoint function upper1(AT, A)
+    return upper1(AT, A), Δ -> (nothing, upper1(AT, Δ))
+end
+
+Zygote.@adjoint function LinearAlgebra.isposdef(x)
+    LinearAlgebra.isposdef(x), _-> ()
```
torfjelde, line 317:
```suggestion
    return LinearAlgebra.isposdef(x), _-> ()
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+(ib::Inverse{<:CorrBijector})(Y::AbstractMatrix{<:Real}) = inv_link_lkj(Y)
+(ib::Inverse{<:CorrBijector})(Y::AbstractArray{<:AbstractMatrix{<:Real}}) = map(ib, Y)
+
+
+logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real}) = log_abs_det_jac_lkj(y)
+function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
+    
+    if !LinearAlgebra.isposdef(X)
+        println("!isposdef(X)")
+        return NaN # prevent Cholesky decomposition to break inference
```
devmotion, line 27: This will create a type instability. In general, we shouldn't check that the matrix is positive definite. IMO users should ensure positive definiteness if required.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
+# See stan doc for parametrization method:
+# https://mc-stan.org/docs/2_23/reference-manual/correlation-matrix-transform-section.html
+# (7/30/2020) their "manageable expression" is wrong...
+
+function upper1(AT, A)
```
devmotion, line 5: Can we make use of `UpperTriangular` (which uses a view) instead of copying the matrix?

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+end
+
+
+function log_abs_det_jac_lkj(y)
+    # println("log_abs_det_jac_lkj $(typeof(y) == Matrix{Float64})")
+    # it's defined on inverse mapping
+    K = size(y, 1)
+    
+    z = tanh.(y)
+    left = 0
```
devmotion, line 43: This creates a type instability.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    for i = 1:(K-1), j = (i+1):K
+        left += (K-i-1) * log(1 - z[i, j]^2)
+    end
+    
+    right = 0
```
devmotion, line 48: This creates a type instability.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,135 @@
...
+    for i = 1:(K-1), j = (i+1):K
+        right += log(cosh(y[i, j])^2)
+    end
+    
+    return  (0.5 * left - right)
```
devmotion, line 53: We should avoid `Float64` literals since it will promote everything:
```suggestion
    return left / 2 - right
```

src/compat/tracker.jl
```diff
@@ -12,6 +12,8 @@ using .Tracker: Tracker,
 using Compat: eachcol
 using LinearAlgebra
 
+import Base.*
+
```
devmotion, line 16: IMO readability and clarity is increased by implementing the overloaded functions as `function Base.:*(...)` and removing the import:
```suggestion
```

src/compat/tracker.jl
```diff
@@ -440,3 +442,148 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+end
+
+upper1(AT::TrackedMatrix, A::TrackedMatrix) = track(upper1, AT, A)
+@grad function upper1(AT_tracked, A_tracked)
+    AT = data(AT_tracked)
+    A = data(A_tracked)
+    return upper1(AT, A), Δ -> (nothing, upper1(AT, Δ))
+end
+
+# Workaround for Tracker ambiguous bug. See: https://github.com/FluxML/Tracker.jl/issues/74
```
devmotion, line 573: This is type piracy. If possible, the definitions should be added upstream.

src/compat/zygote.jl
```diff
@@ -193,3 +193,126 @@ end
...
+        return (Δw,)
+    end
+    =#
+end
+
+@adjoint function upper1(AT, A)
+    return upper1(AT, A), Δ -> (nothing, upper1(AT, Δ))
+end
+
+Zygote.@adjoint function LinearAlgebra.isposdef(x)
```
devmotion, line 316: This should also be fixed upstream if needed (I still think probably we don't have to call `isposdef` explicitly - and there exist adjoints for `cholesky`, so maybe we can just remove it).

src/bijectors/corr.jl
yiyuezhuo: This corresponds to section "Absolute Jacobian Determinant of the Correlation Matrix *Inverse Transform*", I suspect that the name `log_abs_det_jac_lkj` may wrongly imply that it's defined with "forward transform".

src/compat/tracker.jl
yiyuezhuo: Because `w_tracked` is a `TrackedMatrix`, which has the attribute `data` but doesn't have a `data` method dispatch. `w = data(w_tracked)` was used mistakely since I guessed it's in the `@grad` block, but not like so.

src/bijectors/corr.jl
yiyuezhuo: Yes.

src/bijectors/corr.jl
yiyuezhuo: I mean this test requirement https://github.com/TuringLang/Bijectors.jl/blob/master/test/transform.jl#L66 . I found no trouble without this property. On the other hand, I try many alternative implementations to satisfy this test, as shown in commented out code.

src/compat/zygote.jl
yiyuezhuo: Yes.

src/bijectors/corr.jl
yiyuezhuo: The Stan implementation for that occasionally broken transition is to reject this move and continue sampling. As I run following Stan code, written by @joshualeond
```stan
data {
  int<lower=1> N; // number of observations
  int<lower=1> J; // dimension of observations
  vector[J] y[N]; // observations
  vector[J] Zero; // a vector of Zeros (fixed means of observations)
}
parameters {
  corr_matrix[J] Omega; 
  vector<lower=0>[J] sigma; 
}
transformed parameters {
  cov_matrix[J] Sigma; 
  Sigma <- quad_form_diag(Omega, sigma); 
}
model {
  y ~ multi_normal(Zero,Sigma); // sampling distribution of the observations
  sigma ~ cauchy(0, 5); // prior on the standard deviations
  Omega ~ lkj_corr(1); // LKJ prior on the correlation matrix 
}
``` 

Few "error" info can be seen, like:
```
Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:

Informational Message: The current Metropolis proposal is about to be rejected because of the following issue:
Exception: correlation_model_namespace::log_prob: Sigma is not symmetric. Sigma[1,2] = inf, but Sigma[2,1] = inf (in '/home/yiyuezhuo/notebooks5/tmp/correlation.stan', line 12, column 2 to column 22)Exception: correlation_model_namespace::log_prob: Sigma is not symmetric. Sigma[1,2] = -inf, but Sigma[2,1] = -inf (in '/home/yiyuezhuo/notebooks5/tmp/correlation.stan', line 12, column 2 to column 22)

Exception: correlation_model_namespace::log_prob: Sigma is not symmetric. Sigma[1,2] = -inf, but Sigma[2,1] = -inf (in '/home/yiyuezhuo/notebooks5/tmp/correlation.stan', line 12, column 2 to column 22)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
Exception: correlation_model_namespace::log_prob: Sigma is not symmetric. Sigma[1,2] = inf, but Sigma[2,1] = inf (in '/home/yiyuezhuo/notebooks5/tmp/correlation.stan', line 12, column 2 to column 22)
If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.

If this warning occurs sporadically, such as for highly constrained variable types like covariance matrices, then the sampler is fine,
but if this warning occurs often then your model may be either severely ill-conditioned or misspecified.
```

But the inference still runs and gets reasonable a result (I guess). But for Turing's [current implementation](https://github.com/TuringLang/Turing.jl/issues/702), such "error" will break inference entirely, so I call `isposdef` before `cholesky` to prevent entire failure.

src/bijectors/corr.jl
yiyuezhuo: `PDBijector` use a similar copy based implementation https://github.com/TuringLang/Bijectors.jl/blob/master/src/bijectors/pd.jl#L16 https://github.com/JuliaLang/julia/blob/master/stdlib/LinearAlgebra/src/triangular.jl#L164. Besides, I think I can just remove this function if https://github.com/TuringLang/Bijectors.jl/blob/master/test/transform.jl#L67 is not required.

src/compat/tracker.jl
yiyuezhuo: Adding to `Tracker` or `DistributionsAD` are exactly better. But `Tracker` is not actively maintained, so a PR for this and `Symmetric` to `DistributionsAD`?

yiyuezhuo:
Something seems broken 🤔, I will push the remaining modification and add a IJulia gist to show correctness later.

src/bijectors/corr.jl
devmotion: I think it's better to not perform these checks if possible since (depending on the employed matrix type) they perform a Cholesky decomposition to check if the matrix is positive definite.

IMO one could also thinl about a keword argument, similar to the one to `cholesky`, that allows to mute these errors. Additionally, since these errors are caused by numerical issues (?), a user could use algebraic transformations of `X` or e.g. PositiveFactorizarions.jl to avoid them. Also Bijectors is not written solely for Turing, so one should not only focus on its integration in Turing when designing this bijector.

src/compat/tracker.jl
devmotion: Tracker is still maintained (a new version was released 4 weeks ago), it's just not the default AD backend in Flux anymore and hence not heavily maintained anymore.

src/bijectors/corr.jl
yiyuezhuo: I see. If not considering Turing or HMC sampling integration, it is obvious that it should be removed. But I found that will lead to a very ugly hack to simulate Stan's reject behavior, which requires some strange value given by proposal at least will not halt the whole procedure. If so, some error should be mapped to NaN/Inf (ret) value instead of a vanilla fatal error.

torfjelde: Good catch, and I agree. This is something that ought to be either handled by the user (by calling `isposdef` and making the logpdf `-Inf` to reject the sample) or inside Turing somehow, but not in Bijectors.

src/compat/zygote.jl
yiyuezhuo: Removed `isposdef`.

src/compat/tracker.jl
yiyuezhuo: Removed `type piracy` code.

src/compat/tracker.jl
yiyuezhuo: Removed type piracy code.

src/bijectors/corr.jl
yiyuezhuo: Replace it with `right = eltype(y)(0)`, is it correct? 😕

src/bijectors/corr.jl
yiyuezhuo: Replace it with `left = eltype(y)(0)`, is it correct? 😕

src/bijectors/corr.jl
yiyuezhuo: Ok, I removed related code.

yiyuezhuo:
I have no idea why CI failed at something like: `ERROR: LoadError: UndefVarError: TuringUniform not defined`😕. Anyway here is a gist to show related tests, I use similar files to quick debug modification since unit test takes too long and lacks configurability.

https://gist.github.com/yiyuezhuo/27f0b9adb26f693b804a1d6e62fb997d

src/compat/tracker.jl
devmotion: `Tracker.data(w_tracked)` should work fine.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
+    
+    return -log_abs_det_jac_lkj(b(X)) # It may be more efficient if we can use un-contraint value to prevent call of b
+end
+logabsdetjac(b::CorrBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = mapvcat(X) do x
+    logabsdetjac(b, x)
+end
+
+
+function log_abs_det_jac_lkj(y)
```
devmotion, line 24: Why do we need this separate function? It seems this could just be the implementation of
```suggestion
function logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    # it's defined on inverse mapping
+    K = size(y, 1)
+    
+    z = tanh.(y)
```
devmotion, line 28: I guess this should just be computed inside of the loop below, the allocations here seem unnecessary.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    left = eltype(y)(0)
```
devmotion, line 29: That might be incorrect, it might be better to just compute the initial summand. In any case, usually one would use `zero(eltype(y))` instead of `eltype(y)(0)`.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    for i = 1:(K-1), j = (i+1):K
+        left += (K-i-1) * log(1 - z[i, j]^2)
+    end
+    
+    right = eltype(y)(0)
```
devmotion, line 34: Same here.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+end
+
+
+function log_abs_det_jac_lkj(y)
+    # it's defined on inverse mapping
+    K = size(y, 1)
+    
+    z = tanh.(y)
+    left = eltype(y)(0)
+    for i = 1:(K-1), j = (i+1):K
```
devmotion, line 30: Julia uses column-major order, and hence it is more efficient to loop over rows in each column than to loop over columns for each row. I.e., the loop over index `i` should be the inner loop, and the one over `j` the outer loop.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+        left += (K-i-1) * log(1 - z[i, j]^2)
+    end
+    
+    right = eltype(y)(0)
+    for i = 1:(K-1), j = (i+1):K
```
devmotion, line 35: Same here.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+        for j in i:K
+            w[i, j] = w[i-1, j] * sqrt(1 - z[i-1, j]^2)
+        end
+    end
+
+    for i in 1:K
+        for j in (i+1):K
+            w[i, j] = w[i, j] * z[i, j]
+        end
+    end
```
devmotion, line 66: It would be good to avoid this loop and the allocation of `z` if possible (haven't thought through if it is possible).

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    w[1,1] = 1
+    for j in 1:K
+        w[1, j] = 1
+    end
+
+    for i in 2:K
+        for j in 1:(i-1)
+            w[i, j] = 0
+        end
+        for j in i:K
```
devmotion, line 57: Again, the loops should be changed.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+
+    #=
+    # This implementation will not work when w[i-1, j] = 0.
+    # Though it is a zero measure set, unit matrix initialization will not work.
+
+    for i=2:K, j=(i+1):K
+        z[i, j] = w[i, j] / w[i-1, j] * z[i-1, j] / sqrt(1 - z[i-1, j]^2)
+    end
+    =#
```
torfjelde, line 92:
```suggestion
    For `(i, j)` in the loop below, we define

        z₍ᵢ₋₁, ⱼ₎ = w₍ᵢ₋₁,ⱼ₎ * ∏ₖ₌₁ⁱ⁻² (1 / √(1 - z₍ₖ,ⱼ₎²))

    and so

        z₍ᵢ,ⱼ₎ = w₍ᵢ,ⱼ₎ * ∏ₖ₌₁ⁱ⁻¹ (1 / √(1 - z₍ₖ,ⱼ₎²))
               = (w₍ᵢ,ⱼ₎ * / √(1 - z₍ᵢ₋₁,ⱼ₎²)) * (∏ₖ₌₁ⁱ⁻² 1 / √(1 - z₍ₖ,ⱼ₎²))
               = (w₍ᵢ,ⱼ₎ * / √(1 - z₍ᵢ₋₁,ⱼ₎²)) * (w₍ᵢ₋₁,ⱼ₎ * ∏ₖ₌₁ⁱ⁻² 1 / √(1 - z₍ₖ,ⱼ₎²)) / w₍ᵢ₋₁,ⱼ₎
               = (w₍ᵢ,ⱼ₎ * / √(1 - z₍ᵢ₋₁,ⱼ₎²)) * (z₍ᵢ₋₁,ⱼ₎ / w₍ᵢ₋₁,ⱼ₎)
               = (w₍ᵢ,ⱼ₎ / w₍ᵢ₋₁,ⱼ₎) * (z₍ᵢ₋₁,ⱼ₎ / √(1 - z₍ᵢ₋₁,ⱼ₎²))

    which is the above implementation.
    =#
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    
+    right = eltype(y)(0)
+    for i = 1:(K-1), j = (i+1):K
+        right += log(cosh(y[i, j])^2)
+    end
+    
+    return left / 2 - right
+end
+
+function inv_link_w_lkj(y)
```
devmotion, line 42: I guess this should just be
```suggestion
function (ib::Inverse{<:CorrBijector})(y::AbstractMatrix{<:Real})
```
and return `w' * w` instead. All these different functions with (IMO) not really descriptive names should be avoided if possible.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    end
+    
+    return w
+end
+
+function inv_link_lkj(y)
+    w = inv_link_w_lkj(y)
+    return w' * w
+end
+
```
devmotion, line 75: I'd suggest to remove this (see above):
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+        z[i, j] = w[i, j] / w[i-1, j] * z[i-1, j] / sqrt(1 - z[i-1, j]^2)
+    end
+    =#
+    for i=2:K, j=(i+1):K
+        p = w[i, j]
+        for ip in 1:(i-1)
+            p *= 1 / sqrt(1-z[ip, j]^2)
+        end
+        z[i,j] = p
+    end
```
devmotion, line 99: Again the loops should be switched.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    # Though it is a zero measure set, unit matrix initialization will not work.
+
+    for i=2:K, j=(i+1):K
+        z[i, j] = w[i, j] / w[i-1, j] * z[i-1, j] / sqrt(1 - z[i-1, j]^2)
+    end
+    =#
+    for i=2:K, j=(i+1):K
+        p = w[i, j]
+        for ip in 1:(i-1)
+            p *= 1 / sqrt(1-z[ip, j]^2)
```
devmotion, line 96:
```suggestion
            p /= sqrt(1-z[ip, j]^2)
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    
+    return w
+end
+
+function inv_link_lkj(y)
+    w = inv_link_w_lkj(y)
+    return w' * w
+end
+
+function link_w_lkj(w)
```
devmotion, line 76: Probably this should just be
```suggestion
function  (b::CorrBijector)(x::AbstractMatrix{<:Real}) ```
with the additional code from `link_lkj` below.

File: src/bijectors/corr.jl [hunk 46]
```diff
@@ -0,0 +1,109 @@
...
+    
+    y = atanh.(z)
+    return y
+end
+
+function link_lkj(x)
+    w = cholesky(x).U + zero(x) # convert to dense matrix
+    r = link_w_lkj(w) 
+    return r
+end
```
devmotion, line 109:
```suggestion
```

devmotion:
This test error is due to an ancient version of DistributionsAD (version 0.1.1).

src/bijectors/corr.jl
torfjelde: Uncertain whether or not to add this tbh. In hindsight it's quite clear what's going on, I just didn't have the precedence of the operators `*` and `\` in mind and so it was unclear to me what exactly the above implementation was. Had to do the derivation before realizing.

src/bijectors/corr.jl
torfjelde: Gotcha :+1:

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+    for j=2:K
+        z[1, j] = w[1, j]
+    end
+
+    #=
+    # This implementation will not work when w[i-1, j] = 0.
+    # Though it is a zero measure set, unit matrix initialization will not work.
+
+    for i=2:K, j=(i+1):K
+        z[i, j] = w[i, j] / w[i-1, j] * z[i-1, j] / sqrt(1 - z[i-1, j]^2)
```
torfjelde, line 90:
```suggestion
        z[i, j] = (w[i, j] / w[i-1, j]) * (z[i-1, j] / sqrt(1 - z[i-1, j]^2))
```

src/bijectors/corr.jl
torfjelde: Should we also make all these loops `@inbounds`? Possibly together with adding a `@assert size(y, 1) == size(y, 2)` statement above or something?

src/bijectors/corr.jl
torfjelde: I'm guessing the reason why @yiyuezhuo has done this is to be able to define custom adjoints. Not possible to do so if we use `(b::Bijector)(x)`. 

So I think leaving the `(inv_)link_w_lkj` is fine, but redefine `(inv_)link_lkj` to their corresponding `b::Bijector` definitions.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
+# See stan doc for parametrization method:
+# https://mc-stan.org/docs/2_23/reference-manual/correlation-matrix-transform-section.html
+# (7/30/2020) their "manageable expression" is wrong...
+
+struct CorrBijector <: Bijector{2} end
+
+(b::CorrBijector)(X::AbstractMatrix{<:Real}) = link_lkj(X)
```
torfjelde, line 7:
```suggestion
function (b::CorrBijector)(X::AbstractMatrix{<:Real})
    w = cholesky(x).U.data # convert to dense matrix
    r = link_w_lkj(w) 
    return r
end
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,109 @@
...
+(b::CorrBijector)(X::AbstractArray{<:AbstractMatrix{<:Real}}) = map(b, X)
+
+(ib::Inverse{<:CorrBijector})(Y::AbstractMatrix{<:Real}) = inv_link_lkj(Y)
```
torfjelde, line 10:
```suggestion
function (ib::Inverse{<:CorrBijector})(Y::AbstractMatrix{<:Real})
    w = inv_link_w_lkj(y)
    return w' * w
end
```

src/bijectors/corr.jl
torfjelde: See other comment; we likely need this for custom adjoints.

src/bijectors/corr.jl
torfjelde: Agree with removal of this; see comments on `(b::CorrBijector)` and `(ib::Inversed{<:CorrBijector})(y)` above.

src/bijectors/corr.jl
torfjelde: See above comments.

devmotion:
The problem seems to be that DistributionsAD 0.6.3 doesn't support FillArrays 0.9 yet which is however pulled in before DistributionsAD (it's only used in the tests). DistributionsAD 0.1 seems to be compatible with FillArrays 0.9 though... I guess it didn't upper bound the compatibility.

src/bijectors/corr.jl
devmotion: For Zygote (or rather ChainRules nowadays) it's definitely possible to define adjoints for `(::Bijector)(...)` (see e.g. the support of Distances in Zygote). Does it not work for other AD backends?

yiyuezhuo: I defined custom adjoint on `link_w_lkj` but not on `link_lkj`, so link_w_lkj is just an "anonymous" code block requiring a custom gradient.

src/bijectors/corr.jl
yiyuezhuo: `z` is used in three reverse AD backends. Though it's possible to optimize ForwardDiff specifically.

src/bijectors/corr.jl
yiyuezhuo: Merged.

src/bijectors/corr.jl
yiyuezhuo: Merged.

src/bijectors/corr.jl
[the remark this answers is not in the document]
yiyuezhuo: Renamed.

src/bijectors/corr.jl
yiyuezhuo: Merged.

src/bijectors/corr.jl
yiyuezhuo: Changed.

src/bijectors/corr.jl
yiyuezhuo: Merged.

src/bijectors/corr.jl
yiyuezhuo: Changed.

src/bijectors/corr.jl
yiyuezhuo: Changed looping order and added `@inbounds`.

src/bijectors/corr.jl
yiyuezhuo: Changed.

src/bijectors/corr.jl
yiyuezhuo: Replaced it with `zero(eltype(y))` as suggested by @torfjelde . Does "initial summand" means something like `left = (K-1-1) * log(1 - tanh(y[1, 2])^2)`?

src/bijectors/corr.jl
yiyuezhuo: Removed the allocations.

src/bijectors/corr.jl
devmotion: Hmm I'm not sure how that relates to my comment? I just wanted to point out that in Zygote and ChainRules you can also define adjoints for functions such as `(ib::Inverse{<:CorrBijector})(y::AbstractMatrix{<:Real}) = ...` if needed - there is no need to introduce some function `inv_link_lkj` for this purpose.

src/bijectors/corr.jl
devmotion: You mean because of the adjoint? At least for Zygote that should work in exactly the same way, regardless of how you define the function.

src/bijectors/corr.jl
devmotion: Yes.

BTW one should also not implement `log(1 - tanh(y[...])^2)` but use the fact that this is equivalent to `-2 * log(exp(y) + exp(-y))` for a numerically stable implementation.

src/bijectors/corr.jl
torfjelde: Woah, really? If so I guess there's no need. Back when we started this overhaul of Bijectors, it was not possible to define adjoints for those kind of signatures. If that's not the case anymore, then I agree with your comments @devmotion

devmotion: In Zygote you can write something like https://github.com/FluxML/Zygote.jl/blob/956575ee2c732dee25324b59ba43fbb471a52d9a/src/lib/distances.jl#L3-L10. However, we should definitely use ChainRules (https://www.juliadiff.org/ChainRulesCore.jl/dev/index.html#frule-and-rrule) which supports this as well.

yiyuezhuo: I mean, let's say these is a function `f` which needs AD:

```julia
function f(x)
  y = f1(x)
  z = f2(y)
  return z
end
```
where `f1` is AD-friendly and `f2` needs a custom adjoint. So I wrote an adjoint for `f2` only. Are you suggesting to write an adjoint for `f` and call `pullback` for `f1` there? It's not clear to me.

src/bijectors/corr.jl
yiyuezhuo: Are you suggest `w = w .* z`? It's wrong since diagnonal elements are not zeros. And `z` is used in 3 Reverse AD backend.

src/bijectors/corr.jl
yiyuezhuo: Replaced.

src/bijectors/corr.jl
yiyuezhuo: Replaced.

src/compat/tracker.jl
yiyuezhuo: Those code are removed.

src/compat/tracker.jl
yiyuezhuo: These code are removed.

src/bijectors/corr.jl
yiyuezhuo: These comments are removed.

src/bijectors/corr.jl
yiyuezhuo: Updated.

src/bijectors/corr.jl
yiyuezhuo: Updated.

src/bijectors/corr.jl
yiyuezhuo: This is removed. While I don't think the resultant `-logabsdetjac(inv(b),(b(X)))` is very easy to understand.

src/bijectors/corr.jl
yiyuezhuo: Merged `logabsdetjac_lkj_inv` into `logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})`.

yiyuezhuo:
I still don't think `inv_link_w_lkj` and `link_w_lkj` should be removed. They represent operation not involved in cholesky, so I found they're useful to debug and lay the foundation for something like `LKJ_corr` distribution, defined by Stan:

https://mc-stan.org/docs/2_24/functions-reference/cholesky-lkj-correlation-distribution.html

I personally want `LKJ_corr` more than `LKJ`, as stan [doc](https://mc-stan.org/docs/2_24/functions-reference/lkj-correlation.html) say:

> However, it is much better computationally to work directly with the Cholesky factor of  Σ, so this distribution should never be explicitly used in practice.

But `Distributions.jl` provides only previous one so I have no choice.

On the other hand, I think defining `adjoint` for `(b::CorrBijector)(x::AbstractMatrix{<:Real})` is more clumsy, as shown in another comment.

devmotion:
> I personally want LKJ_corr more than LKJ

Sure, I understand that. However, IMO the correct approach is then to define such a `MatrixDistribution`, maybe called `LKJCholesky`, that operates on the Cholesky factor. IMO as long as there exists no such distribution, code should be written as efficient and readable as possible for `LKJ`. IMO the code is much more difficult to read if we add a lot of functions that are not used for dispatch and have to be tracked down even though we could just put the implementation into one single function.

> They represent operation not involved in cholesky

I can see that we might not want to add an adjoint for the Cholesky decomposition which is defined elsewhere. However, the argument doesn't apply to
```julia
function (ib::Inverse{<:CorrBijector})(y::AbstractMatrix{<:Real})
    w = inv_link_w_lkj(y)
    return w' * w
end
```
I don't see why it would be helpful here to introduce a separate function `inv_link_w_lkj` (with an arguably quite cryptic name) if we can define the adjoint for `(ib::Inverse{<:CorrBijector})(y::AbstractMatrix{<:Real})` directly (if needed).

yiyuezhuo:
Ok.. I have tried to remove `inv_link_w_lkj` but it seems that the method does not work for `ReverseDiff` (while works for `Zygote` and `Tracker`).  MWE:

```julia
using ReverseDiff: @grad, gradient

struct F
end

function (::F)(x)
    return sum(2*x)
end

(f::F)(x) = track(f, x)  # similar code works for `Tracker`
@grad function (f::F)(x_tracked)  # `function (::F)(x_tracked)` got same result
    x = value(x_tracked)
    return 2*x, function(Δ)
        Δx = similar(x)
        fill!(Δx, 2)
    end
end
```

Raised error:

```
LoadError: MethodError: no method matching gensym(::Expr)
Closest candidates are:
  gensym(!Matched::Symbol) at expr.jl:15
  gensym(!Matched::String) at expr.jl:12
  gensym() at expr.jl:10
  ...
in expression starting at In[16]:2

Stacktrace:
 [1] @grad(::LineNumberNode, ::Module, ::Any) at /home/yiyuezhuo/.julia/packages/ReverseDiff/Thhqg/src/macros.jl:182
 [2] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1091
```

ReverseDiff version: `[37e2e3b7] ReverseDiff v1.4.2` (latest).

devmotion:
Seems there's a bug in `ReverseDiff.@grad`.

yiyuezhuo:
Yes, I found this [issue](https://github.com/JuliaDiff/ReverseDiff.jl/issues/135). But there's no fix after 2 month... Is renaming `inv_link_w_lkj` to `inv_link_chol_lkj` a workaround?

devmotion:
Maybe check if the fix mentioned in the issue resolves the problem and open a PR to ReverseDiff if it works?

devmotion:
Although, I guess

> Is renaming inv_link_w_lkj to _inv_link_chol_lkj a workaround?

might be good enough for now. Let's just open an issue in Bijectors that reminds us that we should fix it at a later point, i.e., that we should remove all the unnecessary functions for LKJ (and all other bijectors).

torfjelde:
That sounds good to me:)

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+        # left += (K-i-1) * log(1 - tanh(y[i, j])^2) # lacks numerically stable
+        left += (K-i-1) * 2 * (log(2) - log(exp(y[i,j]) + exp(-y[i,j])))
+    end
+    
+    right = zero(eltype(y))
+    @inbounds for j=2:K, i=1:(j-1)
+        right += log(cosh(y[i, j])^2)
+    end
+    
+    return left / 2 - right
```
devmotion, line 37: It seems the multiplication with 2 here cancels with the division by 2 below? Additionally, the computation of `right` can also be simplified to `2 * (-log(2) + log(exp(-y[i, j]) + exp(y[i, j]))`. Taken together take makes `(K - i + 1) * (log(2) - log(exp(y[-y[i, j]) + exp(y[i, j])))` for each combination of `i` and `j`. Therefore I'd suggest I'd suggest
```suggestion
    result = float(zero(eltype(y))
    for j in 2:K, i in 1:(j - 1)
        @inbounds abs_y_i_j = abs(y[i, j])
        result += (K - i + 1) * (logtwo - (abs_y_i_j + log1pexp(-2 * abs_y_i_j)))
    end
    
    return result
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+end
+function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
+    return -logabsdetjac(inv(b),(b(X))) # It may be more efficient if we can use un-contraint value to prevent call of b
+end
+logabsdetjac(b::CorrBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = mapvcat(X) do x
+    logabsdetjac(b, x)
+end
```
devmotion, line 44:
```suggestion
function logabsdetjac(b::CorrBijector, X::AbstractArray{<:AbstractMatrix{<:Real}})
    return mapvcat(X) do x
        logabsdetjac(b, x)
    end
end
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+    
+    right = zero(eltype(y))
+    @inbounds for j=2:K, i=1:(j-1)
+        right += log(cosh(y[i, j])^2)
+    end
+    
+    return left / 2 - right
+end
+function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
+    return -logabsdetjac(inv(b),(b(X))) # It may be more efficient if we can use un-contraint value to prevent call of b
```
devmotion, line 40: The comment is not clear to me. In any case, we should try to stick to the limit of 92 characters per line and put it above the code.
```suggestion
    return -logabsdetjac(inv(b), b(X))
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+    and so
+
+        z₍ᵢ,ⱼ₎ = w₍ᵢ,ⱼ₎ * ∏ₖ₌₁ⁱ⁻¹ (1 / √(1 - z₍ₖ,ⱼ₎²))
+               = (w₍ᵢ,ⱼ₎ * / √(1 - z₍ᵢ₋₁,ⱼ₎²)) * (∏ₖ₌₁ⁱ⁻² 1 / √(1 - z₍ₖ,ⱼ₎²))
+               = (w₍ᵢ,ⱼ₎ * / √(1 - z₍ᵢ₋₁,ⱼ₎²)) * (w₍ᵢ₋₁,ⱼ₎ * ∏ₖ₌₁ⁱ⁻² 1 / √(1 - z₍ₖ,ⱼ₎²)) / w₍ᵢ₋₁,ⱼ₎
+               = (w₍ᵢ,ⱼ₎ * / √(1 - z₍ᵢ₋₁,ⱼ₎²)) * (z₍ᵢ₋₁,ⱼ₎ / w₍ᵢ₋₁,ⱼ₎)
+               = (w₍ᵢ,ⱼ₎ / w₍ᵢ₋₁,ⱼ₎) * (z₍ᵢ₋₁,ⱼ₎ / √(1 - z₍ᵢ₋₁,ⱼ₎²))
+
+    which is the above implementation.
+    =#
```
devmotion, line 107: IMO such explanations should be part of a docstring.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+    return -logabsdetjac(inv(b),(b(X))) # It may be more efficient if we can use un-contraint value to prevent call of b
+end
+logabsdetjac(b::CorrBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = mapvcat(X) do x
+    logabsdetjac(b, x)
+end
+
+
+function _inv_link_chol_lkj(y)
+    @assert size(y, 1) == size(y, 2)
+    K = size(y, 1)
```
devmotion, line 49: Again, better use
```suggestion
    K = LinearAlgebra.checksquare(y)
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+
+    z = tanh.(y)
+    w = similar(z)
+    
+    w[1,1] = 1
+    @inbounds for j in 1:K
+        w[1, j] = 1
+    end
```
devmotion, line 57: `w[1, 1]` is set twice?

devmotion: In general, this whole block could just be part of the loop below, it seems?
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+
+    @inbounds for j in 1:K
+        for i in j+1:K
+            w[i, j] = 0
+        end
+        for i in 2:j
+            w[i, j] = w[i-1, j] * sqrt(1 - z[i-1, j]^2)
+        end
```
devmotion, line 65: The array `z` and the loops below and above are not needed, it seems you could just write
```suggestion
        w[1, j] = 1
        for i in 2:j
            u = w[i-1, j]
            z = tanh(y[i-1, j])
            w[i-1, j] = u * z
            w[i, j] = u * sqrt(1 - z^2)
        end
        for i in (j + 1):K
            w[i, j] = 0
        end
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+logabsdetjac(b::CorrBijector, X::AbstractArray{<:AbstractMatrix{<:Real}}) = mapvcat(X) do x
+    logabsdetjac(b, x)
+end
+
+
+function _inv_link_chol_lkj(y)
+    @assert size(y, 1) == size(y, 2)
+    K = size(y, 1)
+
+    z = tanh.(y)
```
devmotion, line 51: `z` is not needed, see below:
```suggestion
```

src/bijectors/corr.jl
devmotion: BTW can't we just return a triangular matrix so we don't have to fill the other elements?

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+            w[i, j] = w[i-1, j] * sqrt(1 - z[i-1, j]^2)
+        end
+    end
+
+    @inbounds for j in 2:K
+        for i in 1:j-1
+            w[i, j] = w[i, j] * z[i, j]
+        end
+    end
+
```
devmotion, line 73: Not needed, see above:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+            w[i, j] = w[i, j] * z[i, j]
+        end
+    end
+    
+    return w
+end
+
+function _link_chol_lkj(w)
+    @assert size(w, 1) == size(w, 2)
+    K = size(w, 1)
```
devmotion, line 79: Again I'd suggest
```suggestion
    K = LinearAlgebra.checksquare(w)
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+
+    z = zero(w)
+    
+    @inbounds for j=2:K
+        z[1, j] = w[1, j]
+    end
```
devmotion, line 85: Again I guess this loop can just be integrated with the loop below.
```suggestion
    z = similar(w, float(eltype(w)))
```

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
...
+    @inbounds for j=3:K, i=2:j-1
+        p = w[i, j]
+        for ip in 1:(i-1)
+            p /= sqrt(1-z[ip, j]^2)
+        end
+        z[i,j] = p
+    end
+    
+    y = atanh.(z)
+    return y
```
devmotion, line 117: The loop from above can be integrated here, and also I don't think it's necessary to create an additional array `y` (BTW again I'm wondering if we couldn't just return a triangular matrix - this would avoid having to fill all the zeros):
```suggestion
    @inbounds for j in 1:K
        z[i, j] = j == 1 ? 0 : atanh(w[1, j])
        for i in 2:(j-1)
            p = w[i, j]
            for ip in 1:(i-1)
                p /= sqrt(1 - z[ip, j]^2)
            end
            z[i,j] = atanh(p)
        end
        for i in j:K
            z[i, j] = 0
        end
    end
    
    return z
```

src/compat/reversediff.jl
```diff
@@ -180,4 +181,105 @@ lower(A::TrackedMatrix) = track(lower, A)
     return lower(Ad), Δ -> (lower(Δ),)
 end
 
+
+_inv_link_chol_lkj(y::TrackedMatrix) = track(_inv_link_chol_lkj, y)
+@grad function _inv_link_chol_lkj(y_tracked)
```
devmotion, line 186: The suggestions above also apply to the custom adjoints.

src/compat/reversediff.jl
```diff
@@ -180,4 +181,105 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+
+    w1 = copy(w) # cache result
+
+    @inbounds for j in 2:K
+        for i in 1:j-1
+            w[i, j] = w[i, j] * z[i, j]
+        end
+    end
+
+    return w, Δw -> begin
```
devmotion, line 217: IMO pullbacks should be defined as a function, as suggested in the ChainRules docs. I.e., rather write
```julia
    function pullback_inv_link_chol_lkj(Δw)
        ...
    end

    return w, pullback_inv_link_chol_lkj

File: test/ad/distributions.jl [hunk 64]
```diff
@@ -307,6 +314,7 @@
         DistSpec((df, A) -> InverseWishart(df, to_posdef(A)), (3.0, A), B, to_posdef),
         DistSpec((df, A) -> TuringWishart(df, to_posdef(A)), (3.0, A), B, to_posdef),
         DistSpec((df, A) -> TuringInverseWishart(df, to_posdef(A)), (3.0, A), B, to_posdef),
+        DistSpec(() -> LKJ(3, 1.), (), A, to_corr), # AD for parameters of LKJ requires more DistributionsAD supports
```
devmotion, line 317: Can you add it just to the broken distributions? We should make sure to open an issue at DistributionsAD if it's not working, I guess it's difficult to keep track of it if it's just mentioned in the code.

test/transform.jl
```diff
@@ -196,6 +196,34 @@ let
 end
 end
 
+@testset "correlation matrix" begin
+let
```
devmotion, line 200: The `let` block should not be needed?
```suggestion
```

devmotion (changes requested):

src/bijectors/corr.jl
yiyuezhuo: It seems that `abs` is not required? Let x = y[i,j]

(K-i-1) * (log(2) - log(exp(x) + exp(-x))) - log(cosh(y[i, j])^2)
(K-i-1) * (log(2) - log(2cosh(x))) - 2log(abs(cosh(x)))
(K-i-1) * (log(2) - log(2) - log(cosh(x))) - 2log(cosh(x))
(K-i+1) * (-log(cosh(x)))
(K-i+1) * ( -log( (1+exp(-2x))/(2exp(-x)) ) )
(K-i+1) * ( -log(1+exp(-2x)) + log(2exp(-x)))
(K-i+1) * (-log1pexp(-2x) + logtwo -x)
(K-i+1) * (logtwo - (x + log1pexp(-2x)))

src/bijectors/corr.jl
yiyuezhuo: Stan's doc lists only the inverse `logabsdetjac`. I think that they just calculate `logabsdet_inv(b_X)` instead of `logabsdet_inv(b(X))` (an extra `b` call. It's even more inefficient considering that `b` call cholesky decomposition), since they keep the un-constraint value `b_X` to that phase.

On the other hand, I found Turing's usage of `logpdf_with_trans` makes it impossible to remove the extra cost.

src/bijectors/corr.jl
devmotion: Mathematically it doesn't matter if you move `x` or `-x` out of the logarithm but for increased numerical stability it should be the larger of both terms, i.e. `abs(x)` (as done e.g. also in the implementation of StatsFuns.logaddexp).

test/transform.jl
yiyuezhuo: The `let` usage is consistent with the above `matrix` testset.

devmotion: Oh that's not a good example and some quite old code I guess 😄 Since you test only one specific distribution you could just remove the `for` loop as well, in addition to the `let` statement, and define `dist = ...`.

src/bijectors/corr.jl
yiyuezhuo: I see, it's useful when `-2x` is positive.

src/bijectors/corr.jl
yiyuezhuo: It is not needed by ForwardDiff, but it's not free for other 3 reversediff backends since it can be used to avoid an extra tanh call.

yiyuezhuo: There's a test to ensure input and output have same type:
https://github.com/TuringLang/Bijectors.jl/blob/b0aaa98f90958a167a0b86c8e8eca9b95502c42d/test/transform.jl#L66
So a zero filling is necessary, a triangular matrix just postpone the filling to another point.

src/bijectors/corr.jl
yiyuezhuo: The `z` without `atanh` is useful for 3 reversediff backends. For example:
```
Δw[ip, j] += Δz[i,j] * w[i,j] * pd / (1-z[ip,j]^2) * z[ip,j]
```

test/transform.jl
yiyuezhuo: Removed `let` and simplified the test.

test/ad/distributions.jl
yiyuezhuo: Added broken test.

src/compat/reversediff.jl
yiyuezhuo: Refactored.

src/bijectors/corr.jl
yiyuezhuo: Yes, integrated.

src/bijectors/corr.jl
yiyuezhuo: Moved it to docstring.

src/bijectors/corr.jl
yiyuezhuo: Added explanation.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,118 @@
+# See stan doc for parametrization method:
+# https://mc-stan.org/docs/2_23/reference-manual/correlation-matrix-transform-section.html
+# (7/30/2020) their "manageable expression" is wrong...
```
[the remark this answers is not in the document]
yiyuezhuo: Moved it to docstring and added some content.

src/bijectors/corr.jl
devmotion: It might make sense for the reverse-mode AD backends to keep `z` (and even in that case it might be more efficient to fill both `y` and `z` in one loop since then you don't have to perform all the useless `atanh(0)` computations), but for these we define the custom adjoints so the argument shouldn't apply here?

src/bijectors/corr.jl
devmotion: > It is not needed by ForwardDiff, but it's not free for other 3 reversediff backends since it can be used to avoid an extra tanh call.

But it's the standard method here that is used for ForwardDiff, for reverse-mode you defined custom adjoints anyways. And even in the reverse-mode case it probably makes sense to move everything into one loop and fill both `z` and `w` there to avoid all the unnecessary computations. Then you can even make `z` a `LowerTriangular` which would probably speed up the computations in the backward pass.

src/bijectors/corr.jl
devmotion: And in the reverse-mode implementation you can even make `z` an `UpperTriangular` so you don't have to fill the zeros and could maybe speed up the computations in the backward pass.

src/bijectors/corr.jl
devmotion: > There's a test to ensure input and output have same type:

That seems weird but shouldn't be touched in this PR. @torfjelde Is that a general design choice?

src/compat/tracker.jl
```diff
@@ -440,3 +440,105 @@ lower(A::TrackedMatrix) = track(lower, A)
...
+        @inbounds for j=2:K
+            Δw[j, j] = 0
+            Δtmp = zero(eltype(Δz)) # Δtmp_mat[j-1,j]
+            for i in (j-1):-1:2
+                p = w[i, j] / tmp_mat[i-1, j]
+                ftmp = sqrt(1 - p^2)
+                d_ftmp_p = -p / ftmp
+                d_p_tmp = -w[i,j] / tmp_mat[i-1, j]^2
+
+                Δp = Δz[i,j] / (1-p^2) + Δtmp * tmp_mat[i-1, j] * d_ftmp_p # TODO: simplify
```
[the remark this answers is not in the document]
yiyuezhuo: It should be removed in commit `50b837d`, but I hard reset a version which removed the comment and forget to remove it again... I removed it in latest commit now.

src/bijectors/corr.jl
```diff
@@ -0,0 +1,130 @@
+"""
+    CorrBijector <: Bijector{2}
+
+A bijector implementation of Stan's parametrization method for Correlation matrix:
+https://mc-stan.org/docs/2_23/reference-manual/correlation-matrix-transform-section.html
+
+Note:(7/30/2020) their "manageable expression" is wrong, used expression is derived from 
+scratch.
```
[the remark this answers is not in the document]
yiyuezhuo: Added more description.

yiyuezhuo:
> I have just two minor comments left, IMO we can merge this if they are addressed and tests pass. Could you also bump the version number to 0.8.3 so that we can make a new release (**after we fixed the Distribution bounds**)?
> 
> Great work and thanks a lot for this implementation (and your patience with all of my comments 😄)!

@devmotion I appreciate your comment and suggestions very much and have learned many Julia best practices and numerical tricks from them. 🎉

devmotion (approved):

devmotion:
Thanks again @yiyuezhuo!

PR #133, opened 2020-08-23
Use separate environment for tests

devmotion:
Since we dropped Julia < 1.3, we can use the [new way of specifying test dependencies on Julia >= 1.2](https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1). This seems to fix the issue with Zygote not being compatible with FillArrays 0.9 as well (before Distributions 0.23.9 pulled in FillArrays 0.9, but it seems with this new setup Pkg realizes immediately that it should use Distributions 0.23.8 in the tests). I guess we should change the test dependencies in the same way in all other packages that require Julia >= 1.2.


devmotion:
@torfjelde This is the only change needed before tagging a new release.

torfjelde (approved):
Awesome stuff @devmotion ! Thanks!

torfjelde:
Btw, should we bump the minor version rather than the patch version? Given that we're now sort of dropping support for Julia 1.3?

devmotion:
According to ColPrac [dropping support for older Julia versions must be a non-breaking feature](http://colprac.sciml.ai/#dropping-support-for-earlier-versions-of-julia). If we would be post-1.0.0, then bumping the minor version would make sense since then older Julia versions could still get patch releases. However, since we are pre-1.0.0, I guess we should not bump the minor version since this would lead to a breaking release which it isn't.

torfjelde:
Ah, reading that it makes sense :+1:

PR #136, opened 2020-08-30
Fix for #135

### LINKED ISSUE #135: rand call to dist of MvTransformed not using rng
Minor thing I noticed:

https://github.com/TuringLang/Bijectors.jl/blob/a3ece9122b8e1c9aa3f81105f6a384d85a09e856/src/transformed_distribution.jl#L134

should be:
```julia
res = hcat([td.transform(rand(rng, td.dist)) for i = 1:num_samples]...)
```

torfjelde:
Fixes #135.

### DIFF
--- a/src/transformed_distribution.jl
+++ b/src/transformed_distribution.jl
@@ -131,7 +131,7 @@ rand(td::MvTransformed) = td.transform(rand(td.dist))
 rand(rng::AbstractRNG, td::MvTransformed) = td.transform(rand(rng, td.dist))
 # TODO: implement more efficiently for flows
 function rand(rng::AbstractRNG, td::MvTransformed, num_samples::Int)
-    res = hcat([td.transform(rand(td.dist)) for i = 1:num_samples]...)
+    res = hcat([td.transform(rand(rng, td.dist)) for i = 1:num_samples]...)
     return res
 end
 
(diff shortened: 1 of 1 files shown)


src/transformed_distribution.jl
```diff
@@ -131,7 +131,7 @@ rand(td::MvTransformed) = td.transform(rand(td.dist))
 rand(rng::AbstractRNG, td::MvTransformed) = td.transform(rand(rng, td.dist))
 # TODO: implement more efficiently for flows
 function rand(rng::AbstractRNG, td::MvTransformed, num_samples::Int)
-    res = hcat([td.transform(rand(td.dist)) for i = 1:num_samples]...)
+    res = hcat([td.transform(rand(rng, td.dist)) for i = 1:num_samples]...)
```
devmotion, line 134: Maybe a good opportunity to change it to
```suggestion
    res = reduce(hcat, [td.transform(rand(rng, td.dist)) for _ in 1:num_samples])
```
or something like
```suggestion
    samples = rand(rng, td.dist, num_samples)
    res = reduce(hcat, map(axes(samples, 2)) do i
        return td.transform(view(samples, :, i))
    end)
```

torfjelde: Good idea! Done :+1:

PR #142, opened 2020-10-10
Fix maporbroadcast for ReverseDiff

mohamed82008:
This PR fixes https://github.com/TuringLang/Turing.jl/issues/1385 by making sure broadcasting is used instead of mapping for ReverseDiff. Broadcasting mixed TrackedArrays and arrays of TrackedReals breaks Tracker but ReverseDiff can handle it.

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -243,8 +243,6 @@ end
 
 include("interface.jl")
 
-# Broadcasting here breaks Tracker for some reason
-maporbroadcast(f, x::AbstractArray{<:Any, N}...) where {N} = map(f, x...)
 maporbroadcast(f, x::AbstractArray...) = f.(x...)
 
 # optional dependencies
--- a/src/compat/reversediff.jl
+++ b/src/compat/reversediff.jl
@@ -1,7 +1,7 @@
 module ReverseDiffCompat
 
 using ..ReverseDiff: ReverseDiff, @grad, value, track, TrackedReal, TrackedVector, 
-    TrackedMatrix, TrackedArray
+    TrackedMatrix
 using Requires, LinearAlgebra
 
 using ..Bijectors: Log, SimplexBijector, maphcat, simplex_link_jacobian, 
--- a/src/compat/reversediff.jl
+++ b/src/compat/reversediff.jl
@@ -46,8 +46,6 @@ function Base.maximum(d::LocationScale{<:TrackedReal})
     end
 end
 
-maporbroadcast(f, x::Union{AbstractArray, TrackedArray, AbstractArray{<:TrackedReal}}...) = f.(x...)
-
 logabsdetjac(b::Log{1}, x::Union{TrackedVector, TrackedMatrix}) = track(logabsdetjac, b, x)
 @grad function logabsdetjac(b::Log{1}, x::AbstractVector)
     return -sum(log, value(x)), Δ -> (nothing, -Δ ./ value(x))
--- a/src/compat/reversediff.jl
+++ b/src/compat/reversediff.jl
@@ -1,7 +1,7 @@
 module ReverseDiffCompat
 
 using ..ReverseDiff: ReverseDiff, @grad, value, track, TrackedReal, TrackedVector, 
-    TrackedMatrix
+    TrackedMatrix, TrackedArray
 using Requires, LinearAlgebra
 
 using ..Bijectors: Log, SimplexBijector, maphcat, simplex_link_jacobian, 
@@ -45,6 +45,8 @@ function Base.maximum(d::LocationScale{<:TrackedReal})
         return m
     end
 end
+
+maporbroadcast(f, x::Union{AbstractArray, TrackedArray, AbstractArray{<:TrackedReal}}...) = f.(x...)
 
 logabsdetjac(b::Log{1}, x::Union{TrackedVector, TrackedMatrix}) = track(logabsdetjac, b, x)
 @grad function logabsdetjac(b::Log{1}, x::AbstractVector)
--- a/src/compat/tracker.jl
+++ b/src/compat/tracker.jl
@@ -12,16 +12,18 @@ using .Tracker: Tracker,
 using Compat: eachcol
 using LinearAlgebra
 
-# Broadcasting here breaks Tracker for some reason
-maporbroadcast(f, x::Union{AbstractArray, TrackedArray, AbstractArray{<:TrackedReal}}...) = map(f, x...)
+# Broadcasting here breaks Tracker
+const TrackedT = Union{TrackedArray, AbstractArray{<:TrackedReal}}
 maporbroadcast(f, x::TrackedArray...) = f.(x...)
-function maporbroadcast(
-    f,
-    x1::TrackedArray{T, N},
-    x::AbstractArray{<:TrackedReal}...,
-) where {T, N}
-    return f.(convert(Array{TrackedReal{T}, N}, x1), x...)
-end
+maporbroadcast(f, x::Union{TrackedArray, AbstractArray{<:TrackedReal}}...) = map(f, x...)
+maporbroadcast(f, x1::TrackedT, x2::AbstractArray) = map(f, x1, x2)
+maporbroadcast(f, x1::AbstractArray, x2::TrackedT) = map(f, x1, x2)
+maporbroadcast(f, x1::TrackedT, x2::AbstractArray, x3::AbstractArray) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::AbstractArray, x2::TrackedT, x3::AbstractArray) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::AbstractArray, x2::AbstractArray, x3::TrackedT) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::TrackedT, x2::TrackedT, x3::AbstractArray) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::AbstractArray, x2::TrackedT, x3::TrackedT) = map(f, x1, x2, x3)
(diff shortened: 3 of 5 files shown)


src/compat/tracker.jl
```diff
@@ -12,6 +12,8 @@ using .Tracker: Tracker,
 using Compat: eachcol
 using LinearAlgebra
 
+# Broadcasting here breaks Tracker for some reason
+maporbroadcast(f, x::Union{AbstractArray, TrackedArray, AbstractArray{<:TrackedReal}}...) = map(f, x...)
 maporbroadcast(f, x::TrackedArray...) = f.(x...)
```
devmotion, line 17: The definitions of `maporbroadcast` are quite dangerous since they easily lead to invalidations, similar to
```julia
julia> f(x::AbstractArray...) = 1

julia> f(x::Matrix...) = 2
f (generic function with 2 methods)

julia> f()
2
```
Fortunately, it seems to work in the case of `Union`s (at least with Julia 1.5.2),
```julia
julia> g(x::AbstractVector...) = 1

julia> g(x::AbstractVecOrMat...) = 2
g (generic function with 2 methods)

julia> g()
1
```
so this definition here might be the most problematic one.

mohamed82008: I mean we never really call the function without arguments. And unless there is a better solution that works, invalidations are a secondary concern.

devmotion: I think it would be easy to fix this issue - and at the same time improve the current design and implementation of `maporbroadcast`. Hence I don't think this problem should be a secondary concern.

Instead of implementing `maporbroadcast(f, x...)` for special types of `x` one could just define two traits `MapStyle` and `BroadcastStyle` (well, maybe use a different name...), and then define
```julia
maporbroadcast(f, x::AbstractArray...) = maporbroadcast(f, maporbroadcaststyle(x...), x...)

struct MapStyle end
struct BroadcastStyle end

maporbroadcast(f, ::MapStyle, x::AbstractArray...) = map(f, x...)
maporbroadcast(f, ::BroadcastStyle, x::AbstractArray...) = f.(x...)

maporbroadcaststyle(::Tuple{}) = MapStyle() # or BroadcastStyle, but maybe the type inference of map is better?
maporbroadcaststyle(::AbstractArray) = BroadcastStyle()
function maporbroadcaststyle(x::AbstractArray, y::AbstractArray, ys::AbstractArray...)
    return maporbroadcaststyle(maporbroadcaststyle(x), maporbroadcaststyle(y, ys...))
end

# promotion rule
maporbroadcaststyle(::MapStyle, ::MapStyle) = MapStyle()
maporbroadcaststyle(::BroadcastStyle, ::BroadcastStyle)
maporbroadcaststyle(::MapStyle, ::BroadcastStyle) = MapStyle()
maporbroadcaststyle(::BroadcastStyle, ::MapStyle) = MapStyle()

# to be implemented for TrackedArray etc.
maporbroadcaststyle(::TrackedArray) = MapStyle()
```

BTW I wanted to have a look at the corresponding implementation in DistributionsAD but I noticed that actually `maporbroadcast` is removed there.

mohamed82008: Please open another PR with this suggestion after this is merged.

mohamed82008: Your implementation is adding significant complexity to the code so it would be nice to evaluate its benefits in a separate thread, e.g. how many invalidations are eliminated. This PR is just a bugfix.

devmotion:
We should add some tests here, it seems the existing ones did not cover the problem. In what way is `maporbroadcast` actually covered by any tests? Do tests fail if e.g. `map` is used for all arrays?

src/compat/tracker.jl
devmotion: IMO the code is not complex, it's just what you actually want to achieve (base whether `map` or `broadcast` is used on the arguments) without listing all possible combinations of arrays. Thus most importantly, it is easily extendable by just defining `maporbroadcaststyle` for arrays that should use broadcasting (or not, whatever the default is).

mohamed82008:
>Do tests fail if e.g. map is used for all arrays?

No but we get the problem in the Turing issue which is an array of `TrackedReal`s when applying a 1-dim `TruncatedBijector`. `maporbroadcast` is tested but only for correctness not the output type. I can add that issue's case as a test case.

src/Bijectors.jl
```diff
@@ -243,8 +243,6 @@ end
 
 include("interface.jl")
 
-# Broadcasting here breaks Tracker for some reason
-maporbroadcast(f, x::AbstractArray{<:Any, N}...) where {N} = map(f, x...)
```
devmotion, line 247: I still think it would make sense to keep this definition (and only change it for ReverseDiff). If the arrays are of the same dimension, it should be simpler to not invoke the broadcast machinery (for which adjoints require quite complex implementations e.g. in Tracker and Zygote that even make use of ForwardDiff).

test/ad/distributions.jl
```diff
@@ -550,4 +550,10 @@
             )
         end
     end
+
+    @testset "Turing issue 1385" begin
+        dist = arraydist(truncated.(Laplace.(0, [1, 2]), -10.0, 70.0))
+        x = ReverseDiff.track(rand(dist))
+        @test typeof(bijector(dist)(x)) <: ReverseDiff.TrackedArray
+    end
```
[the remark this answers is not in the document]
devmotion: Maybe it would also be good to really test the underlying issue by testing the output of `maporbroadcast` for ReverseDiff (and making sure, of course, that it fails on master) and not use any bijectors at all here.

devmotion (changes requested):
I added some comments. IMO the only thing needed seems to be a special implementation of `maporbroadcast` for ReverseDiff, I don't think it's good to change the default implementation.

In general, it seems the issue with `map` is a more general problem of ReverseDiff (and not of Bijectors) and ideally should be fixed there. But I'm fine with a workaround for ReverseDiff for now.

src/compat/reversediff.jl
```diff
@@ -46,6 +46,8 @@ function Base.maximum(d::LocationScale{<:TrackedReal})
     end
 end
 
+maporbroadcast(f, x::Union{AbstractArray, TrackedArray, AbstractArray{<:TrackedReal}}...) = f.(x...)
+
```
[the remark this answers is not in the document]
devmotion: However (see my comments below), actually I think the default implementation should not be changed.

test/ad/distributions.jl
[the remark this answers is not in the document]
devmotion: BTW it seems the test fails.

src/Bijectors.jl
mohamed82008: Tracker's map also reduces to an array of TrackedReals. So broadcasting by default is needed for 2 out of 3 AD backends. So it makes sense as the default. Only when we have mixed TrackedArrays and arrays of TrackedReals that Tracker fails hence the need for the map workaround. ReverseDiff broadcasting doesn't fail in this case.

src/compat/tracker.jl
```diff
@@ -12,14 +12,18 @@ using .Tracker: Tracker,
...
-end
+maporbroadcast(f, x::Union{TrackedArray, AbstractArray{<:TrackedReal}}...) = map(f, x...)
+maporbroadcast(f, x1::TrackedT, x2::AbstractArray) = map(f, x1, x2)
+maporbroadcast(f, x1::AbstractArray, x2::TrackedT) = map(f, x1, x2)
+maporbroadcast(f, x1::TrackedT, x2::AbstractArray, x3::AbstractArray) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::AbstractArray, x2::TrackedT, x3::AbstractArray) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::AbstractArray, x2::AbstractArray, x3::TrackedT) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::TrackedT, x2::TrackedT, x3::AbstractArray) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::AbstractArray, x2::TrackedT, x3::TrackedT) = map(f, x1, x2, x3)
+maporbroadcast(f, x1::TrackedT, x2::AbstractArray, x3::TrackedT) = map(f, x1, x2, x3)
```
devmotion, line 26: This does not seem like a robust solution, and needs tests. I still think the easiest solution would be to just change the ReverseDiff part in this PR - since this is what you want to fix and what you added a test for.

test/runtests.jl
```diff
@@ -35,5 +35,12 @@ end
 
 if !is_TRAVIS && (GROUP == "All" || GROUP == "AD")
     include("ad/distributions.jl")
+    if AD == "ReverseDiff"
```
devmotion, line 38:
```suggestion
    if AD == "All" || AD == "ReverseDiff" 
```

test/runtests.jl
```diff
@@ -35,5 +35,12 @@ end
...
+        @testset "Turing issue 1385" begin
+            dist = arraydist(truncated.(Laplace.(0, [1, 2]), -10.0, 70.0))
+            x = ReverseDiff.track(rand(dist))
+            @test typeof(bijector(dist)(x)) <: ReverseDiff.TrackedArray
```
devmotion, line 42: Can this test be simplified to just check the output of `maporbroadcast`? Or at least a test with only `maporbroadcast` be added?

devmotion (changes requested):

yebai:
@mohamed82008 can you take another look, and see whether we can address the comments then merge this PR?

yebai:
I'm closing this PR since `arraydist` and `filldist` will be replaced by[ `Product` ](https://github.com/JuliaStats/Distributions.jl/pull/1391) in the future.

PR #146, opened 2020-11-19
Add Distributions 0.24

devmotion:


willtebbutt:
For visibility, this PR is (I _believe_) now blocking for [AbstractGPs](https://github.com/JuliaGaussianProcesses/AbstractGPs.jl/pull/73) upgrade to `FillArrays` 0.10.

devmotion:
Unfortunately, FillArrays 0.10 is still blocked by DistributionsAD - just Distributions 0.24 is compatible with DistributionsAD.

briochemc:
+1 for merging this! (I *think* I need it to release a new version of a package that depends on a package that depends on Bijectors 🙂)

test/ad/distributions.jl
```diff
@@ -34,20 +35,35 @@
...
+    to_positive(x::AbstractArray{<:AbstractArray}) = to_positive.(x)
+
+    # Create vectors in probability simplex.
+    # Custom implementation since `StatsFuns.softmax` is not compatible with Zygote.
+    function to_simplex(x::AbstractArray; dims=1)
+        maxx = maximum(x, dims=dims)
+        y = exp.(x .- maxx)
+        return y ./ sum(y, dims=dims)
+    end
+    to_simplex(x::AbstractArray{<:AbstractArray}; dims=1) = to_simplex.(x; dims=dims)
```
torfjelde, line 56: We already depend on `NNlib` for this exact reason,  you can just replace this with `NNlib.softmax` :+1:

test/ad/utils.jl
```diff
@@ -107,58 +110,65 @@ function test_ad(dist::DistSpec; kwargs...)
...
+                x -> f_allargs(unpack(x, inds, xorig, θorig...)...)
             end
+            test_ad(f_test, xtest, broken; kwargs...)
         end
     end
 end
 
-function test_ad(f, x; rtol = 1e-6, atol = 1e-6, ad = AD)
-    finitediff = FiniteDiff.finite_difference_gradient(f, x)
+function test_ad(f, x, broken; rtol = 1e-6, atol = 1e-6)
```
torfjelde, line 139: Maybe just make `broken = ()` default?

torfjelde (approved):
Looks good! Got a couple of suggested changes, but nothing major. Also out of curiosity: why the change from `FiniteDiff` to `FiniteDifferences`?

devmotion:
> Also out of curiosity: why the change from `FiniteDiff` to `FiniteDifferences`?

I changed it since we use `FiniteDifferences` in the DiststributionsAD tests (from which most updates/fixes are copied).

test/ad/utils.jl
devmotion: Sounds reasonable.

test/ad/distributions.jl
devmotion: Ah nice :+1:

I didn't realize this when writing the tests in DistributionsAD (and there NNlib is an indirect dependency).

devmotion: BTW it seems Bijectors uses only `NNlib.softplus` but nothing esle from `NNlib`?

devmotion:
I applied all suggestions.

test/ad/distributions.jl
torfjelde: Yeah but that's for the same reason I believe. Also, the PR for rational quadratic splines (which no one seems to want to look at :sweat_smile: ) uses `NNlib.softmax` for the above reason.

torfjelde (approved):
Sweet: LGTM!

Issue #154, opened 2021-01-04
Zygote ad backend for normalizing flows


themrzmaster:
Hi,
The Normalizing flow example uses Tracker, a discontinued AD package.
I am trying to fit a NF using Zygote, but I have some problems.

Example:
```
@model function gen()
    x ~ Exponential(0.5)
    y ~ Normal(0, x)
end
s = sample(gen(), NUTS(), 500)
train_data = hcat(s[:x].data, s[:y].data)' |> Array

b = PlanarLayer(2) ∘ LeakyReLU{Float64, 1}(0.05) ∘ RadialLayer(2) ∘ LeakyReLU{Float64, 1}(0.05) ∘ PlanarLayer(2)
d = MvNormal(zeros(2), ones(2))
tb = transformed(d, b);
loss(tb :: Bijectors.TransformedDistribution, x :: Matrix{Float64}) = begin
    return sum(-logpdf(tb, x))
end
function nf_train(tb, x, opt, ps, epochs)
    @showprogress for i ∈ 1:epochs
        train_loss, back = Zygote.pullback(() -> loss(tb, x), ps)
        gs = back(one(train_loss))
        Flux.update!(opt, ps, gs)
    end
end
nf_train(tb, train_data, ADAM(), Zygote.Params(Flux.params(b)), 10)
```

I get the error:

**Mutating arrays is not supported**

on 

`gs = back(one(train_loss))`

Any way I can make this work?

Thanks

devmotion:
I can't run your example and it's a bit difficult to comment without knowing the exact error message. It is also a bit unclear to me which packages you used here - did you load DistributionsAD?

themrzmaster:
@devmotion I've updated the example with the full code.
Yes, I've loaded DistributionsAD
`using Turing, Bijectors, Flux, ProgressMeter, DistributionsAD`

devmotion:
Thanks, now I can run the code. The error is caused by AD problems of the Roots package. The inverse of planar layers is computed using a root-finding algorithm in the Roots package (see https://github.com/TuringLang/Bijectors.jl/blob/a85414420073edba575df13e01196b73dd191386/src/bijectors/planar_layer.jl#L122 and the paper mentioned in the comments), and it seems it is not compatible with Zygote.

devmotion:
[...]

PR #155, opened 2021-01-05
Use NonlinearSolve instead of Roots

devmotion:
Motivated by https://github.com/TuringLang/Bijectors.jl/issues/154, this PR replaces Roots with NonlinearSolve which is the root solving package that replaced Roots in the SciML ecosystem. NonlinearSolve states in its README
> Performance is key: the current methods are made to be highly performant on scalar and statically sized small problems.

and

> The current algorithms should support automatic differentiation, though improved adjoint overloads are planned to be added in the current update.


cpfiffer (approved):
Looks good to me, though maybe @torfjelde has opinions on it. I also agree with adding this instead of Roots -- seems they have a much more AD-friendly perspective.

cpfiffer:
Hm, seems like a fairly big issue -- my imagining is that thunks require some large-ish structural modficiations?

devmotion:
Actually I'm not sure, it could be that it is sufficient to just unthunk always when applying the results of the pullback (if it's not a thunk, unthunking would just be a no-op). But I'm not too familiar with the Zygote internals and hence I created the issue.

If we don't want to wait, my suggestion would be to just mark the failing tests as broken in DistributionsAD and Bijectors and open an issue, so we don't forget about it. IIRC it affects Skellam, Pareto, and NormalInverseGaussian with Zygote.

cpfiffer:
I think that'd be fine. IMO Zygote isn't really the priority AD backend anyway so I don't think we should wait on them.

torfjelde (approved):
Looks good! And I'm fine with the error in Zygote.jl if we make a corresponding issue so as to remind ourselves that this needs to be updated when we have a fix.

devmotion:
OK, I will update the tests in https://github.com/TuringLang/DistributionsAD.jl/pull/143 and then apply the same changes to the Bijectors tests which take more time to run.

PR #157, opened 2021-01-07
Run interface tests, update badges, and fix planar layer

devmotion:


devmotion:
I had a look at the planer layer and found a bug in the implementation of the inverse.

However, one can also rewrite the forward pass in a simpler and, it seems, numerically more stable way. To keep the PR small, I don't include these possible changes here and just try to fix the test errors.

cpfiffer (commented):
Is there a source for the math so I can check it?

devmotion:
The bracket thing is my own derivation to speed up the root finding algorithm (I haven't seen a discussion of it anywhere else), the bugfix (removal of `norm(w, 2)`) can be checked by looking at eq. (22) in the original paper https://arxiv.org/pdf/1505.05770.pdf and plugging in the parameterization of z|| with alpha below the equation (we solve eq. (22) for z = z_|_ + z||).

devmotion:
There are multiple references to the original paper in the source code but maybe not to eq. (22) explicitly at this line.

cpfiffer (approved):
Okay, looks sufficient to me.

PR #158, opened 2021-01-08
Improve planar layer

devmotion:


src/bijectors/planar_layer.jl
```diff
@@ -121,16 +157,21 @@ which implies ``α̂ ∈ [wt_y - |wt_u_hat|, wt_y + |wt_u_hat|]``.
 D. Rezende, S. Mohamed (2015): Variational Inference with Normalizing Flows.
 arXiv:1505.05770
 """
-function find_alpha(wt_y, wt_u_hat, b)
-    # Compute the initial bracket.
+function find_alpha(wt_y::Real, wt_u_hat::Real, b::Real)
```
devmotion, line 160: A bit unrelated since never tested it seems (so maybe it's not useful?): one should define ChainRules rules for this function if one wants to differentiate through the inverse flow - even if it would be optimized by NonlinearSolve at some point, I assume we can always do better here. And everything is available in closed form.

src/bijectors/planar_layer.jl
```diff
@@ -23,74 +23,110 @@ function Base.:(==)(b1::PlanarLayer, b2::PlanarLayer)
...
-    else
-        T = typeof(vec(psi))
-    end
-    log_det_jacobian::T = log.(abs.(1 .+ psi' * u_hat)) # from eq(12)
-    return (rv = transformed, logabsdetjac = log_det_jacobian)
+According to appendix A.1, vector ``û`` defined by
+```math
+û(w, u) = u + (\\log(1 + \\exp{(wᵀu)}) - 1 - wᵀu) \\frac{w}{\\|w\\|²}
+```
+guarantees that the planar layer ``f(z) = z + û tanh(wᵀz + b)`` is invertible.
```
torfjelde, line 45:
```suggestion
guarantees that the planar layer ``f(z) = z + û tanh(wᵀz + b)`` is invertible even if `w, u ∈ ℝᵈ` and `b ∈ ℝ`.
```
It might be a bit redundant, it's just to emphasize that we're not assuming the inputs to be constrained in any way.

src/bijectors/planar_layer.jl
```diff
@@ -23,74 +23,110 @@ function Base.:(==)(b1::PlanarLayer, b2::PlanarLayer)
...
+We can rewrite ``û`` as
+```math
+û = u + (\\log(1 + \\exp{(-wᵀu)}) - 1) \\frac{w}{\\|w\\|²}.
+```
+
+Additionally, we obtain
+```math
+wᵀû = wᵀu + \\log(1 + \\exp{(-wᵀu)}) - 1 = \\log(1 + \\exp{(wᵀu)}) - 1.
```
torfjelde, line 53:
```suggestion
wᵀû = wᵀu + \\log(1 + \\exp{(-wᵀu)}) - wᵀu = \\log(1 + \\exp{(wᵀu)}) - 1.
```

src/bijectors/planar_layer.jl
```diff
@@ -23,74 +23,110 @@ function Base.:(==)(b1::PlanarLayer, b2::PlanarLayer)
...
 end
 
-function (ib::Inverse{<:PlanarLayer})(y::AbstractVector{<:Real})
-    flow = ib.orig
+# Helper functions
+aT_b(a::AbstractVector{<:Real}, b::AbstractVector{<:Real}) = dot(a, b)
+aT_b(a::AbstractVector{<:Real}, b::AbstractMatrix{<:Real}) = permutedims(a) * b
+
+_vec(x::AbstractArray{<:Real}) = vec(x)
+_vec(x::Real) = x
```
torfjelde, line 73: I really hate just have such "utility" functions spread around in different bijector-specific files. Should we maybe put this in `utils.jl` or something?

src/bijectors/planar_layer.jl
```diff
@@ -23,74 +23,110 @@ function Base.:(==)(b1::PlanarLayer, b2::PlanarLayer)
...
+    return (transformed = transformed, wT_û = wT_û, wT_z = wT_z)
+end
+
+(b::PlanarLayer)(z) = _transform(b, z).transformed
 
-    # Find the scalar ``alpha`` from A.1.
-    wt_y = dot(w, y)
-    wt_u_hat = dot(w, u_hat)
-    alpha = find_alpha(wt_y, wt_u_hat, b)
+#=
```
torfjelde, line 87: Is there a reason why this isn't ``` rather than `#=`?

src/compat/tracker.jl
```diff
@@ -225,102 +225,6 @@ end
     end
 end
 
-for header in [
```
torfjelde, line 228: All this ain't necessary anymore? Sweet-sauce!

torfjelde (approved):
Awesome stuff @devmotion ! 
I have one suggested change (typo-fix to one of the docstrings) and just some comments which are non-essential. Happy to have this merged once tests pass!

src/bijectors/planar_layer.jl
devmotion:
```suggestion
guarantees that the planar layer ``f(z) = z + û tanh(wᵀz + b)`` is invertible for all ``w, u ∈ ℝᵈ`` and ``b ∈ ℝ``.
```

src/bijectors/planar_layer.jl
devmotion: No, `-1` is correct here. One obtains the identity by multiplying the equation above with `wᵀ`.

src/bijectors/planar_layer.jl
devmotion: I agree, I considered putting it into a `utils.jl` file first. I was surprised when I noticed that there's no such file yet, and assumed maybe there is a reason for it.

src/bijectors/planar_layer.jl
devmotion: It felt wrong to add a docstring to the generic `forward` method, I thought a regular comment might be better here.

src/compat/tracker.jl
devmotion: Yep, everything seems to work :slightly_smiling_face: I even checked the example from the README, it works and the types (TrackedArray etc.) are the same.

src/bijectors/planar_layer.jl
torfjelde: Ah, sorry yes, but aren't you missing a factor of `wᵀu` ? I.e. shouldn't it me
```suggestion
wᵀû = wᵀu + \\log(1 + \\exp{(-wᵀu)}) - 1 = \\log(1 + \\exp{(wᵀu)}) - 1 - wᵀu.
```

src/bijectors/planar_layer.jl
torfjelde: Yeaaah no I don't think so :sweat_smile: 
But you can either move it to utils.jl to get started or just ignore this comment; we should maybe just make a separate PR for gather all these methods.

src/bijectors/planar_layer.jl
devmotion: No, the second equality is obtained by moving `wᵀu` into the logarithm.

src/bijectors/planar_layer.jl
torfjelde: Ah, nono good point! Just leave it:)

src/bijectors/planar_layer.jl
torfjelde: Maybe just add a "TODO" comment?

src/compat/tracker.jl
torfjelde: Dope stuff:)

src/bijectors/planar_layer.jl
torfjelde: Aaaah ofc, now I see it:
```
log(exp(log(1 + exp(wᵀu)) - wᵀu)) = log((1 + exp(wᵀu)) exp(-wᵀu)) = log(1 + exp(-wᵀu))
```
Sorry about the confusion :+1: Missed the difference in sign initially and got very confused.

src/bijectors/planar_layer.jl
devmotion: I moved them to utils.jl but other methods should be moved in a separate PR.

test/interface.jl
```diff
@@ -255,12 +255,10 @@ end
                 # FIXME: `SimplexBijector` results in ∞ gradient if not in the domain
                 if !contains(t -> t isa SimplexBijector, b)
                     b_logjac_ad = [logabsdet(ForwardDiff.jacobian(b, xs[:, i]))[1] for i = 1:size(xs, 2)]
-                    tol = isclosedform(b) ? 1e-9 : 1e-1
-                    @test logabsdetjac(b, xs) ≈ b_logjac_ad rtol=tol atol=tol
+                    @test logabsdetjac(b, xs) ≈ b_logjac_ad atol=1e-9
```
devmotion, line 258: I noticed that these tests now also pass for planar layers with stricter tolerances. I am not sure how useful `isclosedform` actually is anymore?

torfjelde: Hmm, maybe just leave it for now and open an issue? Might be best to remove in a different PR if so + by default it returns `true` for all bijectors.

devmotion: Sure, I don't want to add anything to this PR :slightly_smiling_face:

PR #161, opened 2021-01-10
update flux @functor on the nf example

themrzmaster:
as described here https://github.com/FluxML/Flux.jl/issues/1339, treelike has been removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,11 +477,11 @@ In those cases, it might be useful to use Flux.jl's `treelike` to extract the pa
 ```julia
 julia> using Flux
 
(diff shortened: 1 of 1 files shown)


devmotion:
`Functors.@functor` is just reexported by Flux, so probably the better approach would be to just implement `Functors.functor` and `Functors.@functor` it in the package and remove it completely from the README. I have some local fixes and will open a PR in the next days.

devmotion:
Fixed by https://github.com/TuringLang/Bijectors.jl/pull/162.

PR #162, opened 2021-01-10
Implement `Functors.functor`

devmotion:
This PR is an alternative to https://github.com/TuringLang/Bijectors.jl/pull/161. As mentioned there, `Flux.@treelike` does not exist anymore. Instead, Flux relies on `Functors.functor` and reexports `Flux.@functor`.

This PR implements `Functors.functor` for bijectors and normalizing flows. Thus users can just call `Flux.params` without implementing `@functor` for the different layers.

A dependency on the lightweight package Functors.jl is added and the optional dependency on Flux is dropped.


test/bijectors/utils.jl
```diff
@@ -192,3 +192,10 @@ function test_logabsdetjac(b::Bijector{0}, xs::AbstractVector; tol=1e-6)
         @test mean(logabsdetjac(b, xs) - logjac_ad) ≤ tol
     end
 end
+
+# Check if `Functors.functor` works properly
+function test_functor(x, xs)
+    _xs, re = Functors.functor(x)
+    @test x == re(_xs)
+    @test _xs == xs
+end
```
torfjelde, line 201: Is this being used anywhere?

torfjelde (changes requested):
This is great stuff!
I have one comment about the testing; it seems like we currently only test a couple of the bijectors?

test/bijectors/utils.jl
devmotion: Yes, in the `norm_flows` tests.

devmotion:
Yes, indeed, so far only `InvertibleBatchnorm`, `PlanarLayer`, and `RadialLayer` are tested in `norm_flows.jl`. I am still a bit confused about the test setup - where would be the best place to add tests for other bijectors? Initially I assume that most interface tests would be handled by `test_bijector` but it seems it is only used in one single test?

torfjelde:
> where would be the best place to add tests for other bijectors? Initially I assume that most interface tests would be handled by test_bijector but it seems it is only used in one single test?

That's correct, but the idea is to start using that approach for all bijectors. So feel free to add it there for now :)

test/bijectors/utils.jl
torfjelde: Ah, missed that :+1:

torfjelde:
Actually, I guess it's somewhat non-trivial to add this to `test_bijector`, as we don't have access to the parameters of the bijector itself?

devmotion:
I assume one would have to specify the parameters explicitly, similar to `test_functor`. However, one could set to `nothing` by default and skip the tests in this case.

devmotion:
This will lead to a stackoverflow error. But in principle this is the idea.

torfjelde:
Oh yeah, rather:

``` julia
function test_bijector(
    b::Bijector,
    xs_true,
    ys_true,
    logjacs_true
    params;
    kwargs...
)
    test_functor(b, params)
    return test_bijector(b, xs_true, ys_true, logjacs_true; kwargs...)
end
```

torfjelde (approved):
I'm happy for now, and we can make it testing the functor functionality part of the tests once we unify testing a bit.

devmotion:
Ah sorry, I forgot about the PR. OK, let's update the tests separately then!

PR #172, opened 2021-03-27
Fix and simplification of Logit

torfjelde:
Simplified Logit implementation a bit and fixed constructor. Currently constructors such as `Logit{1}(0, 1)` are not supported, and this PR fixes this. 

I just need to add some tests for the constructor and this should be good to go.

### DIFF
--- a/src/bijectors/logit.jl
+++ b/src/bijectors/logit.jl
@@ -30,6 +30,7 @@ Base.:(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
 (b::Logit{0})(x::Real) = _logit(x, b.a, b.b)
 (b::Logit)(x) = _logit.(x, b.a, b.b)
 (b::Logit)(x::AbstractArray{<:AbstractArray}) = map(b, x)
+_logit(x, a, b) = logit((x - a) / (b - a))
 
 (ib::Inverse{<:Logit{0}})(y::Real) = _ilogit(y, ib.orig.a, ib.orig.b)
 (ib::Inverse{<:Logit})(y) = _ilogit.(y, ib.orig.a, ib.orig.b)
--- a/src/bijectors/logit.jl
+++ b/src/bijectors/logit.jl
@@ -9,7 +9,7 @@ struct Logit{N, T<:Real} <: Bijector{N}
 end
 Logit(a::Real, b::Real) = Logit{0}(a, b)
 Logit(a::AbstractArray{<:Real, N}, b::AbstractArray{<:Real, N}) where {N} = Logit{N}(a, b)
-function Logit{N}(a, b)
+function Logit{N}(a, b) where {N}
     T = promote_type(typeof(a), typeof(b))
     Logit{N, T}(a, b)
 end
--- a/src/bijectors/logit.jl
+++ b/src/bijectors/logit.jl
@@ -27,17 +27,14 @@ up1(b::Logit{N, T}) where {N, T} = Logit{N + 1, T}(b.a, b.b)
 # For equality of Logit with Float64 fields to one with Duals
 Base.:(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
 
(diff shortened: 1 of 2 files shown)


src/bijectors/logit.jl
```diff
@@ -26,18 +28,12 @@ up1(b::Logit{N, T}) where {N, T} = Logit{N + 1, T}(b.a, b.b)
 Base.:(==)(b1::Logit, b2::Logit) = b1.a == b2.a && b1.b == b2.b
 
 (b::Logit{0})(x::Real) = _logit(x, b.a, b.b)
-(b::Logit{0})(x) = _logit.(x, b.a, b.b)
-(b::Logit{1})(x::AbstractVector) = _logit.(x, b.a, b.b)
-(b::Logit{1})(x::AbstractMatrix) = _logit.(x, b.a, b.b)
-(b::Logit{2})(x::AbstractMatrix) = _logit.(x, b.a, b.b)
-(b::Logit{2})(x::AbstractArray{<:AbstractMatrix}) = map(b, x)
-_logit(x, a, b) = logit((x - a) / (b - a))
```
devmotion, line 34: I guess `_logit` should not be deleted?

torfjelde: I mean, it depends on whether or not you want the code to work, which I wasn't planning on.


:upside_down_face: 

Nice catch, thanks!:)

src/bijectors/logit.jl
```diff
@@ -26,18 +28,12 @@ up1(b::Logit{N, T}) where {N, T} = Logit{N + 1, T}(b.a, b.b)
...
+(b::Logit)(x) = _logit.(x, b.a, b.b)
```
devmotion, line 31: Is the broadcasting overhead too large to remove the `(::Logit{0})(x::Real)` definition above? Speaking of overhead, if `x`, `b.a` and `b.b` are of the same dimension, one could also just use `map` which does not involve the huge broadcasting machinery.

torfjelde: I haven't checked recently, but I did go through the process of checking the overhead of broadcasting a year or so ago and it was sufficient for me to specialize on `Real`.

EDIT: Just checked and it doens't actually matter anymore (on Julia 1.5.3)! Seems like there isn't any overhead for `Real` these days, so it can prob be removed:)

torfjelde: Oh and regarding `map` vs. broadcasting; is there a particular reason why we want to avoid broadcasting?

devmotion: The main reason would be that broadcasting is quite complex and hence  often creates additional overhead and problems. E.g., it sometimes lead to type inference problems and the adjoints in Tracker and Zygote are based on ForwardDiff.

src/bijectors/logit.jl
```diff
@@ -7,9 +7,11 @@ struct Logit{N, T<:Real} <: Bijector{N}
     a::T
     b::T
 end
-function Logit(a, b)
+Logit(a::Real, b::Real) = Logit{0}(a, b)
+Logit(a::AbstractArray{<:Real, N}, b::AbstractArray{<:Real, N}) where {N} = Logit{N}(a, b)
+function Logit{N}(a, b)
```
devmotion, line 12:
```suggestion
function Logit{N}(a, b) where N
```

src/bijectors/logit.jl
torfjelde: Isn't ForwardDiff blazingly fast for broadcasting though, due to it's vectorized mode or w/e?

devmotion: Maybe, I don't know the details there. I just noticed that the implementation is quite non-trivial and still does not work in all cases (IIRC we even noticed some bugs in Bijectors that were caused by this implementation in Tracker, AFAIK at least these problems are fixed now).

devmotion:
Zygote test errors should be fixed by https://github.com/FluxML/Zygote.jl/pull/934.

torfjelde:
Seems like it's good to go?

devmotion (approved):
LGTM

PR #174, opened 2021-04-04
Fix type instabilities of SimplexBijector

devmotion:
Fixes type instabilities discovered in https://github.com/TuringLang/Turing.jl/issues/1276.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.8.16"
+version = "0.9.0"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -168,31 +168,31 @@ isdirichlet(::Distribution) = false
 function link(
     d::Dirichlet,
     x::AbstractVecOrMat{<:Real},
-    proj::Bool = true,
-)
+    ::Val{proj}=Val(true),
+) where {proj}
     return SimplexBijector{proj}()(x)
 end
 
 function link_jacobian(
     d::Dirichlet,
-    x::AbstractVector{T},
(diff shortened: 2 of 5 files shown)


torfjelde:
I'll review, but quick quesh first: is there a reason why you re-introduced the `Val`? I don't think that's necessary here; at least when I experimented, it wasn't.

devmotion:
I could not obtain type stability without it. IMO this seems reasonable since it is not possible to infer the type, or rather the type parameter `proj`, of `SimplexBijector` from a boolean value `proj`.

torfjelde:
For the case when `proj` is a runtime parameter you mean, or nah? It is inferrable if you do `SimplexBijector{true}()`.

devmotion:
Sure, you can hardcode it for the default version with `true` - but the three argument version, with `true` or `false`, is still type unstable.

devmotion:
BTW also with your suggestion in https://github.com/TuringLang/Turing.jl/issues/1276#issuecomment-813074909 `invlink(d, x)` is still type unstable. It seems you accidentally checked `link(d, x)` instead?

torfjelde:
Haha, yeah you're right! Sorry about that. I think I was just cycling through my command history to get the function call and seems like I was a bit too quick.

But regarding the 3-argument version. The reason why we replaced `Val` with `proj` was because when you construct the bijector in a different scope than `Main`, you'll have to hardcode `proj` to get type-stability no matter if we're using `Val` or not. And if you hardcode, even when not using `Val`, you should get type stability due to constant propogation.

Thouhgh in the light of the stuff we saw in the DPPL PR, I'm not so certain about that anymore:) And it seems like even something as simple as

```julia
f(d, x) = Bijectors.invlink(d, x, true)
```

fails :confused:

So I guess I'm in favour of adding back the `Val`.

torfjelde (approved):
LGTM!

But is this considered a breaking change? Should we bump the minor version or? (Also can you bump the version?:) )

devmotion:
Yes, IMO this is a breaking change.

Issue #156, opened 2021-01-07
Skellam, NormalInverseGaussian, and Pareto (partly) are broken with Zygote and SpecialFunctions >= 1


devmotion:
Currently, these distributions are broken due to https://github.com/FluxML/Zygote.jl/issues/873. As discussed in https://github.com/TuringLang/Bijectors.jl/pull/155, the corresponding tests are marked as broken. The issue will be fixed upstream.

See also: https://github.com/TuringLang/DistributionsAD.jl/issues/145

devmotion:
Copy from https://github.com/TuringLang/DistributionsAD.jl/issues/145#issuecomment-853794894:

> The main issue was fixed upstream, only some interesting issue due to `Zygote._zero` with `ChainRulesCore.NotImplemented` for matrices of Skellam variates is left: https://github.com/TuringLang/DistributionsAD.jl/pull/172

torfjelde:
So I guess this can be closed?

devmotion:
No, Skellam is still broken with Zygote in one test.

torfjelde:
Do you know which one? I just had to remove a bunch of `broken = (:zygote, )` to make tests pass in #177 so jus curious which one still remains.

devmotion:
Are you sure? I fixed these in https://github.com/TuringLang/Bijectors.jl/pull/182, I thought.

devmotion:
Ah, it seems you updated the tests before merging the master branch. I updated this issue when I fixed the tests in https://github.com/TuringLang/Bijectors.jl/pull/182, so the remaining error is still not fixed.

torfjelde:
Aaaah, my bad; thanks!

penelopeysm:
We no longer care about Zygote :(.

PR #182, opened 2021-06-03
Update to ChainRulesCore 0.10

devmotion:
Supercedes https://github.com/TuringLang/Bijectors.jl/pull/180 and https://github.com/TuringLang/Bijectors.jl/pull/181.


test/ad/distributions.jl
```diff
@@ -435,17 +433,21 @@
             # Skellam only fails in these tests with ReverseDiff
             # Ref: https://github.com/TuringLang/DistributionsAD.jl/issues/126
             # PoissonBinomial fails with Zygote
+            # Matrix case does not work with Skellam:
+            # https://github.com/TuringLang/DistributionsAD.jl/pull/172#issuecomment-853721493
             filldist_broken = if d.f(d.θ...) isa Skellam
-                (d.broken..., :ReverseDiff)
+                ((d.broken..., :ReverseDiff), (d.broken..., :Zygote, :ReverseDiff))
```
yebai, line 439: Is the redundant `:ReverseDiff` in `(d.broken..., :ReverseDiff)` and `(d.broken..., :Zygote, :ReverseDiff)` intended?

devmotion: Yes, ReverseDiff still doesn't work for both tested sizes (single sample + two samples) whereas Zygote only fails if there are multiple samples. Upstream changes in SpecialFunctions and Zygote fixed most of the issues in https://github.com/TuringLang/Bijectors.jl/issues/156 (and https://github.com/TuringLang/DistributionsAD.jl/issues/145) but this one is still present.

devmotion:
Test errors are caused by NNlib it seems - I don't understand why we end up with such an old version (0.7.9), IIRC the problem is fixed in recent versions of NNlib.

devmotion:
Ah OK, so 0.7.9 was the latest release without explicit ChainRules dependency: ChainRulesCore 0.10 is pulled in first and since no more recent version of NNlib is compatible with it (well, 0.7.21 is but it requires Julia >= 1.6) the old version with the bug is installed. I'll try to force NNlib 0.7.18 explicitly (latest release which supports ChainRulesCore 0.9 + Julia 1.3).

yebai:
Many thanks, @devmotion.

Issue #159, opened 2021-01-08
Do we really need isclosedform anymore?


torfjelde:
After https://github.com/TuringLang/Bijectors.jl/pull/158 it seems like even the inverse transformations using fixed-point iteration have such a high accuracy that `isclosedform` might not have much of a purpose anymore.

I don't feel like doing something about it already, but the issue is opened to serve as a reminder for possible future changes.

Issue #41, opened 2019-09-03
Rename `rv` from `forward(b::Bijector, x)` result


torfjelde:
At the moment, if you call `forward(b, x)` you get back a named tuple `(rv = b(x), logabsdetjac = logabsdetjac(b, x))`. This was introduced in the first PR for the new interface.

When I started out working on this, using `rv` made sense as the transform was always related to some `Distribution`. Now a `Bijector` can be used as a standalone transform to do more than just transform random variables, e.g. constrained-to-unconstrained transformations in optimization. Therefore I suggest we rename `rv`.

A couple of suggestions (in no particular order):
1. `y` since throughout the entire package it's a recurring theme where `x` is "un-transformed" and `y` denotes transformed. Also, in `forward`, `x` is the input.
2. `val` or `value`
3. `res` or `result`

Any suggestions?

xukai92:
I like 1 and 3. The reason I don't like 2 is because `val` or `value` doesn't indicate the value is transfomed or not.

willtebbutt:
I vote 1

torfjelde:
Personally I also like 1. One possible issue is that one can also call `forward(ib::Inversed{<:Bijector}, y)`. Should this also return `y = ...`? We should use the same key in `forward` for both inversed and not, imo.

willtebbutt:
Another alternative is `output`. I guess `y` isn't such a great name if you consider inverse transformations.

torfjelde:
> Another alternative is output. I guess y isn't such a great name if you consider inverse transformations.

True. I think I prefer 3 over `output` but not by much

xukai92:
Similarly, `ans` is a good choice. It's also Julia's default name holder for results.

```bash
➜  ~ julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.1.0 (2019-01-21)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |
```
```julia
julia> 1 + 1
2

julia> ans
2
```

trappmartin:
I prefer 3.

torfjelde:
A final call for votes: @yebai @cpfiffer @mohamed82008 @sharanry

yebai:
[...]

PR #183, opened 2021-06-06
Rewrite: removing dimensionality and allow non-bijective transformations

torfjelde:
This is a draft for a proper overhaul of Bijectors.jl

Goals with this PR are:
1. [ ] Remove the annoying dimensionality by being explict about when we're working with a collection of inputs vs. single input.
   - Also removes the restriction that input and output of bijectors should be the same size, type, etc.
2. [ ] Allow more than just bijectors (differentiable bijections with differentiable inverses). 
   - Releated issues: #58
3. [ ] Resolve some simpler but outstanding issues:
   - #41
   - #159
4. [ ] We have _a lot_ of hacks in this package to make AD work across the board. Some of these vanish due to (1), but I'm also taking this opportunity to try to check what we actually need and what we can get rid of due to improvements upstream.
5. [ ] Add support for mutating methods, e.g. `transform!`, `logabsdetjac!`.
6. [ ] Add a bunch of sane default implementations, e.g. mutating versions, batched versions, etc.

Also, this is likely to result end up including a lot of changes, so we _might_ end up splitting this into multiple PRs once it becomes more than a draft. But for now it's all here.


yebai:
Let's perhaps make a push to finish the work here. @torfjelde

docs/src/index.md
```diff
@@ -0,0 +1,144 @@
+# Bijectors.jl
```
torfjelde, line 1: @devmotion I'm curious what you think of this design:) You can ignore the code for now, the main thing is just whether or not you're happy with the idea of doing 

``` julia
forward(b, x) = forward_single(b, x)

# Non-batched version
forward_single(b, x) = ...

# Batched version
forward_multiple(b, x) = ...
```

We can then introduce a `Batch` type which generalizes `ColVecs` etc., and do

``` julia
forward(b, x::AbstractBatch) = forward_multiple(b, x)
```

My plan is to split this into two PRs:
1. Remove all "official" support for batched computation, thus always assuming that the input given represents a _single_ input (some bijectors might still support it, but there's not going to be a "official" support for it). In this PR there is no `forward_single`, etc., just `forward`.
2. Add support for batching and overload `forward_single` instead of `forward`.

_But_ ideally we'd fully adopt the ChangesOfVariables.jl interface, i.e. replace `forward` with `with_logabsdet_jacobian`. The issue here is that we of course can no longer do

``` julia
with_logabsdet_jacobian(b, x) = with_logabsdet_jacobian_single(b, x)
```

etc. which makes me want to keep `forward` _or_ ChangesOfVariables decides to take on a similar interface and encourage people to instead implement `with_logabsdet_jacobian_single`.

Thoughts?

devmotion: I think it will be a huge improvement if batching is declared more explicitly. Is there a specific reason why one needs `forward_single` and `forward_multiple` instead of just `forward(::MyBijector, x)` and `forward(::MyBijector, x::AbstractBatch)`? More functions means more entry points and hence more possible confusion for developers. If there's no dedicated `_single` function anymore, it would also work better with the CoV API, I assume?

torfjelde: The issue is method ambiguity :confused: If only `forward` or `with_logabsdet_jacobian` is the entry-point, then we cannot provide sane defaults, e.g. `forward(b, x::AbstractBatch)` without every other implementation of `forward` being _very_ explicit.

devmotion: Yeah this is annoying - but there are multiple different options I think. One would be e.g. to implement (possibly different) default implementations but not as `forward(...)` but some `forward_batch_style1(..)` (whatever) and then to "activate" it by defining `forward(b::MyBijector, x::AbstractBatch) = forward_batch_style1(b, x)` etc. I.e., batch support would be enabled manually but without much effort. And one could reduce the amount of code even more with some helper macro.

torfjelde: Yeah that works of course, but I think this will be easy to forget vs. always _just_ implemented `*_single` and batching is guaranteed to at least work.

Both approaches feel sub-optimal :confused:

torfjelde:
No this can be closed:)

PR #184, opened 2021-06-08
Fix bug in Stacked

torfjelde:
Before #177 we had very opinionated constructors that would ensure that if you were constructing a `Stacked` with a `Tuple` as the container for the bijectors, you'd also get a `Tuple` container for the ranges. After relaxing the requirements/assumptions made, it's now possible (but of course not encournaged) to have a combination of `Array` and `Tuple`, e.g. `Stacked((Exp(), ), [1:1, ])`. This will then dispatch to the impl that unrolls the map even though it shouldn't because we only specifiy the type of the first parameter in the current impl.

This PR fixes this bug by making it so we only dispatch to the `@generated` impl if both type-parameters are `Tuple`, and falls back to the non-generated impl if this is not the case.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.9.4"
+version = "0.9.5"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
--- a/src/bijectors/stacked.jl
+++ b/src/bijectors/stacked.jl
@@ -73,13 +73,13 @@ function _transform(x, rs::NTuple{1, UnitRange{Int}}, b::Bijector)
     @assert rs[1] == 1:length(x)
     return b(x)
 end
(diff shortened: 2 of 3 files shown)


torfjelde:
Could someone maybe have a look at this @devmotion @yebai @mohamed82008 ? It's a very small PR but it's holding up tests for PRs in Turing :confused:

devmotion:
> Stacked((Exp(), ), [1:1, ])

Shouldn't we handle this in a type-stable way as well? Both ranges and bijectors are concretely typed and inference shouldn't be a problem. And isn't usually the array of ranges concretely typed?

torfjelde:
> Shouldn't we handle this in a type-stable way as well? Both ranges and bijectors are concretely typed and inference shouldn't be a problem. And isn't usually the array of ranges concretely typed?

Sorry, I was just using terminology wrong. This will be type-stable, but it just won't use `@generated`, i.e. be unrolled.

devmotion (approved):

devmotion:
Ah, I see :+1:

PR #185, opened 2021-06-13
Fixed double double subscripts in Docs

storopoli:
This is not rendering right in Turing.ml Documentation. See here (https://turing.ml/dev/docs/library/bijectors/#Bijectors.compute_r-Tuple{AbstractVector{var%22#s111%22}%20where%20var%22#s111%22%3C:Real,%20Any,%20Any}). I've tried to corrected it but I don't know if the intent is to signal that α is positive constrained (and `y` is negative constrained). I would do something like $\alpha^{+}$ (or $y^{-}$).

Screen Shot 2021-06-13 at 10 27 11

### DIFF
--- a/src/bijectors/radial_layer.jl
+++ b/src/bijectors/radial_layer.jl
@@ -105,15 +105,15 @@ end
 
 Compute the unique solution ``r`` to the equation
 ```math
-\\|y_minus_z0\\|_2 = r \\left(1 + \\frac{α_plus_β_hat - α}{α + r}\\right)
+\\|y_minus_z0\\|_2 = r \\left(1 + \\frac{α_{plus_{β}_{hat}} - α}{α + r}\\right)
 ```
 subject to ``r ≥ 0`` and ``r ≠ α``.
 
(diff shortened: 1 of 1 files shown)


devmotion (commented):
No, it's not related to positivity or negativity, it's just what it says: the sum or difference of two variables. The names in the docstrings are the name of the Julia variables and I chose the descriptive names to make it clearer what's happening there. I'm a bit surprised this shows up on the webpage, it was intended mainly as an explanation of the math behind the implementation and I assumed it would be nicer to use docstring syntax. Maybe it's best to just remove it from the webpage?

storopoli:
Ah! Know I get it. The problem is that the `bijectors.md` for the documentation is kinda hard coded into the jekyll markdown. See here: https://github.com/TuringLang/Turing.jl/blob/master/docs/src/library/bijectors.md

devmotion:
IMO it should be removed completely and be replaced with a proper documentation in Bijectors.

storopoli:
I agree. Thanks for the input.

PR #186, opened 2021-06-13
`OrderedBijector`

torfjelde:
Something similar to `ordered` in Stan has been requested on several occasions, e.g. https://github.com/TuringLang/Turing.jl/discussions/1535.

This PR introduces `OrderedBijector` + `Bijectors.ordered(d::Distribution)` to address this.

I also took the liberty of moving the Tracker-compat stuff into a submodule, similar to the other AD-packages, and made overloads clearer by _not_ importing the functions. It was quite difficult to
figure out what was missing when I made it a submodule when everything was imported.


src/compat/tracker.jl
```diff
@@ -460,3 +469,25 @@ end
 
     return α, find_alpha_pullback
 end
+
+# `OrderedBijector`
+Bijectors._transform_ordered(y::Union{TrackedVector,TrackedMatrix}) = track(Bijectors._transform_ordered, y)
+@grad function Bijectors._transform_ordered(y::AbstractVecOrMat)
+    x, dx = ChainRulesCore.rrule(Bijectors._transform_ordered, data(y))
+    return x, (wrap_chainrules_output ∘ Base.tail ∘ dx)
+end
```
torfjelde, line 478: Do you know if this is "dangerous" in any way @devmotion ? Just seems a bit "too easy".

src/bijectors/ordered.jl
```diff
@@ -0,0 +1,60 @@
...
+
+    return x
+end
+
+function _transform_ordered(y::AbstractMatrix)
+    x = similar(y)
+    x[1, :] = y[1, :]
+    @inbounds for i = 2:size(x, 1)
+        x[i, :] = x[i - 1, :] + exp.(y[i, :])
+    end
```
devmotion, line 36: I guess it would be more efficient to operate on columns instead of rows but I assume it is required by the design of the matrix inputs? Also it seems quite allocation heavy.

devmotion (commented):
I know that Stan uses this transform but already when I read the Stan docs a while ago I wondered how this compares to other possible alternatives (e.g. with log1pexp or completely different approaches?).

src/bijectors/ordered.jl
torfjelde: > I guess it would be more efficient to operate on columns instead of rows but I assume it is required by the design of the matrix inputs? 

Yeah, it's required by the inputs :confused:

> Also it seems quite allocation heavy.

Want me to add some `view`? I didn't bother because I figured: `exp.(...)` will already allocation.

src/compat/tracker.jl
devmotion: Not sure, seems similar to what Zygote does. But the conversion of ChainRules types there is a bit tricky, so I assume also here it could be tricky to implement `wrap_chainrules_output` correctly (e.g. one should not drop NotImplemented and unthunk stuff). But as long as you only apply it to rules that you've defined it is fine I guess. For simple operations the possible overhead might be noticeable but I guess it doesn't matter for expensive operations.

(BTW we already use `rrule` in definitions for other ADs in DistributionsAD: https://github.com/TuringLang/DistributionsAD.jl/blob/fe2070012167b78c84a49733a0a64997e9533812/src/common.jl#L8-L13)

src/bijectors/ordered.jl
torfjelde: Btw, do you think it would be faster to do

```julia
@inbounds for j = 1:size(x, 2), i = 2:size(x, 1)
    x[i, j] = x[i - 1, j] + exp.(y[i, j])
end
```

?

devmotion: > I didn't bother because I figured: exp.(...) will already allocation.

`x[i - 1, :] + exp.(y[i, :])` will allocate an additional array, it seemed strange that one does not broadcast the whole expression and updates `x` in-place.

> Btw, do you think it would be faster to do?

I am not sure, best would be to benchmark it :slightly_smiling_face:

torfjelde:
> I know that Stan uses this transform but already when I read the Stan docs a while ago I wondered how this compares to other possible alternatives (e.g. with log1pexp or completely different approaches?).

No, I haven't considered alternative approaches. I'm all for it though, it's just that I think it's more important to at least have support for this, and _then_ we can experiment with different approaches (unless we already know one is better than the other).

devmotion:
Sure, at this stage I was mostly curious if it is known that another approach would be better.

src/bijectors/ordered.jl
```diff
@@ -0,0 +1,60 @@
...
+ordered(d::Distribution) = Bijectors.transformed(d, OrderedBijector())
+
+(b::OrderedBijector)(y::AbstractVecOrMat) = _transform_ordered(y)
+
+function _transform_ordered(y::AbstractVector)
+    x = similar(y)
+    x[1] = y[1]
+    @inbounds for i = 2:size(x, 1)
+        x[i] = x[i - 1] + exp(y[i])
+    end
```
devmotion, line 26: An alternative would be something like
```julia
x = map(exp, y)
x[begin] = y[begin]
cumsum!(x, x)
```
(or a `map` that directly avoids `exp` of the first element). Could be interesting to benchmark as well, in particular since `cumsum!` accepts a `dims` keyword argument.

src/bijectors/ordered.jl
torfjelde: > Btw, do you think it would be faster to do

Some quick checks tells me that yes this is the case. I'll just go with that then :+1: Also, guaranteed to have waaay fewer allocaitons (i.e. 1).

src/bijectors/ordered.jl
torfjelde: `begin` isn't supported on 1.3 though, right?

torfjelde: I'll give that a go :+1:

I also started doing `map(exp, y)` and `x[1] = y[1]` after, but this doesn't work in the _inverse_ since we're using `log`. BUT that's not a good reason not to do it in this case; I just tried doing it everywhere, failed and I just reverted the change without giving it too much thought (just noted why it failed) :upside_down_face:

torfjelde: Quick benchmark: for the batched-version, it's slower or the same as the for-loop approach :confused:

src/compat/tracker.jl
torfjelde: > Not sure, seems similar to what Zygote does. But the conversion of ChainRules types there is a bit tricky, so I assume also here it could be tricky to implement wrap_chainrules_output correctly (e.g. one should not drop NotImplemented and unthunk stuff). But as long as you only apply it to rules that you've defined it is fine I guess. For simple operations the possible overhead might be noticeable but I guess it doesn't matter for expensive operations.

I was looking at what Zygote was doing, and it doesn't seem like it would be incredibly difficult to make something like `@defer_chainrules` or something for Tracker.jl and ReverseDiff.jl, but for the reasons you mentioned I decided to not be overly ambitious right now.

And I was thinking the same: should be fine since in these cases.

> BTW we already use rrule in definitions for other ADs in DistributionsAD

Oh nice! Wasn't aware :+1:

torfjelde:
Happy with this now @devmotion ? Only thing left is version-bump.

src/chainrules.jl
```diff
@@ -6,3 +6,122 @@ ChainRulesCore.@scalar_rule(
     ),
     (x, - tanh(Ω + b) * x, x - 1),
 )
+
+# `OrderedBijector`
+function ChainRulesCore.rrule(::typeof(_transform_ordered), y::AbstractVector)
```
devmotion, line 11: All these rules should be tested with ChainRulesTestUtils (as done for other rules).

src/bijectors/ordered.jl
```diff
@@ -0,0 +1,78 @@
...
+        else
+            y[i, j] = log(x[i, j] - x[i - 1, j])
+        end
+    end
+
+    return y
+end
+
+logabsdetjac(b::OrderedBijector, x::AbstractVector) = sum(x[2:end])
+logabsdetjac(b::OrderedBijector, x::AbstractMatrix) = vec(sum(x[2:end, :]; dims = 1))
```
devmotion, line 78: Maybe use views?

src/chainrules.jl
```diff
@@ -6,3 +6,122 @@ ChainRulesCore.@scalar_rule(
...
+
+function ChainRulesCore.rrule(::typeof(_transform_inverse_ordered), x::AbstractVector)
+    r = similar(x)
+    @inbounds for i = 1:length(r)
+        if i == 1
+            r[i] = 1
+        else
+            r[i] = x[i] - x[i - 1]
+        end
+    end
```
devmotion, line 65: Why not just `r = diff(x)` and handle the first case separately below?

src/chainrules.jl
```diff
@@ -6,3 +6,122 @@ ChainRulesCore.@scalar_rule(
...
+
+    function _transform_inverse_ordered_adjoint(Δ)
+        Δ_new = similar(x)
+        @assert length(Δ_new) == length(Δ)
+
+        n = length(Δ_new)
+        @inbounds for j = 1:n - 1
+            Δ_new[j] = (Δ[j] / r[j]) - (Δ[j + 1] / r[j + 1])
```
devmotion, line 73: `Δ[j] / r[j]` is already computed in the previous step. Maybe save it?

src/chainrules.jl
```diff
@@ -6,3 +6,122 @@ ChainRulesCore.@scalar_rule(
...
+
+function ChainRulesCore.rrule(::typeof(_transform_inverse_ordered), x::AbstractMatrix)
+    r = similar(x)
+    @inbounds for j = 1:size(x, 2), i = 1:size(x, 1)
+        if i == 1
+            r[i, j] = 1
+        else
+            r[i, j] = x[i, j] - x[i - 1, j]
+        end
+    end
```
devmotion, line 97: Same here, maybe just `r = diff(x; dims=1)`?

src/chainrules.jl
```diff
@@ -6,3 +6,122 @@ ChainRulesCore.@scalar_rule(
...
+
+    function _transform_inverse_ordered_adjoint(Δ)
+        Δ_new = similar(x)
+        n = length(Δ)
+        @assert n == size(Δ_new, 1)
+
+        @inbounds for j = 1:size(Δ_new, 2), i = 1:n - 1
+            Δ_new[i, j] = (Δ[i, j] / r[i, j]) - (Δ[i + 1, j] / r[i + 1, j])
```
devmotion, line 105: Same as above, maybe save the intermediate results for the subsequent step?

src/chainrules.jl
torfjelde: I did this because it makes the code for the adjoint a bit "simpler" by allowing us to use the same index for both `Δ` and `r`, while if we do something like `diff` we need to adjust for this. It just makes it slightly more confusing to read IMO. Plus we need to handle the case of `i = 1` in the computation which also reads a bit worse IMO.

src/bijectors/ordered.jl
torfjelde: Sure :+1:

src/chainrules.jl
torfjelde: True, but it requires a couple of more variables. A quick benchmark shows that it doesn't do anything for perf :confused: 
Could this be because of cache prefetching, i.e. the adjacent variables are already on the CPU and so it doesn't matter.

src/chainrules.jl
torfjelde: Same as above: IMO complicates the indexing and it doesn't do anything for perf (in a quick benchmark).

src/chainrules.jl
torfjelde: Where is the other rules we test? Just need something to go on.

torfjelde:
There's a bug with DistributionsAD.jl for ReverseDiff.jl again :confused: 

Can we move the `DistSpec` tests to DistributionsAD.jl? 

Now that `link`, `invlink`, and `logpdf_with_trans` are mere wrappers around the bijectors, if `logpdf` is AD-compatible for a distribution `d`, then Bijectors.jl ensuring AD-compat for its transform and `logabsdetjac` is sufficient to ensure `logpdf_with_trans` being AD-compatible. That is, Bijectors.jl should be responsible for AD-compat of the bijectors and nothing more.

And the times where PRs to Bijectors.jl are delayed because of DistributionsAD.jl are a bit too many, hence moving this over to DistributionsAD would be helpful in that regard too.

torfjelde:
Good to go?

src/bijectors/ordered.jl
```diff
@@ -0,0 +1,78 @@
...
+  - Note that this transformation and its inverse are the _opposite_ of in this reference.
+"""
+struct OrderedBijector <: Bijector{1} end
+
+"""
+    ordered(d::Distribution)
+
+Returns a `Distribution` whose domain is now ordered vectors.
+"""
+ordered(d::Distribution) = Bijectors.transformed(d, OrderedBijector())
```
devmotion, line 17: Should this be restricted to
```suggestion
ordered(d::ContinuousMultivariateDistribution) = Bijectors.transformed(d, OrderedBijector())
```
?

src/chainrules.jl
devmotion: In ad/chainrules.jl IIRC.

src/bijectors/ordered.jl
```diff
@@ -0,0 +1,78 @@
...
+## See also
+- [Stan's documentation](https://mc-stan.org/docs/2_27/reference-manual/ordered-vector.html)
+  - Note that this transformation and its inverse are the _opposite_ of in this reference.
+"""
+struct OrderedBijector <: Bijector{1} end
+
+"""
+    ordered(d::Distribution)
+
+Returns a `Distribution` whose domain is now ordered vectors.
```
devmotion, line 15:
```suggestion
Return a `Distribution` whose support are ordered vectors, i.e., vectors with increasingly ordered elements.
```

src/bijectors/ordered.jl
torfjelde: Yes, that's a good point :+1: Currently it would work for any `MultivariateDistribution` (`transformed` will fail for non-multivariate due to the dimensionality-thingy).

torfjelde:
Should be good now :+1:

yebai (approved):

PR #187, opened 2021-06-13
Default `bijector` implementation for `TransformedDistribution`

torfjelde:
Currently we're missing an implementation for `bijector(td::TransformedDistribution)` which in turn means that you can't, without a bit of work, use transformed distributions in, say, a DPPL-model.

This is stupid because it's trivial to construct this bijector: just invert `td.transform`, bringing us back the original domain of `td.dist`, and then compose this with the bijector for `td.dist`, which then takes us to real space.

EDIT: Btw, I have the right to call it stupid because I'm the one who didn't do it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.9.5"
+version = "0.9.6"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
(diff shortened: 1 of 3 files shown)


devmotion (approved):

devmotion:
Maybe add a test?

torfjelde:
Yeaaah I was thinking about it, but then when I was about too I went "wait, but I want the test to the implementation..." :sweat_smile: So I don't think there's much point tbh.

devmotion:
Yeah, I guess the main point would be to ensure that it exists and it returns correct values for a random input for transformed distributions with specific transforms (e.g. identity and exponentiation).

torfjelde:
Added a couple :+1:

devmotion (approved):

PR #192, opened 2021-08-05
Fix for bug in `forward` of `Stacked`

torfjelde:
Fixes the bug described in https://github.com/TuringLang/Turing.jl/issues/1674.


yebai:
Can we update the PR and merge it if it is still needed? @torfjelde

yebai (approved):
thanks, @torfjelde!

PR #212, opened 2021-12-10
Use ChangesOfVariables and InverseFunctions

### LINKED ISSUE #199: Support InverseFunctions.jl and ChangesOfVariables.jl
### To do

* [x] Create lightweight packages that define an interface for inverse functions and log-abs-det-jacobians
* [ ] Use `InverseFunctions` and `ChangesOfVariables` in Bijectors (and other packages ...)

### Context

Bijectors is really neat, but it's also kind of a heavy package with many dependencies (necessarily).

I have some variable transformation stuff in BAT.jl that I want to split out into a separate package, and I've recently been experimenting with some [normalizing flow trafos](https://github.com/bat/EuclidianNormalizingFlows.jl) - I'd like to make that kind of stuff compatible with Bijectors, but it would be to heavy as a dependency.

Would you guys (CC @devmotion) be interested in setting up a lightweight variable-trafo-interface package? Maybe just something like

```julia
function with_logabsdet_jacobian end

function with_logabsdet_jacobian(f::Base.ComposedFunction)(x)
    y_inner, ladj_inner = with_logabsdet_jacobian(f.inner, x)
    y, ladj_outer = with_logabsdet_jacobian(f.outer, y_inner)
    (y, ladj_inner + ladj_outer)
end


function inverse end

inverse(trafo::Base.ComposedFunction) = Base.ComposedFunction(inverse(f.inner), inverse(f.outer))

inverse(f) = inv(f)  # Until JuliaLang/julia#42421 is decided one way of the other
```

`with_logabsdet_jacobian` would be the equivalent of `Bijectors.forward`, while `f(x)` would be used to just run the transform without calculating LADJ. So the behavior would be

```
mytrafo(x) == y

with_logabsdet_jacobian(mytrafo, x) == (y, ladj)

inverse(trafo2∘ trafo1)(x) == inverse(trafo1)(inverse(trafo2)(x))
```

No autodiff or anything, that we'd leave to implementations like `Bijectors.Bijector`.

Long-term, it would be nice to get rid of `inverse` in favor of `inv`, if we can get `inv(::ComposedFunction)` into Base (just opened an issue in that regard: JuliaLang/julia#42421).

I'd volunteer to prototype a package (would be quite tiny, basically just the code above) if there's interest.

**Edit: Replaced special argument type `WithLADJ` in proposal with function `with_logabsdet_jacobian`, as suggested by @devmotion**

oschulz:
This PR implements the changes discussed in #199, by adding support for [JuliaMath/ChangesOfVariables](https://github.com/JuliaMath/ChangesOfVariables.jl) and [JuliaMath/InverseFunctions](https://github.com/JuliaMath/InverseFunctions.jl). Both are very lightweight, low-dependency, low-bias packages designed to enable composability of packages that provide/implement or use variable transformation capabilities.

Specifically, this PR adds support for `ChangesOfVariables.with_logabsdet_jacobian(::AbstractBijector, ::Any)` (a direct equivalent of - and indeed modeled after - the current `Bijectors.forward(::AbstractBijector, ::Any)`) and `InverseFunctions.inverse` (a direct equivalent of the current `Base.inv(::AbstractBijector)`).

The following registered packages directly depend on Bijectors, currently: DifferentialEvolutionMCMC DynamicPPL Turing TuringModels Transits ParameterHandling MeasureTheory AdvancedVI AIBECS Soss 

None of those dependent packages define subtypes of `AbstractBijector`, specialize `Bijectors.forward` or seem to specialize `Base.inv` (hope I didn't overlook any). So it seems resonable to deprecate `Bijectors.forward(::AbstractBijector, ::Any)` and `Base.inv(::AbstractBijector)` directly and replace all use of them inside of Bijectors.jl with `ChangesOfVariables.with_logabsdet_jacobian` and `InverseFunctions.inverse`.

The return type of `with_logabsdet_jacobian` is slightly different from `forward` though, it returns a `Tuple` instead of a `NamedTuple{(:rv, :logabsdetjac)}`. It seems that the only package that uses these fields is `MeasureTheory`, in a single place (not at all anymore on the current master branch, it seems). It's handled in the deprecation of `forward` in this PR.

Closes #199.

CC @torfjelde, @devmotion, @willtebbutt, @cscherrer


yebai:
Cc @phipsgabler

devmotion:
@oschulz it seems you forgot to push some changes?

oschulz:
> @oschulz it seems you forgot to push some changes?

Was still working on them. :-) Changes are pushed now.

oschulz:
Don't run workflow yet, still fixing tests locally.

src/Bijectors.jl
```diff
@@ -249,6 +252,11 @@ include("utils.jl")
 include("interface.jl")
 include("chainrules.jl")
 
+Base.@deprecate forward(b::AbstractBijector, x) with_logabsdet_jacobian(b, x)
```
devmotion, line 255: AFAICT `forward` returns a NamedTuple whereas `with_logabsdet_jacobian` returns a Tuple? So this deprecation seems to be incorrect (and hence introduce an unintended breaking change - otherwise one could just remove `forward` directly). The following should work:
```suggestion
Base.@deprecate forward(b::AbstractBijector, x) NamedTuple{(:rv,:logabsdetjac)}(with_logabsdet_jacobian(b, x))
```

src/Bijectors.jl
```diff
@@ -249,6 +252,11 @@ include("utils.jl")
...
+
+import Base.inv
```
devmotion, line 257: A bit annoying and slightly dangerous to import `Base.inv` - but IIRC I ran into the same problem with `@deprecate` at some point :cry:

devmotion (commented):
I didn't check the PR in detail since it's marked as WIP. However, I noticed that the deprecation of `forward` seems to be incorrect since it introduces a breaking change. IMO it would be nice if it is possible to not break the external API in this PR:
- It would make the PR much simpler since it would not have to update the tests (which also more clearly indicates if/how breaking the PR is)
- It would not break downstream packages

oschulz:
Ready for review and CI (could you trigger the workflow, @devmotion ?)

I have one local test failure with Julia v1.7 in "test/transform.jl:151" (section with a comment "This should fail at the minute") but I get the same test failure with the current master branch, so it seems unrelated.

src/Bijectors.jl
oschulz: Oh, neat!

src/Bijectors.jl
oschulz: Yes, that behavior of `@deprecate` is annoying ... I replaced it with a manual version.

oschulz: Using

```julia
@noinline function Base.inv(b::AbstractBijector)
    Base.depwarn("`Base.inv(b::AbstractBijector)` is deprecated, use `InverseFunctions.inverse(b)` instead.", :(Base.inv))
    inverse(b)
end
```

now.

README.md
```diff
@@ -221,18 +221,18 @@ true
 which is always the case for a differentiable bijection with differentiable inverse. Therefore if you want to compute `logabsdetjac(b⁻¹, y)` and we know that `logabsdetjac(b, b⁻¹(y))` is actually more efficient, we'll return `-logabsdetjac(b, b⁻¹(y))` instead. For some bijectors it might be easy to compute, say, the forward pass `b(x)`, but expensive to compute `b⁻¹(y)`. Because of this you might want to avoid doing anything "backwards", i.e. using `b⁻¹`. This is where `forward` comes to good use:
```
devmotion, line 221:
```suggestion
which is always the case for a differentiable bijection with differentiable inverse. Therefore if you want to compute `logabsdetjac(b⁻¹, y)` and we know that `logabsdetjac(b, b⁻¹(y))` is actually more efficient, we'll return `-logabsdetjac(b, b⁻¹(y))` instead. For some bijectors it might be easy to compute, say, the forward pass `b(x)`, but expensive to compute `b⁻¹(y)`. Because of this you might want to avoid doing anything "backwards", i.e. using `b⁻¹`. This is where `with_logabsdet_jacobian` comes to good use:
```

README.md
```diff
@@ -221,18 +221,18 @@ true
...
+julia> with_logabsdet_jacobian(b, x)
+(-0.5369949942509267, 1.4575353795716655)
 ```
 
 Similarily
 
 ```julia
-julia> forward(inv(b), y)
-(rv = 0.3688868996596376, logabsdetjac = -1.4575353795716655)
+julia> forward(inverse(b), y)
```
devmotion, line 231:
```suggestion
julia> with_logabsdet_jacobian(inverse(b), y)
```

README.md
```diff
@@ -201,7 +201,7 @@ julia> logpdf_forward(td, x)
 
 #### `logabsdetjac` and `forward`
```
devmotion, line 202:
```suggestion
#### `logabsdetjac` and `with_logabsdet_jacobian`
```

README.md
```diff
@@ -713,9 +714,9 @@ The distribution interface consists of:
 #### Methods
 The following methods are implemented by all subtypes of `Bijector`, this also includes bijectors such as `Composed`.
 - `(b::Bijector)(x)`: implements the transform of the `Bijector`
-- `inv(b::Bijector)`: returns the inverse of `b`, i.e. `ib::Bijector` s.t. `(ib ∘ b)(x) ≈ x`. In most cases this is `Inverse{<:Bijector}`.
+- `inverse(b::Bijector)`: returns the inverse of `b`, i.e. `ib::Bijector` s.t. `(ib ∘ b)(x) ≈ x`. In most cases this is `Inverse{<:Bijector}`.
 - `logabsdetjac(b::Bijector, x)`: computes log(abs(det(jacobian(b, x)))).
-- `forward(b::Bijector, x)`: returns named tuple `(rv=b(x), logabsdetjac=logabsdetjac(b, x))` in the most efficient manner.
+- `with_logabsdet_jacobian(b::Bijector, x)`: returns named tuple `(b(x), logabsdetjac(b, x))` in the most efficient manner.
```
devmotion, line 719:
```suggestion
- `with_logabsdet_jacobian(b::Bijector, x)`: returns the tuple `(b(x), logabsdetjac(b, x))` in the most efficient manner.
```

src/Bijectors.jl
```diff
@@ -35,6 +35,9 @@ using MappedArrays
 using Base.Iterators: drop
 using LinearAlgebra: AbstractTriangular
 
+import ChangesOfVariables: with_logabsdet_jacobian
+import InverseFunctions: inverse
```
devmotion, line 39: Would it be possibly to not import these functions but qualify them when they are implemented? E.g., we could use here
```suggestion
using ChangesOfVariables: ChangesOfVariables, with_logabsdet_jacobian
using InverseFunctions: InverseFunctions, inverse
```
Then one could still call `with_logabsdet_jacobian` and `inverse` without qualifications but has to use `InverseFunctions.inverse` when implementing them.

src/bijectors/composed.jl
```diff
@@ -179,8 +179,8 @@ function logabsdetjac(cb::Composed, x)
     y, logjac = forward(cb.ts[1], x)
```
devmotion, line 179:
```suggestion
    y, logjac = with_logabsdet_jacobian(cb.ts[1], x)
```

src/bijectors/composed.jl
```diff
@@ -179,8 +179,8 @@ function logabsdetjac(cb::Composed, x)
...
     for i = 2:length(cb.ts)
         res = forward(cb.ts[i], y)
```
devmotion, line 181:
```suggestion
        y, res_logjac = with_logabsdet_jacobian(cb.ts[i], y)
```

src/bijectors/composed.jl
```diff
@@ -179,8 +179,8 @@ function logabsdetjac(cb::Composed, x)
...
-        y = res.rv
-        logjac += res.logabsdetjac
+        y = res[1]
+        logjac += res[2]
```
devmotion, line 183:
```suggestion
        logjac += res_logjac
```

src/bijectors/composed.jl
```diff
@@ -195,8 +195,8 @@ end
     for i = 2:N - 1
         temp = gensym(:res)
         push!(expr.args, :($temp = forward(cb.ts[$i], y)))
```
devmotion, line 197: Also replace the `forward` above (can't make a suggestion there):
```suggestion
        temp = gensym(:res_logjac)
        push!(expr.args, :(y, $temp = with_logabsdet_jacobian(cb.ts[$i], y)))
```

src/bijectors/composed.jl
```diff
@@ -195,8 +195,8 @@ end
...
-        push!(expr.args, :(y = $temp.rv))
-        push!(expr.args, :(logjac += $temp.logabsdetjac))
+        push!(expr.args, :(y = $temp[1]))
+        push!(expr.args, :(logjac += $temp[2]))
```
devmotion, line 199:
```suggestion
        push!(expr.args, :(logjac += $temp))
```

src/bijectors/composed.jl
```diff
@@ -212,10 +212,10 @@ function forward(cb::Composed, x)
     
     for t in cb.ts[2:end]
         res = forward(t, rv)
```
devmotion, line 214: Same here, there's another `forward` above:
```suggestion
        rv, res_logjac = with_logabsdet_jacobian(t, rv)
```

src/bijectors/composed.jl
```diff
@@ -212,10 +212,10 @@ function forward(cb::Composed, x)
...
-        rv = res.rv
-        logjac = res.logabsdetjac + logjac
+        rv = res[1]
+        logjac = res[2] + logjac
```
devmotion, line 216:
```suggestion
        logjac += res_logjac
```

src/bijectors/composed.jl
```diff
@@ -225,10 +225,10 @@ end
     for i = 2:length(T.parameters)
         temp = gensym(:temp)
         push!(expr.args, :($temp = forward(cb.ts[$i], y)))
```
devmotion, line 227: Again, another `forward` call above and here:
```suggestion
        temp = gensym(:res_logjac)
        push!(expr.args, :(y, $temp = with_logabsdet_jacobian(cb.ts[$i], y)))
```

src/bijectors/composed.jl
```diff
@@ -225,10 +225,10 @@ end
...
-        push!(expr.args, :(y = $temp.rv))
-        push!(expr.args, :(logjac += $temp.logabsdetjac))
+        push!(expr.args, :(y = $temp[1]))
+        push!(expr.args, :(logjac += $temp[2]))
```
devmotion, line 229:
```suggestion
        push!(expr.args, :(logjac += $temp))
```

src/bijectors/named_bijector.jl
```diff
@@ -125,8 +125,8 @@ function logabsdetjac(cb::NamedComposition, x)
     y, logjac = forward(cb.bs[1], x)
     for i = 2:length(cb.bs)
         res = forward(cb.bs[i], y)
```
devmotion, line 127:
```suggestion
    y, logjac = with_logabsdet_jacobian(cb.bs[1], x)
    for i = 2:length(cb.bs)
        y, res_logjac = with_logabsdet_jacobian(cb.bs[i], y)
```

src/bijectors/named_bijector.jl
```diff
@@ -125,8 +125,8 @@ function logabsdetjac(cb::NamedComposition, x)
...
-        y = res.rv
-        logjac += res.logabsdetjac
+        y = res[1]
+        logjac += res[2]
```
devmotion, line 129:
```suggestion
        logjac += res_logjac
```

src/bijectors/named_bijector.jl
```diff
@@ -141,8 +141,8 @@ end
     for i = 2:N - 1
         temp = gensym(:res)
         push!(expr.args, :($temp = forward(cb.bs[$i], y)))
```
devmotion, line 143: And another `forward` above:
```suggestion
        temp = gensym(:res_logjac)
        push!(expr.args, :(y, $temp = with_logabsdet_jacobian(cb.bs[$i], y)))
```

src/bijectors/named_bijector.jl
```diff
@@ -141,8 +141,8 @@ end
...
-        push!(expr.args, :(y = $temp.rv))
-        push!(expr.args, :(logjac += $temp.logabsdetjac))
+        push!(expr.args, :(y = $temp[1]))
+        push!(expr.args, :(logjac += $temp[2]))
```
devmotion, line 145:
```suggestion
        push!(expr.args, :(logjac += $temp))
```

src/bijectors/named_bijector.jl
```diff
@@ -158,10 +158,10 @@ function forward(cb::NamedComposition, x)
     
     for t in cb.bs[2:end]
         res = forward(t, rv)
```
devmotion, line 160: Another one above and here:
```suggestion
        rv, res_logjac = with_logabsdet_jacobian(t, rv)
```

src/bijectors/named_bijector.jl
```diff
@@ -158,10 +158,10 @@ function forward(cb::NamedComposition, x)
...
-        rv = res.rv
-        logjac = res.logabsdetjac + logjac
+        rv = res[1]
+        logjac = res[2] + logjac
```
devmotion, line 162:
```suggestion
        logjac += res_logjac
```

src/bijectors/named_bijector.jl
```diff
@@ -171,10 +171,10 @@ end
     for i = 2:length(T.parameters)
         temp = gensym(:temp)
         push!(expr.args, :($temp = forward(cb.bs[$i], y)))
```
devmotion, line 173: Another one above and here:
```suggestion
        temp = gensym(:res_logjac)
        push!(expr.args, :(y, $temp = with_logabsdet_jacobian(cb.bs[$i], y)))
```

src/bijectors/named_bijector.jl
```diff
@@ -171,10 +171,10 @@ end
...
-        push!(expr.args, :(y = $temp.rv))
-        push!(expr.args, :(logjac += $temp.logabsdetjac))
+        push!(expr.args, :(y = $temp[1]))
+        push!(expr.args, :(logjac += $temp[2]))
```
devmotion, line 175:
```suggestion
        push!(expr.args, :(logjac += $temp))
```

src/bijectors/planar_layer.jl
```diff
@@ -175,5 +175,5 @@ function find_alpha(wt_y::T, wt_u_hat::T, b::T) where {T<:Real}
     return α0
 end
 
-logabsdetjac(flow::PlanarLayer, x) = forward(flow, x).logabsdetjac
+logabsdetjac(flow::PlanarLayer, x) = forward(flow, x)[2]
```
devmotion, line 178:
```suggestion
logabsdetjac(flow::PlanarLayer, x) = last(with_logabsdet_jacobian(flow, x))
```

src/bijectors/normalise.jl
```diff
@@ -76,28 +76,28 @@ function forward(bn::InvertibleBatchNorm, x)
     logabsdetjac = (
         fill(sum(logs - log.(v .+ bn.eps) / 2), size(x, dims))
     )
-    return (rv=rv, logabsdetjac=logabsdetjac)
+    return (rv, logabsdetjac)
 end
 
-logabsdetjac(bn::InvertibleBatchNorm, x) = forward(bn, x).logabsdetjac
+logabsdetjac(bn::InvertibleBatchNorm, x) = with_logabsdet_jacobian(bn, x)[2]
```
devmotion, line 82: I usually prefer `first` and `last` but I guess it's a matter of taste since constant propagation should handle it :shrug: 
```suggestion
logabsdetjac(bn::InvertibleBatchNorm, x) = last(with_logabsdet_jacobian(bn, x))
```

src/bijectors/normalise.jl
```diff
@@ -76,28 +76,28 @@ function forward(bn::InvertibleBatchNorm, x)
...
 
-(bn::InvertibleBatchNorm)(x) = forward(bn, x).rv
+(bn::InvertibleBatchNorm)(x) = with_logabsdet_jacobian(bn, x)[1]
```
devmotion, line 84:
```suggestion
(bn::InvertibleBatchNorm)(x) = first(with_logabsdet_jacobian(bn, x))
```

src/bijectors/normalise.jl
```diff
@@ -76,28 +76,28 @@ function forward(bn::InvertibleBatchNorm, x)
...
     m = reshape(bn.m, as...)
     v = reshape(bn.v, as...)
 
     x = (y .- b) ./ s .* sqrt.(v .+ bn.eps) .+ m
-    return (rv=x, logabsdetjac=-logabsdetjac(bn, x))
+    return (x, -logabsdetjac(bn, x))
 end
 
-(bn::Inverse{<:InvertibleBatchNorm})(y) = forward(bn, y).rv
+(bn::Inverse{<:InvertibleBatchNorm})(y) = with_logabsdet_jacobian(bn, y)[1]
```
devmotion, line 100:
```suggestion
(bn::Inverse{<:InvertibleBatchNorm})(y) = first(with_logabsdet_jacobian(bn, y))
```

src/bijectors/radial_layer.jl
```diff
@@ -123,4 +123,4 @@ function compute_r(y_minus_z0::AbstractVector{<:Real}, α, α_plus_β_hat)
     return r
 end
 
-logabsdetjac(flow::RadialLayer, x::AbstractVecOrMat) = forward(flow, x).logabsdetjac
+logabsdetjac(flow::RadialLayer, x::AbstractVecOrMat) = forward(flow, x)[2]
```
devmotion, line 126:
```suggestion
logabsdetjac(flow::RadialLayer, x::AbstractVecOrMat) = last(with_logabsdet_jacobian(flow, x))
```

src/interface.jl
```diff
@@ -56,7 +56,7 @@ requires an iterative procedure to evaluate.
 isclosedform(b::Bijector) = true
 
 """
-    inv(b::Bijector)
+inverse(b::Bijector)
```
devmotion, line 59:
```suggestion
    inverse(b::Bijector)
```

src/transformed_distribution.jl
```diff
@@ -85,43 +85,43 @@ Base.length(td::Transformed) = length(td.dist)
 Base.size(td::Transformed) = size(td.dist)
 
 function logpdf(td::UnivariateTransformed, y::Real)
-    res = forward(inv(td.transform), y)
-    return logpdf(td.dist, res.rv) + res.logabsdetjac
+    res = forward(inverse(td.transform), y)
+    return logpdf(td.dist, res[1]) + res[2]
```
devmotion, line 89:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return logpdf(td.dist, x) + logjac
```

src/transformed_distribution.jl
```diff
@@ -85,43 +85,43 @@ Base.length(td::Transformed) = length(td.dist)
...
 end
 
 # TODO: implement more efficiently for flows in the case of `Matrix`
 function logpdf(td::MvTransformed, y::AbstractMatrix{<:Real})
     # batch-implementation for multivariate
-    res = forward(inv(td.transform), y)
-    return logpdf(td.dist, res.rv) + res.logabsdetjac
+    res = forward(inverse(td.transform), y)
+    return logpdf(td.dist, res[1]) + res[2]
```
devmotion, line 96:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return logpdf(td.dist, x) + logjac
```

src/transformed_distribution.jl
```diff
@@ -85,43 +85,43 @@ Base.length(td::Transformed) = length(td.dist)
...
 end
 
 function logpdf(td::MvTransformed{<:Dirichlet}, y::AbstractMatrix{<:Real})
     T = eltype(y)
     ϵ = _eps(T)
 
-    res = forward(inv(td.transform), y)
-    return logpdf(td.dist, mappedarray(x->x+ϵ, res.rv)) + res.logabsdetjac
+    res = forward(inverse(td.transform), y)
+    return logpdf(td.dist, mappedarray(x->x+ϵ, res[1])) + res[2]
```
devmotion, line 104:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return logpdf(td.dist, mappedarray(x->x+ϵ, x)) + logjac
```

src/transformed_distribution.jl
```diff
@@ -85,43 +85,43 @@ Base.length(td::Transformed) = length(td.dist)
...
 end
 
 function _logpdf(td::MvTransformed, y::AbstractVector{<:Real})
-    res = forward(inv(td.transform), y)
-    return logpdf(td.dist, res.rv) + res.logabsdetjac
+    res = forward(inverse(td.transform), y)
+    return logpdf(td.dist, res[1]) + res[2]
```
devmotion, line 109:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return logpdf(td.dist, x) + logjac
```

src/transformed_distribution.jl
```diff
@@ -85,43 +85,43 @@ Base.length(td::Transformed) = length(td.dist)
...
 end
 
 function _logpdf(td::MvTransformed{<:Dirichlet}, y::AbstractVector{<:Real})
     T = eltype(y)
     ϵ = _eps(T)
 
-    res = forward(inv(td.transform), y)
-    return logpdf(td.dist, mappedarray(x->x+ϵ, res.rv)) + res.logabsdetjac
+    res = forward(inverse(td.transform), y)
+    return logpdf(td.dist, mappedarray(x->x+ϵ, res[1])) + res[2]
```
devmotion, line 117:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return logpdf(td.dist, mappedarray(x->x+ϵ, x)) + logjac
```

src/transformed_distribution.jl
```diff
@@ -85,43 +85,43 @@ Base.length(td::Transformed) = length(td.dist)
...
 end
 
 # TODO: should eventually drop using `logpdf_with_trans` and replace with
-# res = forward(inv(td.transform), y)
-# logpdf(td.dist, res.rv) .- res.logabsdetjac
+# res = forward(inverse(td.transform), y)
+# logpdf(td.dist, res[1]) .- res[2]
```
devmotion, line 122:
```suggestion
# x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
# logpdf(td.dist, x) .- logjac
```

src/transformed_distribution.jl
```diff
@@ -163,34 +163,34 @@ Makes use of the `forward` method to potentially re-use computation
 and returns a tuple `(logpdf, logabsdetjac)`.
 """
 function logpdf_with_jac(td::UnivariateTransformed, y::Real)
-    res = forward(inv(td.transform), y)
-    return (logpdf(td.dist, res.rv) + res.logabsdetjac, res.logabsdetjac)
+    res = forward(inverse(td.transform), y)
+    return (logpdf(td.dist, res[1]) + res[2], res[2])
```
devmotion, line 167:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return (logpdf(td.dist, x) + logjac, logjac)
```

src/transformed_distribution.jl
```diff
@@ -163,34 +163,34 @@ Makes use of the `forward` method to potentially re-use computation
...
 end
 
 # TODO: implement more efficiently for flows in the case of `Matrix`
 function logpdf_with_jac(td::MvTransformed, y::AbstractVector{<:Real})
-    res = forward(inv(td.transform), y)
-    return (logpdf(td.dist, res.rv) + res.logabsdetjac, res.logabsdetjac)
+    res = forward(inverse(td.transform), y)
+    return (logpdf(td.dist, res[1]) + res[2], res[2])
```
devmotion, line 173:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return (logpdf(td.dist, x) + logjac, logjac)
```

src/transformed_distribution.jl
```diff
@@ -163,34 +163,34 @@ Makes use of the `forward` method to potentially re-use computation
...
 end
 
 function logpdf_with_jac(td::MvTransformed, y::AbstractMatrix{<:Real})
-    res = forward(inv(td.transform), y)
-    return (logpdf(td.dist, res.rv) + res.logabsdetjac, res.logabsdetjac)
+    res = forward(inverse(td.transform), y)
+    return (logpdf(td.dist, res[1]) + res[2], res[2])
```
devmotion, line 178:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return (logpdf(td.dist, x) + logjac, logjac)
```

src/transformed_distribution.jl
```diff
@@ -163,34 +163,34 @@ Makes use of the `forward` method to potentially re-use computation
...
 function logpdf_with_jac(td::MvTransformed{<:Dirichlet}, y::AbstractVector{<:Real})
     T = eltype(y)
     ϵ = _eps(T)
 
-    res = forward(inv(td.transform), y)
-    lp = logpdf(td.dist, mappedarray(x->x+ϵ, res.rv)) + res.logabsdetjac
-    return (lp, res.logabsdetjac)
+    res = forward(inverse(td.transform), y)
+    lp = logpdf(td.dist, mappedarray(x->x+ϵ, res[1])) + res[2]
+    return (lp, res[2])
```
devmotion, line 187:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    lp = logpdf(td.dist, mappedarray(x->x+ϵ, x)) + logjac
    return (lp, logjac)
```

src/transformed_distribution.jl
```diff
@@ -163,34 +163,34 @@ Makes use of the `forward` method to potentially re-use computation
...
 end
 
 # TODO: should eventually drop using `logpdf_with_trans`
 function logpdf_with_jac(td::MatrixTransformed, y::AbstractMatrix{<:Real})
-    res = forward(inv(td.transform), y)
-    return (logpdf_with_trans(td.dist, res.rv, true), res.logabsdetjac)
+    res = forward(inverse(td.transform), y)
+    return (logpdf_with_trans(td.dist, res[1], true), res[2])
```
devmotion, line 193:
```suggestion
    x, logjac = with_logabsdet_jacobian(inverse(td.transform), y)
    return (logpdf_with_trans(td.dist, x, true), logjac)
```

oschulz:
Sorry I missed to many `forward`s initially, @devmotion !

I fixed a few things, should be ready for another CI run now.

oschulz:
Should we also deprecate `logabsdetjac`? Currently, the default implementation of `with_logabsdet_jacobian` falls back on `logabsdetjac`, while ChangesOfVariables has `with_logabsdet_jacobian` as the primary function. JuliaMath/ChangesOfVariables.jl#3 (still undecided) would add a `logabsdet_jacobian` as an analog of `logabsdetjac` - but even if added, things would work the other way round, `logabsdet_jacobian` would fall back on `with_logabsdet_jacobian`.

The only package that currently seems add methods to `logabsdetjac` is Transits.jl, [in a single place](https://github.com/JuliaAstro/Transits.jl/blob/31e963e9d6db32ca51593542636f1799d76e0144/src/distributions.jl#L67), to define the LADJ of `Kipping13Transform`. That could easily be changed to `with_logabsdet_jacobian`, this would also add support for ChangesOfVariables.jl to Transits.jl. @mileslucas, would that be Ok from your side? Soss.jl also defines a method of `with_logabsdet_jacobian` in a single place in the current release, but that seems gone on the master branch (@cscherrer?).

I'm not sure we can cleanly support both ways in Bijectors.jl (users defining either `with_logabsdet_jacobian` or `logabsdetjac`, and the other function then using a default method it if not specialized as well), at least not without ugly trickery. Removing the "primiary" status from `logabsdetjac` would be breaking - but until we do, users can't code a Bijector the "ChangesOfVariables way", I think.

devmotion:
Maybe leave this for a separate PR as it seems to be a more fundamental change of Bijectors and, I assume, has to be benchmarked carefully? Maybe it would also benefit from an upstream definition of `logabsdet_jacobian` which seems to be another reason not to rush it.

In principle, however, I think one could use something like
```julia
logabsdetjac(b::AbstractBijector, x) = last(with_logabsdet_jacobian(b, x))
with_logabsdet_jacobian(b::AbstractBijector, x) = (b(x), logabsdetjac(b, x))
```
(potentially with some deprecations) to support implementations in both Bijectors and ChangesOfVariables style. But I guess it would be cleaner to do in a separate PR.

oschulz:
> In principle, however, I think one could use something like
> 
> ```julia
> logabsdetjac(b::AbstractBijector, x) = last(with_logabsdet_jacobian(b, x))
> with_logabsdet_jacobian(b::AbstractBijector, x) = (b(x), logabsdetjac(b, x))
> ```

Won't that result in a stack overflow if neither is defined?


> Maybe leave this for a separate PR

Sound good - I think @torfjelde was planning a deeper overhaul anyway?

devmotion:
> Won't that result in a stack overflow if neither is defined?

Yes, it does. But I assumed this would be fine - for a user or developer it's an indication that you should define (at least) one of these methods.

cscherrer:
> Soss.jl also defines a method of `with_logabsdet_jacobian` in a single place in the current release, but that seems gone on the master branch (@cscherrer?).

IIRC this had been an optional dependency, but it wasn't working out because of exported Distributions, and it didn't allow manifolds to be represented an embeddings from lower-dimensional spaces. I think I'll need to wait for https://github.com/TuringLang/Bijectors.jl/pull/183 before I can use it.

oschulz:
I get an intermittent (but frequent) test failure in "test/transform.jl:151": For

```julia
dist = Dirichlet([1000 * one(Float64), eps(Float64)])
x = rand(dist) == [0.9999999999999998, 0.0]
```

The test

```julia
@test logpdf(dist, x .+ ϵ) - _logabsdet(J) ≈ logpdf_turing
```

fails, due to

```julia
julia> using Bijectors, LinearAlgebra

julia> b = Bijectors.SimplexBijector{false}();
julia> x = [1.0 - eps(typeof(1.0)), 0.0];
julia> x2 = [1.0, 0.0];

julia> b(x), b(x2)
([35.35050620855721, 2.220446049250313e-16], [36.04365338911715, 0.0])

julia> logabsdetjac(b, x), logabsdet(Bijectors.jacobian(b, x))[1]
(36.04365338911715, 35.35050620855721)

julia> logabsdetjac(b, x2), logabsdet(Bijectors.jacobian(b, x2))[1]
(36.04365338911715, 36.04365338911715)
```

I get the same one the current master branch, I don't think it's related to this PR.

oschulz:
@devmotion one thing I hadn't realized until now is that Bijectors allows for `forward(b, multiple_xs_as_matrix)`, returning a matrix of result values and a vector of LADJ-values. Our current definition of `ChangesOfVariables.with_logabsdet_jacobian` doesn't really allow for that, I think. What's your take on this?

devmotion:
> Our current definition of ChangesOfVariables.with_logabsdet_jacobian doesn't really allow for that, I think. What's your take on this?

That this will be resolved by the refactoring and use of `Batch` or something similar (hopefully): https://github.com/TuringLang/Bijectors.jl/discussions/178

oschulz:
> That this will be resolved by the refactoring and use of Batch or something similar (hopefully): #178

Ok, then - I guess this break from contract isn't actually harmful for now, then. LADJ's are only ever added/subtracted, which works with a vector just as well as with a `Real`. ;-)

I don't understand where the 

```
Classic: Error During Test at /home/runner/work/Bijectors.jl/Bijectors.jl/test/ad/utils.jl:35
  Test threw exception
  Expression: ≈(ReverseDiff.gradient(f, x), finitediff, rtol=rtol, atol=atol)
  UndefVarError: inverse not defined
```

test error(s) are coming from, though. How can `inverse` not be defined?

devmotion:
> How can inverse not be defined inside ReverseDiff.gradient(f, x) if f(x) works perfectly fine?

The optional ReverseDiff and Tracker support are in different submodules in src/compat/reversediff.jl and src/compat/tracker.jl, to keep ReverseDiff.TrackedArray/Real and Tracker.TrackedArray/Real separate. I guess you have to load Bijectors.inverse (or InverseFunctions.inverse) in the submodule(s).

oschulz:
Will try.

The other persistent test failure can be traced back to this MWE

```julia
using Bijectors, Zygote
m = Bijectors.PartitionMask(3, [1], [2])
cl = Bijectors.Coupling(x -> Bijectors.Scale(x[1]), m)
x = [1.0, 2.0, 3.0]
dy = [4.0, 5.0, 6.0]
Zygote.pullback(cl, x)[2](dy)
```

resulting in

```
ERROR: MethodError: no method matching zero(::Type{Nothing})
```

This seems unrelated to this PR, though, happens on the current master branch of Bijectors as well.

devmotion:
There were quite many changes in Zygote recently, maybe they uncovered a bug in Bijectors or introduced one in Zygote.

oschulz:
Yes, that was my suspicion as well.

oschulz:
> I guess you have to load Bijectors.inverse (or InverseFunctions.inverse) in the submodule(s).

I tried (just pushed it here), but I still get `inverse not defined`.

oschulz:
Here's a more minimal MWE for the Zygote problem:

```julia
using Bijectors, Zygote
using Bijectors: PartitionMask, combine

m = PartitionMask(3, [1], [2])
a, b, c = ([1.0], [2.0], [3.0])

y = combine(m, a, b, c)
Zygote.pullback(combine, m, a, b, c)[2](y)
```

results in

```
ERROR: LoadError: MethodError: no method matching zero(::Type{Nothing})
```

oschulz:
Ok, added an rrule for `combine(m::PartitionMask, x_1, x_2, x_3)`, that took case of the Zygote errors (I hope the rrule is correct).

Now all that's left seem to be the two `UndefVarError: inverse not defined`.

oschulz:
Ok, I `pkg> test Bijectors` passed locally for me now. Could someone trigger the CI workflow?

src/bijectors/coupling.jl
```diff
@@ -118,6 +118,20 @@ Combines `x_1`, `x_2`, and `x_3` into a single vector.
 """
 @inline combine(m::PartitionMask, x_1, x_2, x_3) = m.A_1 * x_1 .+ m.A_2 * x_2 .+ m.A_3 * x_3
 
+function ChainRulesCore.rrule(::typeof(combine), m::PartitionMask, x_1, x_2, x_3)
```
devmotion, line 121: Can you move the rule to src/chainrules.jl and add a test in test/ad/chainrules.jl?

src/bijectors/coupling.jl
```diff
@@ -118,6 +118,20 @@ Combines `x_1`, `x_2`, and `x_3` into a single vector.
...
+    prj = map(ChainRulesCore.ProjectTo, (x_1, x_2, x_3))
```
devmotion, line 122: Can we just define three separate projectors? Then we also don't have to rely on constant propagation in the pullback:
```suggestion
    proj_x_1 = ChainRulesCore.ProjectTo(x_1)
    proj_x_2 = ChainRulesCore.ProjectTo(x_2)
    proj_x_3 = ChainRulesCore.ProjectTo(x_3)
```

src/bijectors/coupling.jl
```diff
@@ -118,6 +118,20 @@ Combines `x_1`, `x_2`, and `x_3` into a single vector.
...
+
+    function _transform_ordered_adjoint(ΔΩ)
```
devmotion, line 124: Can we use maybe the name
```suggestion
    function combine_pullback(ΔΩ)
```
here? I think this would be a bit more in line with the usual CR conventions.

src/bijectors/coupling.jl
```diff
@@ -118,6 +118,20 @@ Combines `x_1`, `x_2`, and `x_3` into a single vector.
...
+        Δ = ChainRulesCore.unthunk(ΔΩ)
+        dx_1, dx_2, dx_3 = partition(m, Δ)
+
+        return ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), prj[1](dx_1), prj[2](dx_2), prj[3](dx_3)
```
devmotion, line 128:
```suggestion
        return ChainRulesCore.NoTangent(), ChainRulesCore.NoTangent(), proj_x_1(dx_1), proj_x_2(dx_2), proj_x_3(dx_3)
```

src/bijectors/coupling.jl
```diff
@@ -118,6 +118,20 @@ Combines `x_1`, `x_2`, and `x_3` into a single vector.
...
+    end
+
+    return combine(m, x_1, x_2, x_3), _transform_ordered_adjoint
```
devmotion, line 131:
```suggestion
    return combine(m, x_1, x_2, x_3), combine_pullback
```

src/bijectors/coupling.jl
```diff
@@ -195,6 +209,7 @@ function couple(cl::Coupling, x::AbstractVector)
     return b
 end
 
+#!!!!!!!!!!!!!
```
devmotion, line 212:
```suggestion
```

test/runtests.jl
```diff
@@ -1,16 +1,19 @@
...
 using FiniteDifferences
 using ForwardDiff
 using Functors
+using InverseFunctions
 using LogExpFunctions
 using ReverseDiff
 using Tracker
 using Zygote
 
+
```
devmotion, line 16:
```suggestion
```

src/bijectors/coupling.jl
oschulz: > Can we use maybe the name

Ooops, copy/paste error :-)

src/bijectors/coupling.jl
oschulz: Oppps, that was a debug marker.

src/bijectors/coupling.jl
oschulz: Will do

oschulz: Done.

oschulz:
Yay, tests green. :-)

README.md
```diff
@@ -542,41 +542,42 @@ Logit{Float64}(0.0, 1.0)
...
+julia> with_logabsdet_jacobian(b, 0.6)         # defaults to `(b(x), logabsdetjac(b, x))`
+(0.4054651081081642, 1.4271163556401458)
 ```
 
-For further efficiency, one could manually implement `forward(b::Logit, x)`:
+For further efficiency, one could manually implement `with_logabsdet_jacobian(b::Logit, x)`:
 
 ```julia
 julia> import Bijectors: forward, Logit
+julia> import ChangesOfVariables: with_logabsdet_jacobian
```
devmotion, line 566: The `import Bijectors: forward` can be removed. I wonder if it is easier for users (or at least in the README) if we use
```suggestion
julia> using Bijectors: Logit

julia> import Bijectors: with_logabsdet_jacobian
```
Alternatively, one could use
```suggestion
julia> using Bijectors: Logit

julia> import ChangesOfVariables: with_logabsdet_jacobian
```
But then one has to add the ChangesOfVariables dependency explicitly.

test/interface.jl
```diff
@@ -847,10 +847,19 @@ end
...
     end
 end
 
+@testset "deprecations" begin
+    b = Bijectors.Exp()
+    x = 0.3
+
+    @test let r = forward(b, x)
+        (r.rv, r.logabsdetjac) == with_logabsdet_jacobian(b, x)
+    end
```
devmotion, line 863: Maybe use something like
```suggestion
    @test @test_deprecated(forward(b, x)) == NamedTuple{(:rv, :logabsdetjac)}(with_logabsdet_jacobian(b, x))
```
to also check that the method is deprecated and that it actually returns a named tuple with the two fields.

test/interface.jl
```diff
@@ -847,10 +847,19 @@ end
...
+    @test inv(b) == inverse(b)
```
devmotion, line 864: Similarly, here maybe use
```suggestion
    @test @test_deprecated(inv(b)) == inverse(b)
```

devmotion (commented):
It looks quite good, I only added some minor comments :slightly_smiling_face: Since the PR is quite huge I'll wait for @torfjelde's approval though.

In the meantime, can you update the version number and fix the deprecation warnings that show up in the tests (https://github.com/TuringLang/Bijectors.jl/runs/4499268477?check_suite_focus=true)? Some of them should already be fixed by using `@test_deprecated` as suggested in my comments.

oschulz:
> fix the deprecation warnings that show up in the tests

On it.

torfjelde:
Sorry for being a bit awol; past week has been busy, preparing to go home for Christmas.
But just had a quick look and this looks great! Thank you @oschulz !

It seems like @devmotion has already done a proper review of this, so tbh I don't have any comments (beyond his latest on deprecation tests and bumping the verison number) :+1: 

So feel free to approve it when you're happy @devmotion . I'll be travelling tomorrow, so won't be able to have a look at this again until Tues at the earliest.

oschulz:
> But just had a quick look and this looks great! Thank you @oschulz !

Thanks! It did get a log bigger than I had expected, initially. :-)

> In the meantime, can you update the version number and fix the deprecation warnings

Version number is up and I think I finally eliminated the last remaining deprecation warnings. Let's see if the tests go through clean this time.

oschulz:
Ok, looks clean.

test/runtests.jl
```diff
@@ -1,11 +1,13 @@
 using Bijectors
 
 using ChainRulesTestUtils
+using ChangesOfVariables
 using Combinatorics
 using DistributionsAD
 using FiniteDifferences
 using ForwardDiff
 using Functors
+using InverseFunctions
```
devmotion, line 10: This shouldn't be needed, should it? Since `inverse` and `with_logabsdet_jacobian` are part of the API now I guess it would make sense to export them?

devmotion (commented):
Great, looks very good! Just one minor question: I think we should re-export `inverse` and `with_logabsdet_jacobian` since they are part of the API now - or is there a specific reason for not exporting them?

test/runtests.jl
oschulz: >  I think we should re-export inverse and with_logabsdet_jacobian since they are part of the API now - or is there a specific reason for not exporting them?

I've been wondering about that myself - I have no strong opinion on it. In principle we don't have to re-export every API we implement methods for, of course - but then these function a central to Bijectors now.

oschulz: We're re-exporting the whole of Distributions, so ... :-)

devmotion: My main argument for exporting them is that they replace `forward` and `inv` which are exported and always available, respectively.

oschulz: Very true.

Ok, re-exports are added (last commit).

oschulz:
Ok, `inverse` and `with_logabsdet_jacobian` are re-exported.

Project.toml
```diff
@@ -1,13 +1,15 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.9.11"
+version = "0.9.12"
```
cpfiffer, line 3: This should probably be 0.10.0, given the magnitude of changes (and that `inv` is now `inverse`, among other things).

oschulz: I don't have a strong opinion in it - @devmotion you did consider it non-breaking, right?

devmotion: I thought about it again, and I still think it is non-breaking *if* we add fallback definitions for `inverse` and `with_logabsdet_jacobian`:
```julia
function inverse(b::AbstractBijector)
    Base.depwarn("`inv(b::AbstractBijector)` is deprecated, please use `inverse(b)` instead", :inverse)
    return inv(b)
end
function with_logabsdet_jacobian(b::AbstractBijector, x)
    Base.depwarn(
        "`forward(b::AbstractBijector, x)` is deprecated, please use `with_logabsdet_jacobian(b, x)`", 
        :with_logabsdet_jacobian,
    )
    return forward(b, x)
end
```
This is the only breaking change I can imagine with this PR: If a function that operates with bijectors is defined with the new API (maybe even in Bijectors) but the bijector at hand only implements the old API. This can lead to a StackOverflow error - but only if for a bijector neither the old nor the new API is implemented, and hence the implementation is broken anyway.

Otherwise, `forward` and `inv` are deprecated and the other changes are merely replacements in the code and tests (to fix deprecation warnings). So even though the PR is quite large the changes itself seem small and well defined.

@oschulz can you add the fallback definitions, and ideally also test them (e.g. with a dummy bijector that only implements the old API)? Then I am convinced that the PR is non-breaking.

oschulz: > @oschulz can you add the fallback definitions, and ideally also test them

Yes, will do.

oschulz: > @oschulz can you add the fallback definitions, and ideally also test them

I think I found a way to do that and defend against the stack overflow, so we can return a meaningful error if neither `forward` or `with_logabsdet_jacobian` is defined, my using a wrapper bijector. The same mechanism can also be used to allow defining Bijectors via `with_logabsdet_jacobian` without defining `logabsdetjac`.

@devmotion, I think we we implement JuliaMath/ChangesOfVariables.jl#3 we could then immediately deprecate `logabsdetjac` and still keep this non-breaking, using the same wrapper trick. Let me try something ...

oschulz: @devmotion I was wrong, my idea doesn't work. And I don't think a stack overflow would be a great solution, it wouldn't tell users what went wrong, and can be hard to debug. Also, there's this problem, we already define defaults

```julia
with_logabsdet_jacobian(b::Bijector, x) = (b(x), logabsdetjac(b, x))
inverse(b::Bijector) = Inverse(b)
```

So even the deprecation fallbacks `inverse(b::AbstractBijector)` and `with_logabsdet_jacobian(b::AbstractBijector, x)` wouldn't really help if a custom bijector inherits from `Bijector` instead of `AbstractBijector`.

Maybe this will have to be v0.10.0 after all?

devmotion: Sorry, I wanted to think a bit about it and see if there is a good and clean solution. I am not too much concerned with StackOverflow errors since they will not be possible anymore when the old API is removed and they only occur if the interface is not implemented (neither the old nor the new API). However, the `::Bijector` defaults are troublesome because they will break downstream bijectors that implement e.g. `inv` in a different way.

So I would suggest making a clean cut: Let's make a breaking release without any deprecations and remove all occurrences of `forward` and `inv` completely. This means we could also revise our decision to reexport `inverse` and `with_logabsdet_jacobian` - maybe it would be a good opportunity to change the current behaviour and not export them?

oschulz: I'm fine with both exporting and not exporting them, with a slight preference for not exporting.

I looked at the package that currently depend on Bijectors directly (DifferentialEvolutionMCMC DynamicPPL Turing TuringModels Transits ParameterHandling MeasureTheory AdvancedVI AIBECS Soss): If we keep the deprecations, I think no changes there would be necessary except for adding Bijectors 0.10 to compat. If we remove the deprecations, some changes in the dependent package would be necessary, though nothing major.

Since @torfjelde is preparing a major revision to Bijectors: maybe we should do a release that's only technically breaking  (the dependent package use `inv` and `forward` in a few places, but don't specialize them at all) here, and then remove the deprecations later on as part of TuringLang/Bijectors.jl#183? That way, the dependent package could switch from the using old to using the new API in the mean time.

devmotion: I like this suggestion, probably this would simplify updates to Bijectors 0.10.

oschulz: Ok, I changed the version number to v0.10.0,

oschulz: Should I remove the export of `inverse` and `with_logabsdet_jacobian`?

devmotion: I think reexporting makes it easy to miss to which package the function actually belongs. Generally, I started to think one should be a bit more careful when it comes to reexporting since it means any breaking change of the upstream definitions seems to require a breaking release in the downstream package as well.

On the other hand, it might seem a bit strange to not export them if they are part of the API :man_shrugging: 

What do you think @torfjelde?

torfjelde: > maybe we should do a release that's only technically breaking (the dependent package use inv and forward in a few places, but don't specialize them at all) here, and then remove the deprecations later on as part of #183? That way, the dependent package could switch from the using old to using the new API in the mean time.

I'm also in favour of this: sounds good :+1:

> Should I remove the export of inverse and with_logabsdet_jacobian?

Personally, I'm in favour of exporting. It's very rare someone does `using Bijectors` _without_ the intention of also using `inverse` and/or `with_logabsdet_jacobian` since implementations of these is essentially the point of Bijectors.jl, hence it seems a bit weird to me if they then need to qualify the usages of these functions :shrug:

oschulz: Ok, so we keep the export? If so, this PR should be good to go from my side.

devmotion: Yes, let's keep it :+1:

devmotion (approved):
Great, thanks a lot @oschulz!

oschulz:
Thanks for all the comments and suggestions!

torfjelde:
Indeed, thank you so much @oschulz ! Great stuff:)

oschulz:
@devmotion if it's fine with you in principle, I'd draft a PR for distributions to "lift" TransformedDistribution from Bijectors into Distributions. I think it will be a very valuable tool to have, and with InverseFunctions and ChangesOf Variables (and support for them in Bijectors) in place we can make it available without a dependency on Bijectors (I think, haven't drafted the code yet).

devmotion:
Sounds reasonable - currently we handle only the special case of affine transformations (limited but hopefully soon in a bit more general way) so I think it would be a valuable addition.

oschulz:
> Sounds reasonable - currently we handle only the special case of affine transformations

I do have a plan to (hopefully) support arbitrary variate types (I definitely want `ValueShapes.NamedTupleDist` to work with this). I will probably need to include defining `Random.gentype` for distributions (I think that was already under discussion?) and I may need to at least define new VariateForm for struct types. @devmotion if that's fine with you at least in principle, I would make a concrete draft PR for discussion.

devmotion:
It would be nice to keep changes as minimal as possible (but e.g. definitions so general that they allow such use cases without breaking changes later on), such that the PR does not become too large and it is less likely that discussions diverge and/or not focus on the main changes.

devmotion:
In particular the Random stuff is a very sensitive area where people tend to have strong opinions :smile:

oschulz:
Understood - I'll try to make something compact.

devmotion:
BTW the current design proposal for `eltype` (but on purpose not including `gentype` etc.) is https://github.com/JuliaStats/Distributions.jl/pull/1433. But maybe it is sufficient to just call regular `rand` of the untransformed distribution for sampling?

oschulz:
> But maybe it is sufficient to just call regular rand of the untransformed distribution for sampling

The tricky bit will be inferring the `VariateForm` of the transformed distribution. `Base._return_type(f, (Random.gentype(orig_dist),))` will allow us to infer that in many cases, falling back on running the trafo on a single rand value and determining the `VariateForm` from the results (in the spirit of what Broadcast does when it can't infer the return type).

PR #214, opened 2022-02-11
Widening the scope of the package and dropping support for batching

torfjelde:
This PR is an attempt at a couple of things:
1. Widening the scope, i.e. not being so restrictive regarding whether or not something is bijective.
2. Removal of expected dimensionality of inputs from the type of the `Bijector`.
   - This had the benefit of disambiguating whether or not an input represents a single input or a _collection_ of inputs, _but_ it's super annoying to deal with both for the user and implementer + restricts us to only working with arrays, and transformations are expected to preserve dimensionality.
   - After this PR there is no longer "official" support for batching. Some bijectors work for batches because there's no need for disambiguation, but we don't claim that this works for all.
3. We now also adopt the ChangesOfVariables.jl interface.
   - There's one issue with this though: https://github.com/TuringLang/Bijectors.jl/pull/183/files#r804539211. A bit uncertain what to do here.

TODOs:
- [ ] Should we remove (basically) unused and partially unsupported methods: `logpdf_with_jac`, `logpdf_forward`, `forward(d::Distribution, ...)`?
  - E.g. `forward` no longer works for multiple samples because we no longer have support for batched inputs.
  - Just removing these seems like the most straight-forward and useful approach atm, so we might as well.


src/Bijectors.jl
```diff
@@ -253,13 +258,18 @@ include("utils.jl")
...
 include("chainrules.jl")
 
-Base.@deprecate forward(b::AbstractBijector, x) NamedTuple{(:rv,:logabsdetjac)}(with_logabsdet_jacobian(b, x))
+Base.@deprecate forward(b::Union{Transform,Function}, x) NamedTuple{(:rv,:logabsdetjac)}(with_logabsdet_jacobian(b, x))
 
-@noinline function Base.inv(b::AbstractBijector)
+@noinline function Base.inv(b::Transform)
     Base.depwarn("`Base.inv(b::AbstractBijector)` is deprecated, use `inverse(b)` instead.", :inv)
     inverse(b)
 end
```
torfjelde, line 266: Should probably just remove these deprecations now that we're making a breaing release

torfjelde:
This is a fairly big one @devmotion , but I would greatly appreciate it if you had a super-quick look. The main part is just removing the dimensionality completely from the definition of the bijectors, in addition to a couple of small things:
1. Adding mutating methods, e.g. `transform!`, `logabsdetjac!`, and `with_logabsdet_jacobian!` (should this maybe go to ChangesOfVariables.jl?).
2. Added docs.
3. Using `elementwise(exp)` to indicate that a method should be applied elementwise (equivalent to `Base.Fix1(broadcast, exp)`).

I'm also going to make a separate PR to remove the stuff related to `ADbackend`.

This will be a huge breaking release, as I think it's time we just rip the band-aid off.

torfjelde:
@yebai you might also want to take a look at this

docs/src/transforms.md
```diff
@@ -0,0 +1,105 @@
...
+- The derivative of `exp` at an input `x` is simply `exp(x)`, hence `logabsdetjac` is simply `x`.
+
+```@repl usage
+using Bijectors
+transform(exp, 1.0)
+logabsdetjac(exp, 1.0)
+with_logabsdet_jacobian(exp, 1.0)
+```
+
+Some transformations is well-defined for different types of inputs, e.g. `exp` can also act elementwise on a `N`-dimensional `Array{<:Real,N}`. To specify that a transformation should be acting elementwise, we use the [`elementwise`](@ref) method:
```
devmotion, line 14:
```suggestion
Some transformations are well-defined for different types of inputs, e.g. `exp` can also act elementwise on an `N`-dimensional `Array{<:Real,N}`.
To specify that a transformation should act elementwise, we use the [`elementwise`](@ref) method:
```

docs/src/transforms.md
```diff
@@ -0,0 +1,105 @@
...
+```julia
+with_logabsdet_jacobian
+```
+
+With mutation:
+
+```@docs
+transform!
+logabsdetjac!
+with_logabsdet_jacobian!
```
devmotion, line 56: I think it's fine to define these methods in Bijectors initially and possibly in the future move them to ChangesOfVariables.

src/Bijectors.jl
```diff
@@ -54,29 +54,28 @@ export  TransformDistribution,
...
         MultivariateTransformed,
-        logpdf_with_jac,
-        logpdf_forward,
         PlanarLayer,
         RadialLayer,
-        CouplingLayer,
-        InvertibleBatchNorm
+        Coupling,
+        InvertibleBatchNorm,
+        elementwise
```
devmotion, line 78: Hmm... I wonder if we could reuse some existing functionality in the ecosystem here. And/or if there is a shorter name. Regarding the first point, e.g., `Transducers.Map` seems similar?

src/Bijectors.jl
```diff
@@ -253,12 +265,10 @@ include("utils.jl")
...
 
-Base.@deprecate forward(b::AbstractBijector, x) NamedTuple{(:rv,:logabsdetjac)}(with_logabsdet_jacobian(b, x))
+Base.@deprecate NamedBijector(bs) NamedTransform(bs)
 
-@noinline function Base.inv(b::AbstractBijector)
-    Base.depwarn("`Base.inv(b::AbstractBijector)` is deprecated, use `inverse(b)` instead.", :inv)
-    inverse(b)
-end
+Base.@deprecate Exp() elementwise(exp) false
+Base.@deprecate Log() elementwise(log) false
```
devmotion, line 271: Are deprecations needed if it is a breaking release? Or would it be sufficient to add them to some changelog/announcement/NEWS.md?

src/bijectors/corr.jl
```diff
@@ -61,26 +61,24 @@ Note: The implementation doesn't follow their "manageable expression" directly,
...
     r = _link_chol_lkj(w) 
     return r + zero(x) 
     # This dense format itself is required by a test, though I can't get the point.
     # https://github.com/TuringLang/Bijectors.jl/blob/b0aaa98f90958a167a0b86c8e8eca9b95502c42d/test/transform.jl#L67
 end
 
-(b::CorrBijector)(X::AbstractArray{<:AbstractMatrix{<:Real}}) = map(b, X)
-
-function (ib::Inverse{<:CorrBijector})(y::AbstractMatrix{<:Real})
+function transform(ib::Inverse{<:CorrBijector}, y::AbstractMatrix{<:Real})
```
devmotion, line 76: Since there are no type parameters anymore:
```suggestion
function transform(ib::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})
```

src/bijectors/corr.jl
```diff
@@ -61,26 +61,24 @@ Note: The implementation doesn't follow their "manageable expression" directly,
...
     w = _inv_link_chol_lkj(y)
     return w' * w
 end
-(ib::Inverse{<:CorrBijector})(Y::AbstractArray{<:AbstractMatrix{<:Real}}) = map(ib, Y)
-
 
-function logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})
+function logabsdetjac(::Inverse{<:CorrBijector}, y::AbstractMatrix{<:Real})
```
devmotion, line 81:
```suggestion
function logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})
```

src/bijectors/exp_log.jl
```diff
@@ -1,49 +1,13 @@
-#############
-# Exp & Log #
-#############
+invertible(::typeof(exp)) = Invertible()
```
devmotion, line 1: Seems similar to InverseFunctions?

src/bijectors/leaky_relu.jl
```diff
@@ -1,101 +1,29 @@
...
-    end
-    return (J .* x, log.(abs.(J)))
-end
-
-# (N=1) Multivariate case
-function (b::LeakyReLU{<:Any, 1})(x::AbstractVecOrMat)
-    return let z = zero(eltype(x))
-        @. (x < z) * b.α * x + (x > z) * x
-    end
+    J = mask * b.α + !mask
```
devmotion, line 20: Isn't this just
```suggestion
    J = x < zero (x) ?  b.α : one(b.α)
```

src/bijectors/leaky_relu.jl
```diff
@@ -1,101 +1,29 @@
...
-        logjac = vec(sum(log.(abs.(J)); dims = 1))  # sum along column
-    end
-
-    y = J .* x
-    return (y, logjac)
+# Array inputs.
+function with_logabsdet_jacobian(b::LeakyReLU, x::AbstractArray)
+    mask = x .< zero(eltype(x))
+    J = mask .* b.α .+ (!).(mask)
+    return J .* x, sum(log.(abs.(J)))
```
devmotion, line 28: Would
```suggestion
    return J .* x, sum(log \circ abs, J)
```
work with AD as well?

src/bijectors/leaky_relu.jl
```diff
@@ -1,101 +1,29 @@
...
-    elseif x isa AbstractMatrix
-        logjac = vec(sum(log.(abs.(J)); dims = 1))  # sum along column
-    end
-
-    y = J .* x
-    return (y, logjac)
+# Array inputs.
+function with_logabsdet_jacobian(b::LeakyReLU, x::AbstractArray)
+    mask = x .< zero(eltype(x))
+    J = mask .* b.α .+ (!).(mask)
```
devmotion, line 27: Just something like
```suggestion
    J = broadcast(b.α, x) do α, xi
        return x < zero(x) ? α : one(α)
    end
```
Could possibly even factor out the scalar method and re-use it here?

src/bijectors/named_bijector.jl
```diff
@@ -25,21 +23,25 @@ julia> (a = 2 * x.a, b = exp(x.b), c = x.c)
...
+function Functors.functor(::Type{<:NamedTransform{names}}, x) where names
     function reconstruct_namedbijector(xs)
-        return NamedBijector{names,typeof(xs.bs)}(xs.bs)
+        return NamedTransform{names,typeof(xs.bs)}(xs.bs)
     end
     return (bs = x.bs,), reconstruct_namedbijector
 end
 
-names_to_bijectors(b::NamedBijector) = b.bs
+# TODO: Use recursion instead of `@generated`?
```
devmotion, line 38: Yeah, not a big fan of generated functions if they are not needed. Isn't it sufficient to use `map` here? I think usually that works quite well with `NamedTuple`.

src/bijectors/named_bijector.jl
```diff
@@ -25,21 +23,25 @@ julia> (a = 2 * x.a, b = exp(x.b), c = x.c)
...
+@generated function inverse(b::NamedTransform{names}) where {names}
+    return :(NamedTransform(($([:($n = inverse(b.bs.$n)) for n in names]...), )))
+end
```
devmotion, line 41: Maybe just
```suggestion
inverse(t::NamedTransform) = NamedTransform(map(inverse, t.bs))
```
?

src/bijectors/named_bijector.jl
```diff
@@ -55,146 +57,52 @@ names_to_bijectors(b::NamedBijector) = b.bs
...
 
-@generated function inverse(b::NamedBijector{names}) where {names}
-    return :(NamedBijector(($([:($n = inverse(b.bs.$n)) for n in names]...), )))
-end
-
-@generated function logabsdetjac(b::NamedBijector{names}, x::NamedTuple) where {names}
+@generated function logabsdetjac(b::NamedTransform{names}, x::NamedTuple) where {names}
     exprs = [:(logabsdetjac(b.bs.$n, x.$n)) for n in names]
     return :(+($(exprs...)))
 end
```
devmotion, line 63: Maybe just
```julia
function logabsdetjac(t::NamedTransform{names}, x::NamedTuple{names}) where {names}
    return sum(map(logabsdetjac, t.bs, x))
end
```
(and possibly, if really needed, another method that reorders `x`)?

src/bijectors/ordered.jl
```diff
@@ -45,8 +47,7 @@ function _transform_ordered(y::AbstractMatrix)
     return x
 end
 
-(ib::Inverse{<:OrderedBijector})(x::AbstractVecOrMat) = _transform_inverse_ordered(x)
-
+transform(ib::Inverse{<:OrderedBijector}, x::AbstractVecOrMat) = _transform_inverse_ordered(x)
```
devmotion, line 50:
```suggestion
transform(ib::Inverse{OrderedBijector}, x::AbstractVecOrMat) = _transform_inverse_ordered(x)
```

src/bijectors/pd.jl
```diff
@@ -7,19 +7,17 @@ function replace_diag(f, X)
...
+transform(b::PDBijector, X::AbstractMatrix{<:Real}) = pd_link(X)
 function pd_link(X)
     Y = lower(parent(cholesky(X; check = true).L))
     return replace_diag(log, Y)
 end
-(b::PDBijector)(X::AbstractArray{<:AbstractMatrix{<:Real}}) = map(b, X)
 lower(A::AbstractMatrix) = convert(typeof(A), LowerTriangular(A))
 
-function (ib::Inverse{<:PDBijector})(Y::AbstractMatrix{<:Real})
+function transform(ib::Inverse{<:PDBijector}, Y::AbstractMatrix{<:Real})
```
devmotion, line 17:
```suggestion
function transform(ib::Inverse{PDBijector}, Y::AbstractMatrix{<:Real})
```

src/bijectors/shift.jl
```diff
@@ -1,37 +1,26 @@
...
-function Functors.functor(::Type{<:Shift{<:Any,N}}, x) where N
-    function reconstruct_shift(xs)
-        return Shift{typeof(xs.a),N}(xs.a)
-    end
-    return (a = x.a,), reconstruct_shift
-end
+Functors.@functor Shift
 
-up1(b::Shift{T, N}) where {T, N} = Shift{T, N + 1}(b.a)
+Base.show(io::IO, b::Shift) = print(io, "Shift($(b.a))")
```
devmotion, line 12: Are such definitions really needed? Isn't that (basically) what you get by default?

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
+
+Abstract type for a transformation.
+
+# Implementing
+
+A subtype of `Transform` of should at least implement [`transform(b, x)`](@ref).
+
+If the `Transform` is also invertible:
+- Required:
+  - [`invertible`](@ref): should return instance of [`Invertible`](@ref) or [`NotInvertible`](@ref).
```
devmotion, line 59: Why not just use `InverseFunctions`? If `inverse` is not defined, it returns by default `NoInverse` (IIRC, name might be slightly different), so you can check if the function is invertible (or rather, if `ìnverse` is defined).

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
 ######################
-"Abstract type for a bijector."
-abstract type AbstractBijector end
+"""
+
+Abstract type for a transformation.
+
+# Implementing
+
+A subtype of `Transform` of should at least implement [`transform(b, x)`](@ref).
```
devmotion, line 55: My main concern here is a bit that the names `Transform` and `transform` seem a bit generic, and if exported likely conflict with exports of other packages.

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
+    can exploit shared computation in the two.
+
+For the above methods, there are mutating versions which can _optionally_ be implemented:
+- [`with_logabsdet_jacobian!`](@ref)
+- [`logabsdetjac!`](@ref)
+- [`with_logabsdet_jacobian!`](@ref)
+"""
+abstract type Transform end
+
+(t::Transform)(x) = transform(t, x)
```
devmotion, line 75: Are both forms needed? Wouldn't it be cleaner to only make them callable? (Or only use `transform`)

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
+
+Broadcast.broadcastable(b::Transform) = Ref(b)
+
+"""
+    transform(b, x)
+
+Transform `x` using `b`, treating `x` as a single input.
+"""
+transform(f::Function, x) = f(x)
```
devmotion, line 84: Not sure if it matters, but Julia won't specialize on `f` here.

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
-"Abstract type of bijectors with fixed dimensionality."
-abstract type Bijector{N} <: AbstractBijector end
+Transform `x` using `b`, storing the result in `y`.
 
-dimension(b::Bijector{N}) where {N} = N
-dimension(b::Type{<:Bijector{N}}) where {N} = N
+If `y` is not provided, `x` is used as the output.
+"""
+transform!(b, x) = transform!(b, x, x)
+transform!(b, x, y) = (y .= transform(b, x))
```
devmotion, line 95: Is broadcasting the best fallback here? Seems at least not obvious from the function name that the arguments are broadcasted. Maybe rather
```suggestion
transform!(b, x, y) = copyto!(y, transform(b, x))
```
?

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
 
-Broadcast.broadcastable(b::Bijector) = Ref(b)
+"""
+    logabsdetjac(b, x)
 
+Return `log(abs(det(J(b, x))))`, where `J(b, x)` is the jacobian of `b` at `x`.
 """
-    isclosedform(b::Bijector)::bool
-    isclosedform(b⁻¹::Inverse{<:Bijector})::bool
+logabsdetjac(b, x) = last(with_logabsdet_jacobian(b, x))
```
devmotion, line 102: Is this needed? Seems like something that - if desired - should maybe go to ChangesOfVariables.

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
+"""
+struct Invertible end
 
-# field contains nested numerical parameters
-Functors.@functor Inverse
+# Useful for checking if compositions, etc. are invertible or not.
+Base.:+(::NotInvertible, ::Invertible) = NotInvertible()
+Base.:+(::Invertible, ::NotInvertible) = NotInvertible()
+Base.:+(::NotInvertible, ::NotInvertible) = NotInvertible()
+Base.:+(::Invertible, ::Invertible) = Invertible()
```
devmotion, line 161: I would recommend removing these definitions. Seems like a misuse of `+`.

src/interface.jl
```diff
@@ -31,111 +46,201 @@ ADBackend(::Val) = error("The requested AD backend is not available. Make sure t
...
-efficiencies, if they exist.
+Functors.@functor Inverse
+
+"""
+    inverse(t::Transform[, ::Invertible])
+
+Returns the inverse of transform `t`.
 """
-with_logabsdet_jacobian(b::Bijector, x) = (b(x), logabsdetjac(b, x))
+inverse(t::Transform) = Inverse(t)
```
devmotion, line 203: Just remove this definition, and it should be sufficient to operate with `InverseFunctions.NoInverse` instead of `Invertible`/`NoInvertible`?

src/interface.jl
torfjelde: Ah, happy to! Just wasn't aware it did this. Will change accordingly.

src/interface.jl
torfjelde: I personally find this useful, and the last time we spoke about this (https://github.com/JuliaMath/ChangesOfVariables.jl/issues/3), there didn't seem to be a desire to add it :confused:

src/bijectors/leaky_relu.jl
torfjelde: Yeah but I believe I did it this way originally due to GPU compat or something (which I obviously should have made a comment regarding...); am slightly worried about making a change to this here :confused:

src/bijectors/shift.jl
torfjelde: Without the type parameter, yeah. I'll remove this here though; I think was just playing around with whether or not we should do this and left it in this PR by a mistake.

src/interface.jl
torfjelde: Okay so one thing I find non-ideal with `InverseFunctions.NoInverse` is that `inverse` sometimes performs operations, e.g. if you didn't implement the inverse of `Scale` lazily, you might have to invert a matrix just to check if it's invertible or not. I will make the change though, but this is annoying :confused:

src/interface.jl
torfjelde: I think we definitively want `transform` to stay consistent with `transform!` + it gives us a method we have ownership over to overload if need be. So I think it's really just a question of whether we want to support calling `t` directly or not.

Buuuut it's so convenient to be able to call `t` directly, e.g. writing `map(t, ...)` rather than `map(Base.Fix1(transform, t), ...)`, etc. I kind of want it :confused:

src/interface.jl
torfjelde: In addition, the behavior I _expect_ when calling `inverse(f)` is not to get some lazy expression. I expect to get a callable back that computes `inverse(f)` in the most efficient way.

For example,

``` julia
inverse(Scale(A))
```

I'd expect to give me a `Scale(inv(A))`, not `Inverse{<:Scale}` which uses `A \ x` under the hood.

I think to support the lazy versions, we should instead add methods such as `transform_inverse` and `with_logabsdet_jacobian_inverse`.

devmotion: > In addition, the behavior I expect when calling inverse(f) is not to get some lazy expression. I expect to get a callable back that computes inverse(f) in the most efficient way.

Isn't that was `inverse` is supposed to return? For instance, `inverse(exp) = log` - you get a callable back that computes the inverse of `exp`?

> I'd expect to give me a Scale(inv(A)), not Inverse{<:Scale} which uses A \ x under the hood.

If you want `Scale(inv(A))`, you can just define `inverse(s::Scale) = Scale(inv(s.A))`, no? And if you want `Inverse{<:Scale}`, you would define `inverse(s::Scale) = Inverse(s)` and `(t::Inverse{<:Scale})(x) = t.x.A \ x`?

src/interface.jl
devmotion: An idea for the last point could be to support a curried form as well, e.g.,
```julia
transform(t::Transform) = Base.Fix1(transform, t)
```
Then you could write `map(transform(t), ...)` which arguably is a bit nicer.

src/interface.jl
torfjelde: True but it also seems like quite a natural naming :confused:

src/interface.jl
torfjelde: Ah good catch!

src/interface.jl
torfjelde: I'm honestly happy with that:) I'll make the change.

torfjelde: Wait. The note above regarding how `transform` might collide with other packages is a possible argument to keep the callable version around? :confused:

src/interface.jl
torfjelde: Done!

src/bijectors/named_bijector.jl
torfjelde: Though I agree with you, I think we should defer reducing usage of `@generated` to a separate PR. I made an issue: https://github.com/TuringLang/Bijectors.jl/issues/243

src/bijectors/leaky_relu.jl
torfjelde: See above comment

src/bijectors/leaky_relu.jl
torfjelde: I don't know which is preferred tbh, but I think I'll leave as is for now though.

src/bijectors/leaky_relu.jl
torfjelde: I'll leave it as is for now. 

Once we've made this big breaking change, we can start looking at improving the overall implementations.

src/Bijectors.jl
torfjelde: Removed!

src/interface.jl
torfjelde: Done!

src/Bijectors.jl
torfjelde: I'm happy to take suggestions, but I'm also okay with `elementwise`, so IMO this shouldn't hold this PR back.

Not too big of a fan to depend use `Transducers.Map` though; seems like unnecessary complexity just to make `Base.Fix1(broadcast, f)`.

torfjelde:
I think I've replied/addressed your comments now @devmotion 

Some I'm of the opinion that we leave until later PRs, given how long this PR has been in the pipeline + a lot of improvements we want to do (e.g. adding a `VecCorr` bijector) are dependent on this PR making it's way through.

src/Bijectors.jl
devmotion: No, I don't want to depend on Transducers either. Unfortunately, it seems we can't just define a curried version
```julia
Base.map(t::Transform) = ... 
```
or
```julia
Base.broadcast(t::Transform) = ...
```
since we would like to use `elementwise` also for functions such as `exp`?

devmotion:
Yeah, let's just make another breaking release if it becomes necessary. I think I checked yesterday and Bijectors only has ~9 direct dependents, so it's not too bad if we iterate and release multiple breaking versions if needed (of course, it would be better to have the optimal design right away but that's completely unrealistic :smile:).

torfjelde:
Bueno! You happy with the current version of the PR then?

devmotion (approved):

PR #229, opened 2022-08-18
Update Github actions to run tests on Julia 1.6

torfjelde:
Now that Julia 1.6 is the latest LTS, we should be running tests on 1.6 rather than 1.3.


devmotion:
IMO we should test *also* the oldest supported version. I've seen too many examples of people accidentally breaking support for older versions when they were not tested.

.github/workflows/AD.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
+          - '1.6'
```
devmotion, line 16:
```suggestion
          - '1.3'
          - '1.6'
```

.github/workflows/Interface.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         version:
-          - '1.3'
+          - '1.6'
```
devmotion, line 17:
```suggestion
          - '1.3'
          - '1.6'
```

devmotion:
BTW I saw the same test error this morning in AbstractDifferentiation: https://github.com/JuliaDiff/AbstractDifferentiation.jl/runs/7891980707?check_suite_focus=true#step:6:400 Maybe some issue with Julia 1.8?

torfjelde:
> IMO we should test also the oldest supported version. I've seen too many examples of people accidentally breaking support for older versions when they were not tested.

Agreed! So IMO we should then bump the supported version to 1.6, no?

devmotion:
My take in similar cases has been so far to not touch the lower bound as long as it does not increase maintenance burden (special handling, debugging etc.) and does not slow down development (preventing new features etc.). I guess here it would be fine to support 1.3 for now. But I am fine either way.

torfjelde:
Gotcha :+1:

I'm going to go with just dropping testing on 1.3 for two reasons:
1. It heavily reduces test times.
2. I'm currently making a lots of changes to Bijectors.jl and will making use of features no longer available on Julia 1.3.

devmotion:
> I'm going to go with just dropping testing on 1.3

OK, in that case it would be good to bump the julia compat to 1.6 right away to avoid tagging a release that unintentionally breaks < 1.6.

PR #230, opened 2022-08-23
Replace `Zygote.@nograd` with `CRC.@non_differentiable`

devmotion:
Fixes deprecation warnings that showed up in Turing.

More stuff should be moved to CRC but intentionally I tried to keep the PR simple.


devmotion:
Test errors on Tracker should be fixed by the new release. Not sure what's holding it back.

devmotion:
#231 seems to unblock Tracker.

PR #240, opened 2023-01-09
Add Enzyme support

devmotion:
I assume there will be quite a few test failures.

(Ideally, we would use some interface package for the AD packages instead of defining them in every package but that's the easiest way to move forward with Enzyme support in Turing...)


torfjelde:
Honestly, we could also just completely remove the AD-related stuff in Bijectors.jl. It's only relevant for the `ADBijector` or whatever it's called, which I don't think is actually used by, well, _anyone_ :shrug:

Issue #220, opened 2022-05-19
Adding a PositiveOrderedBijector


sethaxen:
`OrderedBijector` is only appropriate for bounded real numbers. It would be convenient to also have a `PositiveOrderedBijector` (similar to Stan's `positive_ordered` vector type), which wraps a distribution required to have support `ℝ_{>0}^D`.

Since the exponential is monotonic, the ordering can happen in the latent space, so the bijector is just a composition of `Exp` and `OrderedBijector` in reverse. As a bonus, if `ordered` is passed a distribution with a defined positive bijector, then it could instead return `PositiveOrderedBijector` to avoid incorrect posterior draws.

sethaxen:
More generally, `bijector(d)` for a given continuous univariate distribution `d` is guaranteed to be strictly monotone, so we have 2 cases: the bijector is increasing or decreasing. 

Suppose we have an increasing distribution `Gamma(1, 1)` and a decreasing distribution `-1 * Gamma(1, 1)`:

```julia
julia> dinc = Product(fill(Gamma(1, 1), 10));

julia> ddec = Product(fill(-1 * Gamma(1, 1), 10));

julia> inv(bijector(dinc))(-2:1:2)
5-element Vector{Float64}:
 0.1353352832366127
 0.36787944117144233
 1.0
 2.718281828459045
 7.38905609893065

julia> inv(bijector(ddec))(-2:1:2)
5-element Vector{Float64}:
 -0.1353352832366127
 -0.36787944117144233
 -1.0
 -2.718281828459045
 -7.38905609893065
```

In both cases we can construct an ordered vector of IID elements from these distributions using `OrderedBijector`:

```julia
julia> binc = inv(Bijectors.OrderedBijector()) ∘ bijector(dinc)
Composed{Tuple{Bijectors.TruncatedBijector{1, Float64, Float64}, Inverse{Bijectors.OrderedBijector, 1}}, 1}((Bijectors.TruncatedBijector{1, Float64, Float64}(0.0, Inf), Inverse{Bijectors.OrderedBijector, 1}(Bijectors.OrderedBijector())))

julia> binc(sort(rand(dinc)))
10-element Vector{Float64}:
 -2.022615883470717
 -1.1840408057206144
 -1.1534482444313512
 -2.213221747442945
 -1.7515607559882922
 -0.7470736409590497
 -0.08155306617081076
 -1.9816915175623477
 -1.140842898952562
 -3.1439472077666215

[...]

PR #241, opened 2023-01-30
Throw error if `ordered` unsafe to use

sethaxen:
As noted in https://github.com/TuringLang/Bijectors.jl/issues/220#issuecomment-1409115410, `ordered(d)` does the incorrect thing unless `d` is unconstrained. This PR adds a check that throws an error if this is not the case. Note that examples like this will now error even though they are fine to use with `ordered`, since they don't return a `Identity` bijector:

```julia
julia> bijector(Product(fill(Normal(), 5)))
Bijectors.TruncatedBijector{1, Float64, Float64}(-Inf, Inf)

julia> bijector(Product([Normal(), Cauchy()]))
Bijectors.TruncatedBijector{1, Vector{Float64}, Vector{Float64}}([-Inf, -Inf], [Inf, Inf])
```

### DIFF
--- a/src/bijectors/ordered.jl
+++ b/src/bijectors/ordered.jl
@@ -13,6 +13,8 @@ struct OrderedBijector <: Bijector{1} end
     ordered(d::Distribution)
 
 Return a `Distribution` whose support are ordered vectors, i.e., vectors with increasingly ordered elements.
+
+This transformation is currently only supported for otherwise unconstrained distributions.
 """
 function ordered(d::ContinuousMultivariateDistribution)
     if !isa(bijector(d), Identity)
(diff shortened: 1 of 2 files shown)


torfjelde (approved):
LGTM! Though can you bump the patch version?

Also, if you don't use `Product` but `product_distribution` (`Product` is deprecated), it will actually result in a `DiagNormal`, hence it will work in this particular case. It will still fail for other cases though.

Thanks @sethaxen !

rikhuijzer:
Uhm for the next time, this should have been a breaking release, no? There could be clients running production depending on the old and broken implementation. They now have to adjust or have no output at all, which means this change is breaking.

devmotion:
> depending on the old and broken implementation.

According to semver, bug fixes are non-breaking. The only problem I see here are the now-failing examples in the OP - but it seems they could be fixed by improved definitions of `bijector` for product distributions or an improved check that also supports `TruncatedBijector`s with infinite bounds.

PR #242, opened 2023-01-31
Remove ADBackend and ADBijector

torfjelde:
`ADBijector` is, AFAIK, never used. In addition, it's a very trivial thing to implement on your own.

Because of this, I feel like it's unnecessary to add all the additional complexity that is `ADBackend` just for the sake of maybe one or two use-cases of `ADBijector`. Hence I think we should just drop it as part of the breaking release.


devmotion (commented):
I have never used it, so I'm fine with removing it :stuck_out_tongue: 

I wonder though if it is a reasonably common thing users might want to use, if we should add an example without `ADBijector` to the README and/or to the tests.

torfjelde:
Aight, tests should now be passing.

BUT shouldn't this also technically be another breaking change? I.e. we should bump to 0.12? It's a bit annoying given that we just made a breaking release, but seems like the right course of action IMO. @devmotion @yebai

torfjelde:
Aight, tests should now be passing.

BUT shouldn't this also technically be another breaking change? I.e. we should bump to 0.12? It's a bit annoying given that we just made a breaking release, but seems like the right course of action IMO. @devmotion @yebai

torfjelde:
Aight, tests should now be passing.

BUT shouldn't this also technically be another breaking change? I.e. we should bump to 0.12? It's a bit annoying given that we just made a breaking release, but seems like the right course of action IMO. @devmotion @yebai

devmotion:
Yes, it removes exported and documented functionality, so IMO it deserves another breaking release. I don't think it's too bad as the number of direct dependents is fairly low: https://juliahub.com/ui/Packages/Bijectors/39uFz/0.10.6?page=2

torfjelde:
Version bumped. Should be ready to go once tests pass.

PR #245, opened 2023-02-05
Missing impl of `with_logabsdet_jacobian` for `PDBijector`

torfjelde:
Also added more informative error message in the case of missing `with_logabsdet_jacobian` implementation when calling `transform` and `logabsdetjac`.

### DIFF
--- a/src/bijectors/pd.jl
+++ b/src/bijectors/pd.jl
@@ -30,11 +30,12 @@ end
 
 function logabsdetjac_pdbijector_chol(Xcf::Cholesky)
     # NOTE: Use `UpperTriangular` here because we only need `diag(U)`
-    # and `U` is by default already constructed in `Cholesky`.
-    U = Xcf.U
+    # and `UL` is by default already constructed in `Cholesky`.
+    UL = Xcf.UL
     T = eltype(U)
-    d = size(U, 1)
-    return - sum((d .- (1:d) .+ 2) .* log.(diag(U))) - d * log(T(2))
+    d = size(UL, 1)
+    z = sum((d + 1):(-1):2 .* log.(diag(UL)))
+    return - (z + d * oftype(z, logtwo))
 end
 
 # TODO: Implement explicitly.
--- a/src/bijectors/pd.jl
+++ b/src/bijectors/pd.jl
@@ -34,7 +34,7 @@ function logabsdetjac_pdbijector_chol(Xcf::Cholesky)
     # and `UL` is by default already constructed in `Cholesky`.
     UL = Xcf.UL
     d = size(UL, 1)
-    z = sum((d + 1):(-1):2 .* log.(diag(UL)))
+    z = sum(((d + 1):(-1):2) .* log.(diag(UL)))
     return - (z + d * oftype(z, logtwo))
 end
 
--- a/src/bijectors/pd.jl
+++ b/src/bijectors/pd.jl
@@ -35,7 +35,7 @@ function logabsdetjac_pdbijector_chol(Xcf::Cholesky)
     UL = Xcf.UL
     d = size(UL, 1)
     z = sum(((d + 1):(-1):2) .* log.(diag(UL)))
-    return - (z + d * oftype(z, logtwo))
+    return - (z + d * oftype(z, IrrationalConstants.logtwo))
 end
 
 # TODO: Implement explicitly.
--- a/src/bijectors/pd.jl
+++ b/src/bijectors/pd.jl
@@ -32,7 +32,6 @@ function logabsdetjac_pdbijector_chol(Xcf::Cholesky)
     # NOTE: Use `UpperTriangular` here because we only need `diag(U)`
     # and `UL` is by default already constructed in `Cholesky`.
     UL = Xcf.UL
-    T = eltype(U)
     d = size(UL, 1)
     z = sum((d + 1):(-1):2 .* log.(diag(UL)))
     return - (z + d * oftype(z, logtwo))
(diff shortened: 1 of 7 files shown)


src/bijectors/pd.jl
```diff
@@ -26,12 +25,19 @@ function logabsdetjac(b::PDBijector, X::AbstractMatrix{<:Real})
...
         Xcf = cholesky(X + max(eps(T), eps(T) * norm(X)) * I)
     end
-    return logabsdetjac(b, Xcf)
+    return logabsdetjac_pdbijector_chol(Xcf)
 end
 
-function logabsdetjac(b::PDBijector, Xcf::Cholesky)
+function logabsdetjac_pdbijector_chol(Xcf::Cholesky)
+    # NOTE: Use `UpperTriangular` here because we only need `diag(U)`
+    # and `U` is by default already constructed in `Cholesky`.
```
devmotion, line 33: Not necessarily, could also be `L`.

src/bijectors/pd.jl
```diff
@@ -26,12 +25,19 @@ function logabsdetjac(b::PDBijector, X::AbstractMatrix{<:Real})
...
     U = Xcf.U
```
devmotion, line 34:
```suggestion
    UL = Xcf.UL
```

src/bijectors/pd.jl
```diff
@@ -26,12 +25,19 @@ function logabsdetjac(b::PDBijector, X::AbstractMatrix{<:Real})
...
     T = eltype(U)
```
devmotion, line 35: Not needed, I think, see below:
```suggestion
```

src/bijectors/pd.jl
```diff
@@ -26,12 +25,19 @@ function logabsdetjac(b::PDBijector, X::AbstractMatrix{<:Real})
...
     d = size(U, 1)
```
devmotion, line 36:
```suggestion
    d = size(UL, 1)
```

src/bijectors/pd.jl
devmotion:
```suggestion
    # and `UL` is by default already constructed in `Cholesky`.
```

src/bijectors/pd.jl
```diff
@@ -26,12 +25,19 @@ function logabsdetjac(b::PDBijector, X::AbstractMatrix{<:Real})
...
     return - sum((d .- (1:d) .+ 2) .* log.(diag(U))) - d * log(T(2))
```
devmotion, line 37:
```suggestion
    z = sum((d + 1):(-1):2 .* log.(diag(UL)))
    return - (z + d * oftype(z, logtwo))
```

src/bijectors/pd.jl
torfjelde: By _default_ it's `U` no? But happy to change to `UL`; wasn't aware that was an option :+1:

torfjelde:
Thanks for the comments @devmotion ! Happy with the current one?

src/bijectors/pd.jl
```diff
@@ -26,12 +26,19 @@ function logabsdetjac(b::PDBijector, X::AbstractMatrix{<:Real})
...
+    return logabsdetjac_pdbijector_chol(Xcf)
 end
 
-function logabsdetjac(b::PDBijector, Xcf::Cholesky)
-    U = Xcf.U
-    T = eltype(U)
-    d = size(U, 1)
-    return - sum((d .- (1:d) .+ 2) .* log.(diag(U))) - d * log(T(2))
+function logabsdetjac_pdbijector_chol(Xcf::Cholesky)
+    # NOTE: Use `UpperTriangular` here because we only need `diag(U)`
```
devmotion, line 33: Not completely sure what the intention of the comment is, but I guess this line should be updated as well.
```suggestion
    # NOTE: Use `UL` here because we only need `diag(UL)`
```

src/interface.jl
```diff
@@ -56,7 +56,14 @@ Broadcast.broadcastable(b::Transform) = Ref(b)
 Transform `x` using `b`, treating `x` as a single input.
 """
 transform(f::F, x) where {F<:Function} = f(x)
-transform(t::Transform, x) = first(with_logabsdet_jacobian(t, x))
+function transform(t::Transform, x)
+    res = with_logabsdet_jacobian(t, x)
+    if res isa ChangesOfVariables.NoLogAbsDetJacobian
+        error("`transform` not implemented for $(typeof(b)); implement `transform` and/or `with_logabsdet_jacobian`.")
```
devmotion, line 62: Wouldn't it be more correct to say that logabsdet... is not implemented?At least that's what we check here and why the fallback for transform fails.

src/interface.jl
```diff
@@ -73,7 +80,14 @@ transform!(b, x, y) = copyto!(y, transform(b, x))
 
 Return `log(abs(det(J(b, x))))`, where `J(b, x)` is the jacobian of `b` at `x`.
 """
-logabsdetjac(b, x) = last(with_logabsdet_jacobian(b, x))
+function logabsdetjac(b, x)
+    res = with_logabsdet_jacobian(b, x)
+    if res isa ChangesOfVariables.NoLogAbsDetJacobian
+        error("`logabsdetjac` not implemented for $(typeof(b)); implement `logabsdetjac` and/or `with_logabsdet_jacobian`.")
```
devmotion, line 86: Same here, maybe rather say that with_logabsdet... is used in the fallback for logabsdetjac but not defined?

test/bijectors/utils.jl
```diff
@@ -12,6 +12,8 @@ function test_bijector(
     logjac=nothing,
     test_not_identity=isnothing(y) && isnothing(logjac),
     test_types=false,
+    changes_of_variables_test=true,
```
devmotion, line 15: Maybe could even omit the _test suffix?

devmotion (approved):
Only some minor suggestions, feel free to apply whatever seems reasonable.

PR #246, opened 2023-02-06
Implementation of `VecCorrBijector`

torfjelde:
This PR implements a `VecCorrBijector`, i.e. `CorrBijector` but where the unconstrained space is treated as a `Vector` of the correct size rather than working with a `Matrix` which is actually embedded in a subspace of matrices.

This is particularly when doing stuff like MCMC sampling, where the full matrix representation might give the sampler the false indication that the dimension is `d × d` instead of `(d choose 2)` (as is the case here).

### DIFF
--- a/src/bijectors/corr.jl
+++ b/src/bijectors/corr.jl
@@ -78,19 +78,7 @@ function transform(ib::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})
     return w' * w
 end
 
-function logabsdetjac(::Inverse{CorrBijector}, y::AbstractMatrix{<:Real})
-    K = LinearAlgebra.checksquare(y)
-    
-    result = float(zero(eltype(y)))
-    for j in 2:K, i in 1:(j - 1)
-        @inbounds abs_y_i_j = abs(y[i, j])
-        result += (K - i + 1) * (
-            IrrationalConstants.logtwo - (abs_y_i_j + LogExpFunctions.log1pexp(-2 * abs_y_i_j))
-        )
-    end
-    
-    return result
-end
+logabsdetjac(::Inverse{CorrBijector}, Y::AbstractMatrix{<:Real}) = _logabsdetjac_chol_lkj(Y)
 function logabsdetjac(b::CorrBijector, X::AbstractMatrix{<:Real})
     #=
     It may be more efficient if we can use un-contraint value to prevent call of b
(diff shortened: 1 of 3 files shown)


src/bijectors/corr.jl
```diff
@@ -75,51 +75,158 @@ end
...
+ -0.29813769428942216
+
+julia> inverse(b)(y) ≈ X  # (✓) Round-trip through `b` and its inverse.
+true
+"""
+struct VecCorrBijector <: Bijector end
+with_logabsdet_jacobian(b::VecCorrBijector, x) = transform(b, x), logabsdetjac(b, x)
+
+function transform(::VecCorrBijector, X::AbstractMatrix{<:Real})
+    w = upper_triangular(parent(cholesky(X).U))
```
torfjelde, line 212: I hate this:(

harisorgn: Could we not have `w = cholesky(X).U` and work with a `w::UpperTriangular` instead of a dense matrix? Tried it locally, no real gain in performance though.

torfjelde: > Could we not have w = cholesky(X).U and work with a w::UpperTriangular instead of a dense matrix? 

Yeah, but this won't work with some of the AD backends (I know, it's super-annoying...). If you have a look at our compat-code for ReverseDiff (I believe), I think you'll see that we have to do some custom stuff to compute the pullback.

> Tried it locally, no real gain in performance though.

I don't think we'd expect it to because internally we're iterating over the relevant elements of the matrix anyways, i.e. we're not gaining anything by telling the rest of the computational path that we're actually working on a lower-triangular matrix because it already assumes the given matrix is lower-triangular.

devmotion: I would guess it's mainly for efficiency reasons. But it's difficult to say if there are other implications as well, e.g., regarding AD. An alternative to `.UL` would be something like what's used in PDMats: https://github.com/JuliaStats/PDMats.jl/blob/fff131e11e23403931a42f5bfb3384f0d2b114c9/src/chol.jl#L6-L11 That should also be quite efficient and you could continue working with upper-triangular matrices.

torfjelde: > I would guess it's mainly for efficiency reasons. 

https://github.com/JuliaLang/julia/blob/db7971f49912d1abba703345ca6eb43249607f32/stdlib/LinearAlgebra/src/cholesky.jl#L515-L527

> But it's difficult to say if there are other implications as well, e.g., regarding AD. 

Hmm fair, though IMO something like this seems like it would be a bug with the AD package, no?

> An alternative to .UL would be something like what's used in PDMats: https://github.com/JuliaStats/PDMats.jl/blob/fff131e11e23403931a42f5bfb3384f0d2b114c9/src/chol.jl#L6-L11 That should also be quite efficient and you could continue working with upper-triangular matrices.

Me and @harisorgn were just having a chat and we're thinking of replacing `upper_triangular(parent(cholesky(X).U))` with

```julia
cholesky_upper(x) = upper_triangular(parent(cholesky(X).U))
cholesky_lower(x) = lower_triangular(parent(cholesky(X).L))
```

to make it less likely that we forget or mess up somewhere.

But we can make it

```julia
cholesky_upper(x) = upper_triangular(parent(PDMats.chol_upper(cholesky(X))))
cholesky_lower(x) = lower_triangular(parent(PDMats.chol_lower(cholesky(X))))
```

But are you sure there's not a good reason for why the default is `copy`? Of course it's more mem-intensive, but will stuff like `LowerTriangular(U')` lead to slower computation paths (since you're now working with `adjoint(U)` rather than something that is actually lower-triangular)? E.g. indexing `adjoin(U)` surely involves more computations than indexing `copy(adjoint(U))`.

harisorgn:
I rebased main here to resolve conflicts around reshape and its tests, hope it wasn't overkill in terms of tracking changes.

harisorgn:
Test failure is the `ForwardDiff` issue I mentioned #253 in [this comment](https://github.com/TuringLang/Bijectors.jl/pull/253#issuecomment-1521469558).

harisorgn:
> but always wrap the matrix in a Symmetric, Hermitian, etc. wrapper first. This will dispatch to optimized methods, avoid checks such as ishermitian, and also yield different paths and results in AD.

Thanks David! This seems to work for our issue. The interface tests for v1.6 now fail on the `Hermitian` constructor for `ReverseDiff`. Seems that an rrule for the constructor is accounted for in latest but not in v1.6 (though I couldn't find it explicitly)?

torfjelde:
I messed up and merged the formatting PR into master before this :facepalm: So the result was quite a lot of conflicts. I've resolved them now though :+1:

devmotion:
Regarding `cholesky`/`Hermitian`: I assume it works on Julia >= 1.8 due to https://github.com/JuliaLang/julia/pull/44076.

harisorgn:
>Regarding cholesky/Hermitian: I assume it works on Julia >= 1.8 due to https://github.com/JuliaLang/julia/pull/44076.

Ah yes, thanks once again! Sorry, misread the stack trace on my phone and thought the issue was the `Hermitian` constructor 😅 

How should we handle this then? The `Hermitian` wrapper before `cholesky` solves the `ForwardDiff` issue with the tiny numerical mismatch, but creates an instance of `<: Hermitian{....TrackedArray....}` that `cholesky` can't handle in v1.6 .

torfjelde:
One straight-forward (but hacky) solution is of course:
```julia
cholesky_factor(X::ReverseDiff.TrackedArray) = cholesky_factor(cholesky(X))
```

Also, though I don't see another fix, I'm uncertain if `Hermitian` is the 100% correct solution since during sampling we can easily run into matrices which end up not being hermitian due to numerical issues. If this happens, it seems to me that we'd want to error rather than just continue, no?

harisorgn:
>One straight-forward (but hacky) solution is of course

Was just checking in case I wasn't seeing something less hacky, but let's go with this now 👍 . I've added this method in `src/compat/reversediff.jl` to avoid a `ReverseDiff` dependency outside that submodule, a bit annoying though as it is not grouped with the rest of the methods.

>we can easily run into matrices which end up not being hermitian due to numerical issues. If this happens, it seems to me that we'd want to error rather than just continue, no?

Should we have something like an `ishermitian` check but with tolerances? The `ForwardDiff` issue was numerical but tiny as far as I tested it.

src/compat/reversediff.jl
```diff
@@ -208,9 +227,20 @@ end
...
 # NOTE: Probably doesn't work in complete generality.
 wrap_chainrules_output(x) = x
 wrap_chainrules_output(x::ChainRulesCore.AbstractZero) = nothing
 wrap_chainrules_output(x::Tuple) = map(wrap_chainrules_output, x)
 
+# HACK: To make it work for julia v1.6 .
+# This dispatch does not wrap X in Hermitian before calling cholesky. 
+# cholesky does not work with AbstractMatrix in julia v1.6,
+# and it would error with Hermitian{ReverseDiff.TrackedArray}.
+cholesky_factor(X::ReverseDiff.TrackedArray) = cholesky_factor(cholesky(X))
```
devmotion, line 244: I'd suggest figuring out the Julia version from which on the linked PR was available (see, e.g., Compat.jl for instructions) and then wrap this fix in a `if VERSION < v"..."` block. This would unify the AD backends on more recent versions at least.

test/bijectors/corr.jl
```diff
@@ -0,0 +1,65 @@
...
+
+        @test xinv ≈ xvecinv
+
+        # And finally that the `logabsdetjac` is the same.
+        @test logabsdetjac(bvec, x) ≈ logabsdetjac(b, x)
+
+        # NOTE: `CorrBijector` technically isn't bijective, and so the default `getjacobian`
+        # used in the ChangesOfVariables.jl tests will fail as the jacobian will have determinant 0.
+        # Hence, we disable those tests.
+        test_bijector(b, x; test_not_identity=d != 1, changes_of_variables_test=false)
```
harisorgn, line 32: The `InverseFunctions.test_inverse` that is called in here fails on some calls, but passes on others, even with the same sample `x`. CI on previous commits of this PR failed because of it. Seems weird, not sure why.

harisorgn:
Do we want to keep `CorrBijector` in general? Will it have a purpose after this PR?

torfjelde:
> Do we want to keep CorrBijector in general? Will it have a purpose after this PR?

Good question. I'd say keep it for now + make an issue.

test/bijectors/corr.jl
harisorgn: There was no RNG between these lines, it was after sampling from `LKJ`. I believe the issue was numerical as there was no explicit zero-filling in the old Matrix-version link function, which caused mismatches. I've added zero-filling and tested it a bunch locally, should be fine 🤞

PR #253, opened 2023-04-06
Use AbstractVector in LKJ and LKJCholesky bijectors

harisorgn:
Expands on #246 .

Use `::AbstractVector` in `VecCorrBijector` operations, so we won't need to `transform` to `::AbstractMatrix` and back.

Add bijector for `LKJCholesky`. I believe this was missing and in practice it is the more efficient alternative when working with correlation matrices (avoids Cholesky decompositions on every call).
In `LKJCholesky` there is control over the returned factor (`'U' -> UpperTriangular` or `'L' -> LowerTriangular`). I was wondering whether we want to respect the factor choice and always return the same triangular factor. If yes, we can use `VecTriuBijector` and `VecTrilBijector` to retain information about the original factor in `LKJCholesky` and return it. If no, we can always work with one type, e.g. `UpperTriangular`.

TO DO :

- ~~Add `ChainRulesCore.rrule`s for all link functions that work on `::AbstractVector`, defined in this PR. I have only added one rule for the forward link function, but `ChainRulesTestUtils.test_rrule` complains about type instability and value mismatch. When comparing the values returned by the pullback inside the closure of `rrule` against [the one defined for Zygote](https://github.com/TuringLang/Bijectors.jl/blob/637df3a5a1310149f3a57125ef864494972f7730/src/compat/zygote.jl#L244) I'm getting the same output though. I will have more of a look next week.~~
- ~~Document how I ended up with [`_logabsdetjac_inv_chol`](https://github.com/TuringLang/Bijectors.jl/blob/7f5d0fc747f4ccf06b6f5df5c241d08e61d86939/src/bijectors/corr.jl#L419), so it can be verified.~~ This was based on the Stan manual pages for [correlation matrices](https://mc-stan.org/docs/reference-manual/correlation-matrix-transform.html#absolute-jacobian-determinant-of-the-correlation-matrix-inverse-transform) and [Cholesky factors of correlation matrices](https://mc-stan.org/docs/reference-manual/cholesky-factors-of-correlation-matrices-1).
EDIT 2: I have not documented the formula derivation but added a test for it that passes.
- ~~Remove this dispatch https://github.com/TuringLang/Bijectors.jl/blob/7f5d0fc747f4ccf06b6f5df5c241d08e61d86939/src/bijectors/corr.jl#L320
and use `transpose(W::UpperTriangular)`~~.

Related to the second point, right above, in general it would be nice if we could test these analytical formulas for logabsdetjac derived by hand. I played around with it a bit, but couldn't come up with something.
EDIT : This can be done using AD. I see there is something already implemented along these lines in [test/transform.jl](https://github.com/TuringLang/Bijectors.jl/blob/637df3a5a1310149f3a57125ef864494972f7730/test/transform.jl#L184-L187), just needs some tweaking.

cc @torfjelde if you want to have a look already


src/bijectors/corr.jl
```diff
@@ -205,30 +222,34 @@ julia> y = b(X)  # Transform to unconstrained vector representation.
...
-end
+struct VecTriuBijector <: AbstractVecCorrBijector end
 
-function transform(::Inverse{VecCorrBijector}, y::AbstractVector{<:Real})
-    Y = vec_to_triu1(y)
-    w = _inv_link_chol_lkj(Y)
-    return pd_from_upper(w)
+function transform(::Inverse{VecTriuBijector}, y::AbstractVector{<:Real})
+    # This constructor is compatible with Julia v1.6
+    # for later versions Cholesky(::UpperTriangular) works
```
torfjelde, line 234: This is referring to the `Cholesky` constructor, right?

harisorgn: yep, will reword it

src/bijectors/corr.jl
```diff
@@ -205,30 +222,34 @@ julia> y = b(X)  # Transform to unconstrained vector representation.
...
-function logabsdetjac(::Inverse{VecCorrBijector}, y::AbstractVector{<:Real})
-    return _logabsdetjac_chol_lkj(vec_to_triu1(y))
+logabsdetjac(::Inverse{VecTriuBijector}, y::AbstractVector{<:Real}) = _logabsdetjac_inv_chol(y)
+
+struct VecTrilBijector <: AbstractVecCorrBijector end
+
+function transform(::Inverse{VecTrilBijector}, y::AbstractVector{<:Real})
+    # This constructor is compatible with Julia v1.6
+    # for later versions Cholesky(::LowerTriangular) works
+    L = LowerTriangular(_transpose_matrix(_inv_link_chol_lkj(y)))
```
torfjelde, line 246:
```suggestion
    # HACK: Need to make materialize the transposed matrix to avoid numerical instabilities.
    # If we don't, the return-type can be both `Matrix` and `Transposed`.
    L = LowerTriangular(Matrix(transpose(_inv_link_chol_lkj(y))))
```

src/transformed_distribution.jl
```diff
@@ -77,7 +78,8 @@ bijector(d::LowerboundedDistribution) = bijector_lowerbounded(d)
 bijector(d::PDMatDistribution) = PDBijector()
 bijector(d::MatrixBeta) = PDBijector()
 
-bijector(d::LKJ) = CorrBijector()
+bijector(d::LKJ) = VecCorrBijector()
+bijector(d::LKJCholesky) = d.uplo === 'L' ? VecTrilBijector() : VecTriuBijector()
```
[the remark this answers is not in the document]
torfjelde: That is, we make it
```suggestion
bijector(d::LKJCholesky) = VecCorrBijector(d.uplo)
```

src/bijectors/corr.jl
```diff
@@ -173,6 +173,23 @@ end
...
 inverse(::typeof(vec_to_triu1)) = triu1_to_vec
 
+function vec_to_triu1_row_index(idx)
+    # Assumes that vector was saved in a column-major order
+    # and that vector is one-based indexed.
+    M = _triu1_dim_from_length(idx - 1)
+    return idx - (M*(M-1) ÷ 2)
+end
+
+abstract type AbstractVecCorrBijector <: Bijector end
```
torfjelde, line 183: With this approach, `bijector` becomes type-unstable (see comment on this further below).

We should probably just make `VecCorrBijector` hold the `uplo` of the `Cholesky`, and add an if-statement to the `transform`, etc. This way we can avoid type-instabilities.

torfjelde (changes requested):
It overall looks great:) But I think maybe we should just merge the `VecTriuBijector` and `VecTrilBijector` (+  their abstract type) into a single `VecCorrBijector`. This way we avoid type-instabilities in `bijector` + we reduce the number of types.

harisorgn:
> Have another look and If you think all issues were addressed, we can merge.

Actually let's wait, as the AD tests I've just added on the roundtrip transformation fail, will have a look.

harisorgn:
Looked more into `test_ad` using the roundtrip inverse-and-then-forward transformation :

Test is failing on `cholesky(Matrix{ForwardDiff.Dual...})` . I see [there is no `frule` defined for `cholesky`](https://github.com/JuliaDiff/ChainRules.jl/blob/65b922004165b3fe4d0db001f5c4e99d8774607b/src/rulesets/LinearAlgebra/factorization.jl), not sure if something `ForwardDiff` specific exists elsewhere.

Unsure about the `Tracker` error. Is the plan to keep supporting `Tracker` in general?

`Zygote` is returning `nothing` gradients, that's its "hard zero" IIUC. Not sure if it has to do with my usage of `getproperty` on `Cholesky`, `UpperTriangular` and `LowerTriangular` types.

`ReverseDiff` is passing all tests after fixing the rule for `pd_from_upper`.

All these tests are for AD through `transform` which IIUC are not relevant for Turing.jl usage. Since this PR is adding on the previous one that aims to get LKJ priors working for Turing, it might be worth merging and tackling the remaining AD issues in a future PR?

torfjelde:
Regarding the test-failures, it's a bit strange. 

This seems related: https://github.com/JuliaDiff/ForwardDiff.jl/issues/606.

But I thought this was fixed because they pulled 0.10.33 after the discussion there, and deferred the breaking changes to 0.11. The tests are running on 0.10.35 so I don't get why we're seeing this :confused:

Think this needs a bit further inspection.

And we do actually need to AD through the `transform` in some places, e.g. ADVI.

(Btw, I'm not done with my review, will continue later)

harisorgn:
>I think the Zygote changes are more straightforward unless I'm missing something.

It is for the case of `LKJ` (changing `X.UL` to  `X.U` works) but not for `LKJCholesky`. In the latter case, we construct a `Cholesky` during the inverse transform, as this is the support of the distribution. I'm guessing the `Cholesky` constructor needs an `rrule` for `Zygote` to work.

harisorgn:
`Zygote` is fixed. It was more straightforward than writing new `rrule`s, just passing a `X::Matrix` instead of `X::UpperTriangular` or `X::LowerTriangular` to `Cholesky` and avoid doing `X.data`.

`ForwardDiff` passed twice on the latest commit, but I changed nothing to fix it. Probably has to do with the stochastic nature of the numerical error, like the example above.

src/chainrules.jl
```diff
@@ -156,5 +156,170 @@ function ChainRulesCore.rrule(::typeof(_transform_inverse_ordered), x::AbstractM
...
+    end
+
+    return W, pullback_inv_link_chol_lkj
+end
+
+function ChainRulesCore.rrule(::typeof(pd_from_upper), X::AbstractMatrix)
+    return UpperTriangular(X)' * UpperTriangular(X), Δ -> begin
+        Xu = UpperTriangular(X)
+        return ChainRulesCore.NoTangent(), UpperTriangular(Xu * Δ + Xu * Δ')
+    end
```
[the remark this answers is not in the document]
torfjelde: Also, maybe add a rrule test? That would have caught the missing `unthunk`.

src/bijectors/corr.jl
```diff
@@ -173,11 +173,30 @@ end
...
     VecCorrBijector <: Bijector
 
-Similar to `CorrBijector`, but correlation matrix to a vector,
-and its inverse transforms vector to a correlation matrix.
+A bijector to transform either a correlation matrix or a Cholesky factor of a correlation matrix
+to an unconstrained vector. 
+
+# Fields
+- mode :`Symbol`. Controls the inverse tranformation :
+    - if `mode === :C` returns a correlation matrix
```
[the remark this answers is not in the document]
torfjelde: That is, make the cholesky version into a separate type, e.g. `VecCholCorrBijector`. This will avoid the type-instabilities + moves the conditional handling you have in some functions into multiple dispatch instead.

torfjelde (changes requested):
_Almost_ there! But I think the cholesky-version should just be its own struct so we avoid the type-stabilities.

Otherwise it's looking pretty dank!

And I'll have a look at the ForwardDiff issue.

src/bijectors/corr.jl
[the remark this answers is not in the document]
torfjelde: @harisorgn Any updates on this?:)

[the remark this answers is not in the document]
harisorgn: I was completely off last week. Agree with splitting/specialising the structures, will implement it this week!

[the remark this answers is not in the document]
torfjelde: Ah, no worries! Sweet!

src/chainrules.jl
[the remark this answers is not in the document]
harisorgn: The thing is I was testing the rrules locally as I was adding them and this was passing. Probably `unthunk`ing would be needed if it's part of multiple function calls that get differentiated? I am adding it anyway.

harisorgn:
@torfjelde implemented your suggestions, thanks for the feedback again : )

I couldn't locally reproduce the `DomainError` that comes up in the AD test.

Also disregard my previous confusion about reproducing the `ForwardDiff` numerical error. I was restarting an interface test that wasn't hitting it, hence it was passing. When the right interface test of CI was run, test failed as it fails locally (see comments above). 

So there are still these two errors, plus the `stack` related one that is addressed in another PR here. 

(Apologies for the format, only have phone access for now)

src/bijectors/corr.jl
```diff
@@ -206,29 +215,98 @@ julia> inverse(b)(y) ≈ X  # (✓) Round-trip through `b` and its inverse.
...
+
+function transform(b::Inverse{VecCholeskyBijector}, y::AbstractVector{<:Real})
+    if b.orig.mode === :U
+        # This Cholesky constructor is compatible with Julia v1.6
+        # for later versions Cholesky(::UpperTriangular) works
+        return Cholesky(_inv_link_chol_lkj(y), 'U', 0)
+    else # No need to check for === :L, as it is checked in the VecCholeskyBijector constructor.
+        # HACK: Need to make materialize the transposed matrix to avoid numerical instabilities.
+        # If we don't, the return-type can be both `Matrix` and `Transposed`.
+        return Cholesky(Matrix(transpose(_inv_link_chol_lkj(y))), 'L', 0)
```
torfjelde, line 304: Just realized we should instead do:
```suggestion
        return Cholesky(permutedims(_inv_link_chol_lkj(y), (2, 1)), 'L', 0)
```

torfjelde: `permutedims` makes it concrete so we don't have to do the additional `Matrix`. Quick benchmarking showed that it can be quite a bit faster.

test/transform.jl
```diff
@@ -182,7 +188,23 @@ end
 
     upperinds = [LinearIndices(size(x))[I] for I in CartesianIndices(size(x)) if I[2] > I[1]]
     J = ForwardDiff.jacobian(x->link(dist, x), x)
-    J = J[upperinds, upperinds]
+    J = J[:, upperinds]
```
torfjelde, line 191: What was this for again? Sorry, we might have discussed this before.

test/transform.jl
```diff
@@ -38,7 +38,13 @@ function single_sample_tests(dist)
 
     # Check that invlink is inverse of link.
     x = rand(dist)
-    @test @inferred(invlink(dist, link(dist, copy(x)))) ≈ x atol=1e-9
+
+    if dist isa LKJCholesky
+        x_inv = @inferred Union{Cholesky{Float64, Matrix{Float64}}, Matrix{Float64}} invlink(dist, link(dist, copy(x)))
```
torfjelde, line 43: Shouldn't this be concrete now rather than `Union`? I.e.
```suggestion
        x_inv = @inferred invlink(dist, link(dist, copy(x)))
```

torfjelde (changes requested):
Great stuff @harisorgn :) Really close now!

I had a super-quick look, and made some very minor comments + changes. Once those are addressed, I think we should be good go!

Again, awesome work; I imagine this isn't the most fun PR to work on, so appreciate you seeing this through :heart:

harisorgn:
@torfjelde accidental merge, sorry, was setting up git in a new machine 😅 . Please revert it and I'll implement the last changes.

test/transform.jl
harisorgn: Don't think we have : ) . It's because the output of `dist` is an `AbstractVector` now, so the indices of upper triagular elements don't apply anymore. In this test, x is 3x3 matrix, `link(dist, x)` is length 3 vector, the Jacobian is then a 3x9 matrix, and we are keeping all output elements (as they are all relevant now) and only the `upperinds` of the input elements.

test/transform.jl
harisorgn: Yes, good catch! I missed this one.

torfjelde:
Is it maybe easier if you just take over the other PR?:) #246

PR #256, opened 2023-05-15
Rename `stack` to `stack_transforms`

torfjelde:
On Julia 1.9 we have `Base.stack`, which means that our exported `Bijectors.stack` has a name-clash, and overloading it in a useful way is too dangerous.

So I've renamed it. I also don't think this is being used anywhere but internally.


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.12.3"
+version = "0.12.4"
```
torfjelde, line 3: Technically this should be a breaking release, but `Bijectors.stack` is used only in DPPL tests where it's qualified by module-name anyways.

devmotion (commented):
Couldn't we just extend `Base.stack`? And use Compat to keep support for older Julia versions?

devmotion:
Ah, I see :+1: Then another question is, what is the advantage of having a separate `stack`/`stack_transforms` function over an additional constructor `Stacked(bs...) = Stacked(bs)`?

torfjelde:
> Ah, I see +1 Then another question is, what is the advantage of having a separate stack/stack_transforms function over an additional constructor Stacked(bs...) = Stacked(bs)?

I think the original idea was that it could specialize, e.g. `stack(::Logit{0}...)` could be implemented to result in `Logit{1}`, but now that we don't have these anymore, it's less clear what the advantage is.

src/bijectors/stacked.jl
```diff
@@ -47,7 +47,9 @@ isclosedform(b::Stacked) = all(isclosedform, b.bs)
 
 isinvertible(b::Stacked) = all(isinvertible, b.bs)
 
-stack(bs...) = Stacked(bs)
+
+@deprecate stack stack_transforms
```
devmotion, line 51: Doesn't this mean users still have a name-clash when loading Bijectors?

Also generally, I think we should keep the existing tests when deprecating to make sure that the deprecated function still works.

devmotion (commented):
As long as there's no clear use case for a separate `stack`/`stack_transform` function I am leaning towards just defining an additional `Stacked` constructor. `Stacked` is already exported, so it would also keep the API simpler.

src/bijectors/stacked.jl
torfjelde: > Doesn't this mean users still have a name-clash when loading Bijectors?

Oh crap, I forgot that `@deprecate` exports by default. Should we not export? (that was my original intention)

> Also generally, I think we should keep the existing tests when deprecating to make sure that the deprecated function still works.

Good point; will do!

src/Bijectors.jl
```diff
@@ -66,7 +66,7 @@ export  TransformDistribution,
         Bijector,
         Inverse,
         Stacked,
-        stack,
+        stack_transforms,
```
torfjelde, line 69:
```suggestion
```

test/interface.jl
```diff
@@ -222,7 +219,7 @@ end
 
     # value-test
     x = ones(3)
-    sb = @inferred stack(elementwise(exp), elementwise(log), Shift(5.0))
+    sb = @inferred stack_transforms(elementwise(exp), elementwise(log), Shift(5.0))
```
torfjelde, line 222:
```suggestion
    sb = @inferred Stacked(elementwise(exp), elementwise(log), Shift(5.0))
```

test/norm_flows.jl
```diff
@@ -117,10 +117,10 @@ end
     x = rand(d) .+ 10
     y = b(x)
 
-    sb = stack(b1, b1)
+    sb = stack_transforms(b1, b1)
```
torfjelde, line 120:
```suggestion
    sb = Stacked(b1, b1)
```

test/norm_flows.jl
```diff
@@ -117,10 +117,10 @@ end
...
     @test all((sb ∘ b)(x) .≤ 1.0)
 
-    sb = stack(b1, b2)
+    sb = stack_transforms(b1, b2)
```
torfjelde, line 123:
```suggestion
    sb = Stacked(b1, b2)
```

src/bijectors/stacked.jl
```diff
@@ -21,12 +21,13 @@ b = stack(b1, b2)
...
 ```
 """
-struct Stacked{Bs, Rs} <: Transform
+struct Stacked{Bs, Rs<:Union{Tuple,AbstractArray}} <: Transform
     bs::Bs
     ranges::Rs
 end
 Stacked(bs::Tuple) = Stacked(bs, ntuple(i -> i:i, length(bs)))
 Stacked(bs::AbstractArray) = Stacked(bs, [i:i for i in 1:length(bs)])
+Stacked(bs...) = Stacked(bs, ntuple(i -> i:i, length(bs)))
```
devmotion, line 30: For a follow-up issue/PR/discussion: I think it would be nice if `ranges` to a collection of integers to avoid these one-element ranges.

src/bijectors/stacked.jl
```diff
@@ -47,7 +48,7 @@ isclosedform(b::Stacked) = all(isclosedform, b.bs)
 
 isinvertible(b::Stacked) = all(isinvertible, b.bs)
 
-stack(bs...) = Stacked(bs)
+@deprecate stack(bs...) Stacked(bs...) false
```
devmotion, line 51: I'm starting to wonder - is this deprecation useful at all if `Bijectors.stack` is not exported? Is that helpful for DynamicPPL?

devmotion: Also `stack` is removed from the tests :shrug: So maybe just remove it completely?
```suggestion
```

src/bijectors/stacked.jl
torfjelde: I'm not entirely certain I understand what you mean :thinking:

devmotion: That it would be nice to just use `ntuple(identity, length(bs))` or maybe even `Base.OneTo(length(bs))` or similar ranges instead of `ntuple(i -> i:i, length(bs))`.

torfjelde: Ah yes, I agree. `Stacked` is pretty bad imo, so this is just one out of several improvements we need to make to it.

torfjelde:
@devmotion You good with this? Would be nice to get this merged so tests will pass again and we merge other PRs that are in the pipeline.

devmotion (approved):

PR #257, opened 2023-05-15
Added `Reshape`

torfjelde:
Added a bijector for `Reshape`. This also adds support for `Distributions.ReshapedDistribution`.


torfjelde:
Tests will fail on `julia-latest` until #256 has been merged.

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.12.3"
+version = "0.12.4"
```
devmotion, line 3: I think we need to update the Distributions compat. AFAIK I (?) added ReshapedDistribution relatively recently.

torfjelde: Ah yes, good catch!

Issue #228, opened 2022-08-05
CorrBijector makes posterior improper


sethaxen:
An $n \times n$ correlation matrix has ${n \choose 2} = \frac{n (n-1)}{2}$ degrees of freedom. This is the same as the number of elements in a strict upper triangular $n \times n$ matrix. The `CorrBijector` works by mapping from the correlation matrix first to its unique upper Cholesky factor and then to a strictly upper triangular matrix of unconstrained entries.

The trouble is that in the unconstrained space, we now have $n \times n$ parameters, of which ${n+1 \choose 2} = \frac{n(n+1)}{2}$ have no impact on the log density. These extra parameters have an implicit improper uniform prior on the reals, which makes the posterior distribution in unconstrained space improper. Because these parameters have infinite variance, during adaptation, HMC will learn this, and they will explode in value. I don't know if this will have any negative impact on sampling.

In this demo, we're sampling the uniform distribution on the correlation matrices.
```julia
julia> using Turing, Random

julia> @model function model(n, η)
           R ~ LKJ(n, η)
       end;

julia> mod = model(3, 2.0);

julia> Random.seed!(50);

julia> chns = sample(mod, NUTS(0.99), 1_000; save_state=true)
┌ Info: Found initial step size
└   ϵ = 0.8
┌ Warning: The current proposal will be rejected due to numerical error(s).
│   isfinite.((θ, r, ℓπ, ℓκ)) = (true, false, false, false)
└ @ AdvancedHMC ~/.julia/packages/AdvancedHMC/51xgc/src/hamiltonian.jl:47
Sampling 100%|█████████████████████████████████████████████████████████████████████████| Time: 0:00:11
Chains MCMC chain (1000×21×1 Array{Float64, 3}):

Iterations        = 501:1:1500
Number of chains  = 1
Samples per chain = 1000
Wall duration     = 12.02 seconds
Compute duration  = 12.02 seconds
parameters        = R[1,1], R[2,1], R[3,1], R[1,2], R[2,2], R[3,2], R[1,3], R[2,3], R[3,3]
internals         = lp, n_steps, is_accept, acceptance_rate, log_density, hamiltonian_energy, hamiltonian_energy_error, max_hamiltonian_energy_error, tree_depth, numerical_error, step_size, nom_step_size

Summary Statistics
  parameters      mean       std   naive_se      mcse         ess      rhat   ess_per_sec 
[...]

PR #263, opened 2023-06-05
Make SimplexBijector actually bijective

sethaxen:
Similar to #228, currently the `SimplexBijector` makes transformed distributions improper. A demo from slack:

```julia
julia> using Turing

julia> @model function foo()
           d = Dirichlet(ones(2))
           x ~ filldist(Flat(), length(d))
           Turing.@addlogprob! logpdf(transformed(d), x)
           y = transform(inverse(bijector(d)), x)
           return (; y)
       end;

julia> chns = sample(foo(), NUTS(500, 0.8), MCMCThreads(), 1_000, 4)
┌ Info: Found initial step size
└   ϵ = 3.6
┌ Info: Found initial step size
└   ϵ = 3.6
┌ Info: Found initial step size
└   ϵ = 3.6
┌ Info: Found initial step size
└   ϵ = 12.8
Sampling (4 threads) 100%|█████████████████████████████████████████████████████████████| Time: 0:00:06
Chains MCMC chain (1000×14×4 Array{Float64, 3}):

Iterations        = 501:1:1500
Number of chains  = 4
Samples per chain = 1000
Wall duration     = 5.58 seconds
Compute duration  = 20.62 seconds
parameters        = x[1], x[2]
internals         = lp, n_steps, is_accept, acceptance_rate, log_density, hamiltonian_energy, hamiltonian_energy_error, max_hamiltonian_energy_error, tree_depth, numerical_error, step_size, nom_step_size

Summary Statistics
  parameters                     mean                      std                     mcse    ess_bulk    ess_ta ⋯
      Symbol                  Float64                  Float64                  Float64     Float64     Float ⋯

        x[1]                   0.0359                   1.7472                   0.0299   3450.8940   2517.15 ⋯
        x[2]   -1255806938226913.2500   38456907341159336.0000   12941276533029076.0000      9.0031     15.55 ⋯
                                                                                              3 columns omitted

Quantiles
  parameters                      2.5%                     25.0%                     50.0%                    ⋯
      Symbol                   Float64                   Float64                   Float64                  F ⋯

        x[1]                   -3.4263                   -1.0549                    0.0211                    ⋯
        x[2]   -75987690476946768.0000   -23883185857795536.0000   -10795918685407138.0000   2113099877552726 ⋯

julia> yvals = permutedims(stack(first.(generated_quantities(foo(), chns))), (2, 3, 1));^C

julia> ess(yvals)
2-element Vector{Float64}:
 3450.8940281885793
 3450.8940281885766

julia> dropdims(mean(yvals; dims=(1, 2)); dims=(1, 2))
2-element Vector{Float64}:
 0.5055473060755683
 0.49445269392443175
```

This PR changes `SimplexBijector` to transform a `K`-vector to a `K-1`-vector. Since the `proj` type entry in `SimplexBijector` only impacted the extra `K`th entry of the unconstrained vector, this type entry has been removed. Since the Jacobian is now non-square, triangular return types are no longer used. As a result, the change is marked as breaking.


sethaxen:
Currently tests fails due to these lines, which seem to assume inputs and outputs are the same size
- https://github.com/TuringLang/Bijectors.jl/blob/fc155c9f1fdee3e5bfcb8b2023f37a43606b73e1/src/bijectors/stacked.jl#L77
- https://github.com/TuringLang/Bijectors.jl/blob/fc155c9f1fdee3e5bfcb8b2023f37a43606b73e1/src/transformed_distribution.jl#L95

On Slack, @torfjelde confirmed that these should be fixed.

src/bijectors/simplex.jl
```diff
@@ -213,12 +200,10 @@ function simplex_logabsdetjac_gradient(x::AbstractMatrix)
...
 end
 
-function simplex_link_jacobian(
-    x::AbstractVector{T}, ::Val{proj}=Val(true)
-) where {T<:Real,proj}
+function simplex_link_jacobian(x::AbstractVector{T}) where {T<:Real}
     K = length(x)
     @assert K > 1 "x needs to be of length greater than 1"
-    dydxt = similar(x, length(x), length(x))
+    dydxt = similar(x, K, K - 1)
```
devmotion, line 206: Maybe merge it with the next line while you're at it to make it clearer that `dydx` is initialized with zeros (I had missed that initially when reading the code)?
```suggestion
    dydxt = fill!(similar(x, K, K - 1), 0)
```

devmotion: (BTW the `similar` + `AbstractVector` argument wrongly suggests that the code will work for arbitrary inputs - but the loops below assume that everything uses 1-based indices)

src/bijectors/simplex.jl
```diff
@@ -315,12 +294,10 @@ function add_simplex_link_adjoint!(
...
-function simplex_invlink_jacobian(
-    y::AbstractVector{T}, ::Val{proj}=Val(true)
-) where {T<:Real,proj}
-    K = length(y)
+function simplex_invlink_jacobian(y::AbstractVector{T}) where {T<:Real}
+    K = length(y) + 1
     @assert K > 1 "x needs to be of length greater than 1"
-    dxdy = similar(y, length(y), length(y))
+    dxdy = similar(y, K, K - 1)
     @inbounds dxdy .= 0
```
devmotion, line 301: Same here, maybe use
```suggestion
    dxdy = fill!(similar(y, K, K - 1), 0)
```
?

src/bijectors/simplex.jl
```diff
@@ -213,12 +200,10 @@ function simplex_logabsdetjac_gradient(x::AbstractMatrix)
...
 
-function simplex_link_jacobian(
-    x::AbstractVector{T}, ::Val{proj}=Val(true)
-) where {T<:Real,proj}
+function simplex_link_jacobian(x::AbstractVector{T}) where {T<:Real}
     K = length(x)
     @assert K > 1 "x needs to be of length greater than 1"
-    dydxt = similar(x, length(x), length(x))
+    dydxt = similar(x, K, K - 1)
     @inbounds dydxt .= 0
```
devmotion, line 207:
```suggestion
```

src/bijectors/simplex.jl
```diff
@@ -237,16 +222,10 @@ function simplex_link_jacobian(
...
         end
     end
-    @inbounds sum_tmp += x[K - 1]
-    @inbounds if !proj
-        @simd for i in 1:K
-            dydxt[i, K] = -1
-        end
-    end
-    return UpperTriangular(dydxt)'
+    return dydxt'
```
devmotion, line 225: The adjoint operation seems a bit annoying but I guess the algorithm should be updated in separate PRs if desired.

test/interface.jl
```diff
@@ -168,9 +168,11 @@ end
                 end
                 y = b(x)
                 @test b(param(x)) isa TrackedArray
-                @test log(abs(det(ForwardDiff.jacobian(b, x)))) ≈ logabsdetjac(b, x)
-                @test log(abs(det(ForwardDiff.jacobian(inverse(b), y)))) ≈
-                    logabsdetjac(inverse(b), y)
+                @test log(abs(det(ForwardDiff.jacobian(b, x)[1:end, 1:(end - 1)]))) ≈
```
devmotion, line 171:
```suggestion
                @test logabsdet(ForwardDiff.jacobian(b, x)[:, 1:(end - 1)])[1] ≈
```

test/interface.jl
```diff
@@ -168,9 +168,11 @@ end
...
+                    logabsdetjac(b, x)
+                @test log(
+                    abs(det(ForwardDiff.jacobian(inverse(b), y)[1:(end - 1), 1:end]))
+                ) ≈ logabsdetjac(inverse(b), y)
```
devmotion, line 175:
```suggestion
                @test logabsdet(ForwardDiff.jacobian(inverse(b), y)[1:(end - 1), :])[1] ≈ logabsdetjac(inverse(b), y)
```

src/bijectors/simplex.jl
sethaxen: Agreed, but this should be fixed in a separate PR.

src/bijectors/simplex.jl
sethaxen: Likewise, agreed, but this should be fixed separately.

torfjelde:
Haven't forgotten about this, but the DPPL integration was set back significantly by some other changes we had made. Should be done soon now :+1:

torfjelde:
Aaaalrighty! Does someone want to give this a look-over? I think pasts will pass now, an so it would be nice to get this merged.

sethaxen:
Thanks @torfjelde for the fixes! All LGTM, but someone else needs to review.

devmotion (commented):
Looks good to me, just a minor question.

It's a breaking change it seems, so IMO it would be good to include the correct version bump in the PR to avoid accidentally tagging a non-breaking release.

devmotion (approved):

PR #264, opened 2023-06-06
Use extensions when possible

torfjelde:
Depends on https://github.com/TuringLang/DistributionsAD.jl/pull/245


ext/BijectorsDistributionsADExt.jl
```diff
@@ -0,0 +1,114 @@
...
+        FillMatrixOfUnivariate,
+        MatrixOfUnivariate,
+        FillVectorOfMultivariate,
+        VectorOfMultivariate,
+        TuringScalMvNormal,
+        TuringDiagMvNormal,
+        TuringDenseMvNormal
+end
+
+using LinearAlgebra
```
devmotion, line 33: This is not OK - an extension may only import from the main package or its associated weak dependency/dependencies. Even standard libraries must not be loaded directly.

torfjelde: Ah, thanks! I just copy pasted from the compat files. Will change once I get back to computer

src/Bijectors.jl
```diff
@@ -276,29 +277,27 @@ maporbroadcast(f, x::AbstractArray{<:Any,N}...) where {N} = map(f, x...)
 maporbroadcast(f, x::AbstractArray...) = f.(x...)
 
 # optional dependencies
```
devmotion, line 279: Can we also add the `using Requires` above and instead add here (in a separate `if`-statement!) a
```julia
if !isdefined(Base, :get_extension)
    using Requires
end
```
(Not sure, possibly you have to change the check below wrapping `__init__()` to `@static if !isdefined....` - at least this seems to be the common pattern)

torfjelde: I think the `@static if` instead of `__init__` is only needed if we have anything else we want to put in `__init__`, no? In this case we don't, so we just put both `using Requires` and the definition of `__init__` inside the same branch?

torfjelde: Apparently not :confused: Using `@static` now

torfjelde:
Wanna give it a :+1: ?:)

devmotion:
Mainly good, but from a quick look at the CI logs and test/Project.toml it seems again we forgot to load/test the LazyArrays extension?

torfjelde:
Ah yeah, thanks! I did that, and it indeed seems like DistributionsAD wasn't properly fixed: https://github.com/TuringLang/DistributionsAD.jl/pull/247

torfjelde:
Yay! Wanna give it a thumbs up @devmotion @yebai ?:)

Issue #266, opened 2023-06-08
rational quadratic flows not supporting Float32 input


zuhengxu:
Here is a minimal working example:
```julia
using Bijectors

K = 10; B = 2
Bijectors.RationalQuadraticSpline(randn(Float32,K), randn(Float32, K), randn(Float32, K-1), B)
```
which gives the error:

```julia

ERROR: MethodError: no method matching Bijectors.RationalQuadraticSpline(::Vector{Float64}, ::Vector{Float64}, ::Vector{Float32})

Closest candidates are:
  Bijectors.RationalQuadraticSpline(::T, ::T, ::T) where T<:(AbstractVector)
   @ Bijectors ~/.julia/packages/Bijectors/Fp0rB/src/bijectors/rational_quadratic_spline.jl:80
  Bijectors.RationalQuadraticSpline(::A, ::A, ::A, ::T2) where {T1, T2, A<:AbstractVector{T1}}
   @ Bijectors ~/.julia/packages/Bijectors/Fp0rB/src/bijectors/rational_quadratic_spline.jl:103

Stacktrace:
 [1] Bijectors.RationalQuadraticSpline(widths::Vector{Float32}, heights::Vector{Float32}, derivatives::Vector{Float32}, B::Int64)
   @ Bijectors ~/.julia/packages/Bijectors/Fp0rB/src/bijectors/rational_quadratic_spline.jl:109
 [2] top-level scope
   @ ~/Research/NF-playground/NF-examples/neural_spline_flow/nsf_layer.jl:55
```

This is because https://github.com/TuringLang/Bijectors.jl/blob/dd8a24b02feccc8f1bef180e7419fb03a61ba82f/src/bijectors/rational_quadratic_spline.jl#LL99C1-L107C4 (line 103--104) is not type stable---` .- 0.5` returns Float64 instead Float32, while line 105 maintains Float32. This is probablematic because the three fields of `RationalQuadraticSpline{T}` has to be the same type. 

Same issue applies to https://github.com/TuringLang/Bijectors.jl/blob/dd8a24b02feccc8f1bef180e7419fb03a61ba82f/src/bijectors/rational_quadratic_spline.jl#LL109C1-L123C4

torfjelde:
Solved by #267

PR #267, opened 2023-06-08
making rqs compatible with float32 input

zuhengxu:
A simple fix to issue #266.

### DIFF
--- a/src/bijectors/rational_quadratic_spline.jl
+++ b/src/bijectors/rational_quadratic_spline.jl
@@ -100,8 +100,8 @@ function RationalQuadraticSpline(
     widths::A, heights::A, derivatives::A, B::T2
 ) where {T1,T2,A<:AbstractVector{T1}}
     return RationalQuadraticSpline(
-        (cumsum(vcat([zero(T1)], LogExpFunctions.softmax(widths))) .- 0.5) * 2 * B,
-        (cumsum(vcat([zero(T1)], LogExpFunctions.softmax(heights))) .- 0.5) * 2 * B,
+        cumsum(vcat([zero(T1)], LogExpFunctions.softmax(widths))) * 2 * B .- B,
+        cumsum(vcat([zero(T1)], LogExpFunctions.softmax(heights))) *2 * B .- B, 
         vcat([one(T1)], LogExpFunctions.log1pexp.(derivatives), [one(T1)]),
     )
 end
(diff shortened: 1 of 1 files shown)


torfjelde (approved):
Good stuff @zuhengxu :)

torfjelde:
Btw, can you bump the patch version?

src/bijectors/rational_quadratic_spline.jl
```diff
@@ -100,8 +100,8 @@ function RationalQuadraticSpline(
     widths::A, heights::A, derivatives::A, B::T2
 ) where {T1,T2,A<:AbstractVector{T1}}
     return RationalQuadraticSpline(
-        (cumsum(vcat([zero(T1)], LogExpFunctions.softmax(widths))) .- 0.5) * 2 * B,
-        (cumsum(vcat([zero(T1)], LogExpFunctions.softmax(heights))) .- 0.5) * 2 * B,
+        cumsum(vcat([zero(T1)], LogExpFunctions.softmax(widths))) * 2 * B .- B,
```
devmotion, line 103: We could save a few allocations:
```suggestion
        cumsum(vcat([zero(T1)], LogExpFunctions.softmax(widths))) .* (2 * B) .- B,
```

src/bijectors/rational_quadratic_spline.jl
```diff
@@ -100,8 +100,8 @@ function RationalQuadraticSpline(
...
+        cumsum(vcat([zero(T1)], LogExpFunctions.softmax(heights))) * 2 * B .- B,
```
devmotion, line 104:
```suggestion
        cumsum(vcat([zero(T1)], LogExpFunctions.softmax(heights))) .* (2 * B) .- B,
```

devmotion (commented):
Looks good (a few allocations could be removed), could we also add tests for the bug?

zuhengxu:
> Btw, can you bump the patch version?

just did! 



> Looks good (a few allocations could be removed), could we also add tests for the bug?

@devmotion sure! What kind of tests you want to include? Like the changed code returns the same value as the original one with Float64 inputs?

devmotion:
The failing example in https://github.com/TuringLang/Bijectors.jl/issues/266 would be good, I think. That is, ensuring that the construction does not error. And then probably, as you suggested (I think), a quick consistency check against the `Float64` version would be reasonable.

zuhengxu:
> The failing example in #266 would be good, I think. That is, ensuring that the construction does not error. And then probably, as you suggested (I think), a quick consistency check against the `Float64` version would be reasonable.

Tests included! No error.

Issue #265, opened 2023-06-06
Add API function to retrieve size of bijector output from bijector input


sethaxen:
As originally noted in https://github.com/TuringLang/Bijectors.jl/pull/263#issuecomment-1577424300, the following lines assume that inputs and outputs of a bijector have the same shape, which is no longer required:
- https://github.com/TuringLang/Bijectors.jl/blob/fc155c9f1fdee3e5bfcb8b2023f37a43606b73e1/src/bijectors/stacked.jl#L77
- https://github.com/TuringLang/Bijectors.jl/blob/fc155c9f1fdee3e5bfcb8b2023f37a43606b73e1/src/transformed_distribution.jl#L95

To fix this, these methods need an API function that computes the output size of a bijector from the input size, defaulting to the same size.

Alternatively, if Bijectors is meant to support non-array types, such as arbitrary `Struct`s, then these size checks make no sense, and it is perhaps better to generalize this code.

PR #270, opened 2023-06-17
Added `output_length` and `output_size`

torfjelde:
This is needed to ensure that transformed distributions are handled correctly now that transformations can change the support.

Related: #265


src/transformed_distribution.jl
```diff
@@ -101,8 +100,8 @@ end
 ##############################
 
 # size
-Base.length(td::Transformed) = length(td.dist)
-Base.size(td::Transformed) = size(td.dist)
+Base.length(td::Transformed) = output_length(td.transform, size(td.dist))
```
devmotion, line 103: It seems this is the only time `output_length` is used:
```suggestion
Base.length(td::Transformed) = only(output_size(td.transform, size(td.dist)))

File: src/interface.jl [hunk 2]
```diff
@@ -35,6 +35,26 @@ function logabsdetjac(f::Columnwise, x::AbstractMatrix)
...
+    output_length(f, sz::Tuple)
+
+Returns the output length of `f` given the input length `len` or size `sz`.
+"""
+output_length(f, len::Int) = only(output_size(f, (len,)))
+output_length(f, len::Tuple{Int,Vararg{Int}}) = only(output_size(f, len))
+    sz = output_size(f, len)
+    @assert length(sz) == 1
+    return first(sz)
+end
```
torfjelde, line 56:
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -232,6 +232,17 @@ function logabsdetjac(::Inverse{VecCorrBijector}, y::AbstractVector{<:Real})
     return _logabsdetjac_inv_corr(y)
 end
 
+function output_size(::VecCorrBijector, sz::Tuple{Int,Int})
+    @assert sz[1] == sz[2]
```
torfjelde, line 236:
```suggestion
    sz[1] == sz[2] || error("sizes should be equal; received $(sz)")
```

torfjelde:
Cheers @devmotion ! I did all your suggestions + removed `output_length` due to it's limited use. Also added more descriptive error :+1:

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.12.8"
+version = "0.13.0"
```
torfjelde, line 3: Just for the record: this isn't actually a breaking change, but I want to wait with a new release until we've merged this PR + #271 and #263 , i.e. we have have proper support for everything.

torfjelde:
I'm fixing `Stacked` atm and I think that should probably go in this PR otherwise it'll buggy.

Project.toml
torfjelde: With the changes to `Stacked`, this is now indeed a breaking PR.

src/bijectors/stacked.jl
```diff
@@ -23,78 +23,157 @@ b([0.0, 1.0]) == [b1(0.0), 1.0]  # => true
...
+    )
+end
+function Stacked(bs::Tuple, ranges_in::Tuple)
+    ranges_out = determine_output_ranges(bs, ranges_in)
+    return Stacked{typeof(bs),typeof(ranges_in)}(
+        bs, ranges_in, ranges_out, sum(length, ranges_in), sum(length, ranges_out)
+    )
+end
+Stacked(bs::AbstractArray, ranges::Tuple) = Stacked(bs, [ranges...])
+Stacked(bs::Tuple, ranges::AbstractArray) = Stacked([bs...], ranges)
```
devmotion, line 45:
```suggestion
Stacked(bs::AbstractArray, ranges::Tuple) = Stacked(bs, collect(ranges))
Stacked(bs::Tuple, ranges::AbstractArray) = Stacked(collect(bs), ranges)
```

src/transformed_distribution.jl
```diff
@@ -1,20 +1,19 @@
+function variateform(d::Distribution, b)
```
devmotion, line 1: Do we need this function? AFAICT it's only used in the constructor of `TransformedDistribution` below. Could we directly use `Distributions.variate_form` instead?

torfjelde: > Do we need this function? AFAICT it's only used in the constructor of TransformedDistribution below.

Yeah I introduced because of the constructor since we need to determine what the variate form is going to be using the underlying distribution and the transformation. 

> Could we directly use Distributions.variate_form instead?

I don't quite see how that method helps. AFAIK that is only meant to return the variateform of a given `Sampleable` type; here we're just trying to determine what the variate form should be when we combine a distribution with a transformation.

torfjelde:
Would be nice to get this merged somewhat quickly if there's nothing signficant holding it back, as everything related to "proper support for contrainsted distributions in Turing" is held back by this :grimacing: 

Note that the `Stacked` _really_ needs a rework, as have been noted many times before, but I don't think we should do that now.

src/transformed_distribution.jl
devmotion: Yeah, I guess you're right - I assumed one could use something like `variate_form(b(d))` but the point is just to compute the variate type of "`b(d)`" here.

src/interface.jl
```diff
@@ -35,6 +35,21 @@ function logabsdetjac(f::Columnwise, x::AbstractMatrix)
...
+
+Returns the output size of `f` given the input size `sz`.
+"""
+output_size(f, sz) = sz
+
+"""
+    output_length(f, len::Int)
+    output_length(f, sz::Tuple)
+
+Returns the output length of `f` given the input length `len` or size `sz`.
```
devmotion, line 49: Should be
```suggestion

Returns the output length of `f` given the input length `len`.
```

src/transformed_distribution.jl
```diff
@@ -101,8 +100,8 @@ end
 ##############################
 
 # size
-Base.length(td::Transformed) = length(td.dist)
-Base.size(td::Transformed) = size(td.dist)
+Base.length(td::Transformed) = only(output_size(td.transform, size(td.dist)))
```
devmotion, line 103: Based on the docstring here https://github.com/JuliaStats/Distributions.jl/blob/625104475541b6774c3a452facae7e7a989c0fdc/src/matrixvariates.jl#L18 this should be
```suggestion
Base.length(td::Transformed) = prod(output_size(td.transform, size(td.dist)))
```

devmotion (approved):
Two minor comments, otherwise it looks good to me.

torfjelde:
Awesome; thanks David! :heart:

PR #271, opened 2023-06-18
Vector-version for `PDBijector`

torfjelde:
Similar vein to #246 and #263.

NOTE: This is based on #270 .


torfjelde:
I made an issue to remind us that we should improve this implementation in the future. But for now I think we should just make it work, and then we can make it fast later.

torfjelde:
Just pushing this once more: the release of 0.13 is just waiting for this PR. We should really get this in the same batch because it will technically be breaking. Otherwise we'll have to release 0.14 immediately after 0.13 :confused: 

Could someone maybe just give it a quick look-over and hit approve? :grimacing: There really isn't anything controversial in here + it just reuses existing implementation.

torfjelde:
Well shit. I messed up. My previous comment no longer holds true :grimacing:

src/utils.jl
```diff
@@ -18,3 +18,94 @@ cholesky_factor(X::AbstractMatrix) = cholesky_factor(cholesky(Hermitian(X)))
...
+
+Return a mask for elements of `X` above the `k`th diagonal.
+"""
+function triu_mask(X::AbstractMatrix, k::Int)
+    # Ensure that we're working with a square matrix.
+    LinearAlgebra.checksquare(X)
+
+    # Using `similar` allows us to respect device of array, etc., e.g. `CuArray`.
+    m = similar(X, Bool)
+    return triu(.~m .| m, k)
```
devmotion, line 33: Why not just
```suggestion
    return triu!(fill!(m, true), k)
```
?

ext/BijectorsTrackerExt.jl
```diff
@@ -529,7 +529,27 @@ end
 
 # NOTE: Probably doesn't work in complete generality.
 wrap_chainrules_output(x) = x
+function wrap_chainrules_output(x::ChainRulesCore.Thunk)
```
devmotion, line 532: Is this covered by the tests?

devmotion (commented):
What do you mean by your last comment, what exactly does not hold anymore?

src/utils.jl
torfjelde: Original reasoning behind this line was to keep it Zygote-compat, but now that we've moved it into a separate function, I guess we can just mark `triu_mask` as no-grad?

src/bijectors/pd.jl
```diff
@@ -40,3 +40,48 @@ end
 function with_logabsdet_jacobian(b::PDBijector, X)
     return transform(b, X), logabsdetjac(b, X)
 end
+
+struct PDVecBijector <: Bijector end
+
+function _triu_dim_from_length(d)
```
harisorgn, line 46: Duplicate of the one in utils.jl . Delete this one?

ext/BijectorsTrackerExt.jl
torfjelde: No it isn't! _But_ do we want to? :confused: 
I just started writing a test, and the realized it means I have to move `wrap_chainrules__output` to Bijectors.jl itself rather than as an extension. _But_ it's only really used for Tracker (it is also used for one part in ReverseDiffjl, but this can be dropped in favor of the macro that has been added to ReverseDiff.jl).

devmotion: > No it isn't! 

But does that mean it's not needed?

test/bijectors/pd.jl
```diff
@@ -1,13 +1,37 @@
...
+        # used in the ChangesOfVariables.jl tests will fail as the jacobian will have determinant 0.
+        # Hence, we disable those tests.
+        test_bijector(b, x; test_not_identity=true, changes_of_variables_test=false)
+
+        # Check that output sizes are computed correctly.
+        tdist = transformed(dist, b)
+        @test length(tdist) == length(y)
+        @test tdist isa MultivariateDistribution
+
+        dist_unconstrained = transformed(MvNormal(zeros(length(tdist)), I), inverse(b))
```
harisorgn, line 33: this is a constrained distribution, right?

torfjelde:
> Well shit. I messed up. My previous comment no longer holds true grimacing

Haha, I wanted to defer the release of Bijectors 0.13 until this had been merged, but in my excitement I forgot and made a release. Hence my comments regarding this above ain't no truth anymore :grimacing: 

But I think just releasing this as 0.13.1 is fine so np :+1:

src/bijectors/pd.jl
harisorgn: 🤦  yep, all good.

torfjelde:
I believe comments have been addressed:)

harisorgn (approved):

PR #273, opened 2023-06-19
Add CDFBijector and QuantileBijector

sethaxen:
This PR adds two new bijectors: `CDFBijector` and `QuantileBijector`, which do exactly what they sound like.

To-do:
- [x] add code
- [x] add docstrings with doctests
- [x] add tests


yebai:
@sethaxen I added a commit with some docs and tests, together with a vector bijector implementation. Can you take a look and help us get it merged?

PR #276, opened 2023-07-13
Fix links

cpfiffer:
Some of the links were missing the `library` part of the URL. 

I noticed that our docs don't seem to render the stable URL though (https://turinglang.github.io/Bijectors.jl/stable) -- not sure why that is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Bijectors.jl
 
-[![Stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://turinglang.github.io/Bijectors.jl/stable)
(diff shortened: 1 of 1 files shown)


devmotion:
AFAICT the links are actually correct - changing them means linking to the (possibly outdated) copy of the docs on the Turing webpage instead of the actual Bijectors docs.

devmotion:
https://github.com/TuringLang/Bijectors.jl/pull/262 was supposed to fix TagBot which did not trigger the build of the stable docs - but TagBot is pretty unreliable and broken (no maintainers...), so maybe it still doesn't work properly.

cpfiffer:
Ah -- good point. I'll close for now, seems the issue is basically with the stable docs link.

PR #280, opened 2023-08-04
AD fix for PDBijector

torfjelde:
Addresses the issues described in https://github.com/TuringLang/Turing.jl/pull/2018#issuecomment-1653408574 for ReverseDiff.jl by adding a few new functions and corresponding rrules:
- `cholesky_lower`/`cholesky_upper`: does what `cholesky_factor` does, but always returning a raw `Matrix`, thus making it compatible with the likes of ReverseDiff.
- Defers ReverseDiff-differentiation of `permutedims` to ChainRules.

In addition, fixes a missing import in ReverseDiffExt (IMO we should never `import` but instead always qualify the methods we're overloading in extensions to avoid these sorts of issues, but I'll make a separate PR for this after this has gone through).

Note that this does not address the issue for Tracker.jl.


src/utils.jl
```diff
@@ -19,6 +19,65 @@ cholesky_factor(X::Cholesky) = X.U
 cholesky_factor(X::UpperTriangular) = X
 cholesky_factor(X::LowerTriangular) = X
 
+# TODO: Add `check` as an argument?
```
torfjelde, line 22: Should we?

src/utils.jl
```diff
@@ -19,6 +19,65 @@ cholesky_factor(X::Cholesky) = X.U
...
+    cholesky_lower(X)
+
+Return the lower triangular Cholesky factor of `X` as a `Matrix`
+rather than `LowerTriangular`.
+
+!!! note
+    This is a thin wrapper around `cholesky(Hermitian(X)).L`
+    but with a custom `ChainRulesCore.rrule` implementation.
+"""
+cholesky_lower(X::AbstractMatrix) = lower_triangular(parent(cholesky(Hermitian(X)).L))
```
torfjelde, line 33: I wrap in `Hermitian` to effectively do the same as the current implementation of `cholesky_factor` but I believe `cholesky(::Hermitian)` is only valid starting from Julia 1.8 (going by a comment in BijectorsReverseDiffExt), so we need to fix this.

src/utils.jl
```diff
@@ -19,6 +19,65 @@ cholesky_factor(X::Cholesky) = X.U
...
+function ChainRulesCore.rrule(::typeof(cholesky_lower), X::AbstractMatrix)
```
torfjelde, line 34: Do we need a
```suggestion
function ChainRulesCore.rrule(::typeof(cholesky_lower), X::AbstractMatrix)
    project_to = ProjectTo(X)
```
?

src/utils.jl
```diff
@@ -19,6 +19,65 @@ cholesky_factor(X::Cholesky) = X.U
...
+    # TODO: Do we need a special rule for `Hermitian`?
```
torfjelde, line 35: I don't think so, since this `Hermitian(Hermitian(...))` should just be a no-op :shrug:

src/utils.jl
```diff
@@ -19,6 +19,65 @@ cholesky_factor(X::Cholesky) = X.U
...
+    H, hermitian_pullback = ChainRulesCore.rrule(Hermitian, X, :L)
+    # TODO: what do we do with `check` here?
```
torfjelde, line 37:
```suggestion
```

src/utils.jl
torfjelde: Note that this won't work with ReverseDiff :confused:

devmotion:
Is there a particular reason for defining a ChainRule-rule and applying it to ReverseDiff even though the only (?) broken backends are ReverseDiff and Tracker? It would seem a bit more natural to define a rule for ReverseDiff directly (ignoring Tracker as discussed in Turing).

torfjelde:
> Is there a particular reason for defining a ChainRule-rule and applying it to ReverseDiff even though the only (?) broken backends are ReverseDiff and Tracker? It would seem a bit more natural to define a rule for ReverseDiff directly (ignoring Tracker as discussed in Turing).

Nah. I had the same thought, but a) I'm fairly familiar with defining rrules using ChainRules, and not at all with ReverseDiff, b) it's easy to make use of the `rrule` in defining a `Tracker.@grad`, and c) I figured this wouldn't hurt Zygote perf vs. _not_ defining this :shrug:

ext/BijectorsReverseDiffExt.jl
```diff
@@ -253,11 +255,19 @@ end
...
+    vals::TrackedVector{<:Real}, k::Int, dim::Int
+)
 
 @grad_from_chainrules _link_chol_lkj(x::TrackedMatrix)
 @grad_from_chainrules _inv_link_chol_lkj(x::TrackedVector)
 
+@grad_from_chainrules Bijectors.cholesky_lower(X::TrackedMatrix)
+@grad_from_chainrules Bijectors.cholesky_upper(X::TrackedMatrix)
+
+# TODO: Type-piracy; probably shouldn't do this.
```
devmotion, line 268: No, this should really not be defined in Bijectors. I'm sure this will lead to surprising debugging and issues when Bijectors is (not) loaded.

devmotion:
One can just define the pullback in a function and reuse it without defining an `rrule`? From the code it seems the `rrule` does not provide any performance improvements over the already existing `rrule`s, so there's no benefit for ChainRules-compatible AD backends. Defining a ReverseDiff-rule is (sometimes) very similar to defining a Tracker rule (see, e.g., https://github.com/TuringLang/DistributionsAD.jl/blob/5847e86f7783ea8f745a7465c2f4f9020c729051/ext/DistributionsADReverseDiffExt.jl#L38-L43).

torfjelde:
> One can just define the pullback in a function and reuse it without defining an rrule? From the code it seems the rrule does not provide any performance improvements over the already existing rrules, so there's no benefit for ChainRules-compatible AD backends.

But isn't it fair to assume that an `rrule` will generally lead to improvements in type-stability, and thus such a `rrule` also benefitting the likes of Zygote? 

But I'm happy to not use a `rrule` here, if that is preferred. I always just do it by default because we generally have good tools to ensure that this works as intended + given the amount of type-instabilities I've encountered with Zygote, I've "arrived" at the conclusion that writing a `rrule` to avoid tracing through the full callstack is generally considered to beneficial :shrug:

devmotion:
Regarding https://github.com/JuliaDiff/ReverseDiff.jl/issues/236, the macro has always had issues and limitations (one bug was just fixed recently), so in my experience it's not really the case (and possibly a too strong expectation) that it brings CR-compatibility to ReverseDiff.

test/ad/utils.jl
```diff
@@ -1,32 +1,36 @@
 # Figure out which AD backend to test
 const AD = get(ENV, "AD", "All")
 
-function test_ad(f, x, broken=(); rtol=1e-6, atol=1e-6)
-    finitediff = FiniteDifferences.grad(central_fdm(5, 1), f, x)[1]
+function test_ad(f, x, broken=(); rtol=1e-6, atol=1e-6, use_forwarddiff_as_truth=false)
+    truth = if use_forwarddiff_as_truth
```
devmotion, line 5: Why is this needed, why does finite differencing fail? Because the positive-definiteness is violated by perturbations? The common fix (also used in DistributionsAD) for such problems is to test AD of a different function, e.g., `g(A) = f(A * A' + I)` instead of `f` directly.

torfjelde: Ah, of course! Cheers!

torfjelde:
> Regarding https://github.com/JuliaDiff/ReverseDiff.jl/issues/236, the macro has always had issues and limitations (one bug was just fixed recently), so in my experience it's not really the case (and possibly a too strong expectation) that it brings CR-compatibility to ReverseDiff.

Yeaaah but it's so darn convenient :disappointed:

torfjelde:
Btw, @devmotion the fact that I'm facing issues when using `@grad_from_chainrules` is not the macro's fault, but it's because we're just calling `value` on the inputs, while `@grad_from_chainrules` does not (which is, arguably, the correct way of doing things).

src/utils.jl
torfjelde: I think this is the last remaining question @devmotion . I'm thinking "let's not, until we start using it"?

src/utils.jl
torfjelde: This is actually not a problem anymore since we're now defining the adjoint to circumvent the `cholesky` on tracked completely.

torfjelde:
So this is quite confusing. I'm trying to transition `VecCorrBijector` to also use the `cholesky_lower`, etc. and I'm now running into the following issue, _despite_ CI passing for the current implementation (which doesn't have a custom adjoint for Zygote):
```julia
julia> d = 4
4

julia> dist = LKJ(d, 2.0)
LKJ{Float64, Int64}(
d: 4
η: 2.0
)


julia> b = bijector(dist)
Bijectors.VecCorrBijector()

julia> x = rand(dist)
4×4 Matrix{Float64}:
  1.0        -0.19264   -0.63806    0.0930006
 -0.19264     1.0        0.259633  -0.168056
 -0.63806     0.259633   1.0        0.170947
  0.0930006  -0.168056   0.170947   1.0

julia> # (✓) Works!
       Zygote.gradient(x) do x
           sum(cholesky(Hermitian(x)).U)
       end

([1.141585190493663 1.7498257982625538 1.7225797690347182 1.6454362799052624; 0.0 0.5205946318942621 1.0163692667364 1.0672427905346635; 0.0 0.0 0.4745109346480554 0.8467387954612068; 0.0 0.0 0.0 0.5399333180126097],)

julia> # (×) Fails!
       Zygote.gradient(x) do x
           sum(parent(cholesky(Hermitian(x)).U))
       end
ERROR: MethodError: no method matching UpperTriangular(::NamedTuple{(:data,), Tuple{FillArrays.Fill{Float64, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}})

Closest candidates are:
  UpperTriangular(::UpperTriangular)
   @ LinearAlgebra ~/.julia/juliaup/julia-1.9.2+0.x64.linux.gnu/share/julia/stdlib/v1.9/LinearAlgebra/src/triangular.jl:21
  UpperTriangular(::ChainRulesCore.AbstractThunk)
   @ ChainRulesCore ~/.julia/packages/ChainRulesCore/0t04l/src/tangent_types/thunks.jl:68
  UpperTriangular(::TrackedMatrix)
   @ DistributionsADTrackerExt ~/.julia/packages/DistributionsAD/Ufc05/ext/DistributionsADTrackerExt.jl:131
  ...

Stacktrace:
 [1] (::Zygote.var"#1010#1013"{Cholesky{Float64, Matrix{Float64}}})(Δ::NamedTuple{(:data,), Tuple{FillArrays.Fill{Float64, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}})
   @ Zygote ~/.julia/packages/Zygote/JeHtr/src/lib/array.jl:630
 [2] (::Zygote.var"#3461#back#1014"{Zygote.var"#1010#1013"{Cholesky{Float64, Matrix{Float64}}}})(Δ::NamedTuple{(:data,), Tuple{FillArrays.Fill{Float64, 2, Tuple{Base.OneTo{Int64}, Base.OneTo{Int64}}}}})
   @ Zygote ~/.julia/packages/ZygoteRules/OgCVT/src/adjoint.jl:71
 [3] Pullback
   @ ./REPL[29]:3 [inlined]
 [4] (::Zygote.Pullback{Tuple{var"#51#52", Matrix{Float64}}, Tuple{Zygote.var"#3461#back#1014"{Zygote.var"#1010#1013"{Cholesky{Float64, Matrix{Float64}}}}, Zygote.Pullback{Tuple{typeof(cholesky), Hermitian{Float64, Matrix{Float64}}}, Tuple{Zygote.ZBack{ChainRules.var"#cholesky_HermOrSym_pullback#2122"{Hermitian{Float64, Matrix{Float64}}, Cholesky{Float64, Matrix{Float64}}}}, Zygote.Pullback{Tuple{Type{NoPivot}}, Tuple{}}}}, Zygote.var"#3027#back#778"{Zygote.var"#772#776"{Matrix{Float64}}}, Zygote.Pullback{Tuple{typeof(parent), UpperTriangular{Float64, Matrix{Float64}}}, Tuple{Zygote.var"#2184#back#299"{Zygote.var"#back#298"{:data, Zygote.Context{false}, UpperTriangular{Float64, Matrix{Float64}}, Matrix{Float64}}}}}, Zygote.var"#3299#back#918"{Zygote.var"#back#917"{Hermitian{Float64, Matrix{Float64}}}}}})(Δ::Float64)
   @ Zygote ~/.julia/packages/Zygote/JeHtr/src/compiler/interface2.jl:0
 [5] (::Zygote.var"#75#76"{Zygote.Pullback{Tuple{var"#51#52", Matrix{Float64}}, Tuple{Zygote.var"#3461#back#1014"{Zygote.var"#1010#1013"{Cholesky{Float64, Matrix{Float64}}}}, Zygote.Pullback{Tuple{typeof(cholesky), Hermitian{Float64, Matrix{Float64}}}, Tuple{Zygote.ZBack{ChainRules.var"#cholesky_HermOrSym_pullback#2122"{Hermitian{Float64, Matrix{Float64}}, Cholesky{Float64, Matrix{Float64}}}}, Zygote.Pullback{Tuple{Type{NoPivot}}, Tuple{}}}}, Zygote.var"#3027#back#778"{Zygote.var"#772#776"{Matrix{Float64}}}, Zygote.Pullback{Tuple{typeof(parent), UpperTriangular{Float64, Matrix{Float64}}}, Tuple{Zygote.var"#2184#back#299"{Zygote.var"#back#298"{:data, Zygote.Context{false}, UpperTriangular{Float64, Matrix{Float64}}, Matrix{Float64}}}}}, Zygote.var"#3299#back#918"{Zygote.var"#back#917"{Hermitian{Float64, Matrix{Float64}}}}}}})(Δ::Float64)
   @ Zygote ~/.julia/packages/Zygote/JeHtr/src/compiler/interface.jl:45
 [6] gradient(f::Function, args::Matrix{Float64})
   @ Zygote ~/.julia/packages/Zygote/JeHtr/src/compiler/interface.jl:97
 [7] top-level scope
   @ REPL[29]:2
```
which then also breaks `cholesky_lower` :confused:

torfjelde:
Seems like something similar to https://github.com/FluxML/Zygote.jl/blob/29fa32a688fb4454d2ee81b9ba2a6484e4468bda/src/lib/array.jl#L356-L357 is needed.

Adding those, I indeed manage to get stuff running, but the resulting adjoint type is `UpperTriangular` no matter what :confused: 

``` julia
julia> Zygote.@adjoint LinearAlgebra.parent(x::UpperTriangular) = parent(x), Δ -> (UpperTriangular(Δ),)

julia> Zygote.@adjoint LinearAlgebra.parent(x::LowerTriangular) = parent(x), Δ -> (LowerTriangular(Δ),)

julia> x
2×2 Matrix{Float64}:
  1.0       -0.612002
 -0.612002   1.0

julia> last(Zygote.gradient(x) do x
           sum(parent(cholesky(Hermitian(x)).U))
       end)
2×2 UpperTriangular{Float64, Matrix{Float64}}:
 1.0428  1.77385
  ⋅      0.632226

julia> last(Zygote.gradient(x) do x
           sum(parent(cholesky(Hermitian(x)).L))
       end)
2×2 UpperTriangular{Float64, Matrix{Float64}}:
 1.0428  1.77385
  ⋅      0.632226
```

PR #281, opened 2023-08-07
AD fix for CorrBijector

torfjelde:
This is a sibling-PR of #280, making use of the functionality introduced there to ensure that `CorrBijector` and it's siblings are also working as intended.


ext/BijectorsZygoteExt.jl
```diff
@@ -244,4 +240,9 @@ end
         return replace_diag(log, Y)
     end
 end
+
+# TODO: Remove these as soon as https://github.com/FluxML/Zygote.jl/pull/1444 is merged.
+@adjoint LinearAlgebra.parent(x::LinearAlgebra.UpperTriangular) = parent(x), ȳ -> (LinearAlgebra.UpperTriangular(ȳ),)
+@adjoint LinearAlgebra.parent(x::LinearAlgebra.LowerTriangular) = parent(x), ȳ -> (LinearAlgebra.LowerTriangular(ȳ),)
```
torfjelde, line 246: I've added this here to check if tests are succeeding; will remove as soon as  the mentioned PR goes through.

src/bijectors/corr.jl
```diff
@@ -321,7 +325,7 @@ function _link_chol_lkj(W::UpperTriangular)
     return z
 end
 
-_link_chol_lkj(W::LowerTriangular) = _link_chol_lkj(transpose(W))
+_link_chol_lkj_from_lower(W::AbstractMatrix) = _link_chol_lkj_from_upper(transpose_eager(W))
```
torfjelde, line 328: This one is a bit weird because I just saw that it has it's own chainrule defined while the evaluation itself does _not_ have a specialized implementation..

torfjelde:
> In latest version cholesky takes AbstractMatrix.

But we only call `cholesky` with `Hermitian`, which _is_ defined, no?

> Maybe related, LKJCholesky default [uplo='L'](https://juliastats.org/Distributions.jl/latest/cholesky/#Distributions.LKJCholesky); Hermitian default [uplo='U'](https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.Hermitian)

The issue is two-fold. First, we're giving the transformation in `ForwardDiff.jacobian(...)` a upper-triangular but it expects a lower-triangular (it ends up calling `cholesky_lower`).
But even if we fix this, i.e. pass `x.U` instead, lit just returns 0s:

``` julia
julia> using Bijectors

julia> dist = LKJCholesky(3, 1)
LKJCholesky{Float64}(
d: 3
η: 1.0
uplo: L
)


julia> x = rand(dist)
LinearAlgebra.Cholesky{Float64, Matrix{Float64}}
L factor:
3×3 LinearAlgebra.LowerTriangular{Float64, Matrix{Float64}}:
  1.0         ⋅         ⋅ 
 -0.107859   0.994166   ⋅ 
 -0.38706   -0.246285  0.888554

julia> link(dist, x)
3-element Vector{Float64}:
 -0.10828016710654258
 -0.40833723913774234
 -0.2737439074161078

julia> link(dist, x.L)
3-element Vector{Float64}:
 0.0
 0.0
 0.0
```

This was caused by this line: https://github.com/TuringLang/Bijectors.jl/blob/5bffcfb5a3c63b7515b3275acfdf742149dcbb04/src/utils.jl#L31

Here we ended up calling `Hermitian(X, :U)` on a `X` which is lower-triangular; I've now made this call `Hermitian(X, :L)`. Now we have `transpose(cholesky_lower(x.L)) == cholesky_upper(x.U)`, as expected.

torfjelde:
Alrighty; tests are at least passing locally for me now. Let's see if Julia 1.6 also works.

torfjelde:
Looks like we're ready to go:)

PR #286, opened 2023-08-16
Remove unnecessary (and incorrect) Zygote rule for PDBijector

yebai:
Addresses https://github.com/TuringLang/Turing.jl/pull/2018#issuecomment-1680944327

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -30,6 +30,9 @@ ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
+[extras]
+Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
+
 [extensions]
 BijectorsDistributionsADExt = "DistributionsAD"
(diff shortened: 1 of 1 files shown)


devmotion:
Long-term ideally we would not define Zygote-specific rules anymore but mainly/only ChainRules definitions.

To this PR: Doesn't this require https://github.com/FluxML/Zygote.jl/pull/1444? If so, I guess we should add/update the lower bound for Zygote.

Project.toml
```diff
@@ -30,6 +30,9 @@ ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
 Tracker = "9f7883ad-71c0-57eb-9f7f-b5c9e6d3789c"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
+[extras]
```
torfjelde, line 33: Why is only Zygote added here?

devmotion: It doesn't matter for our tests because we have test/Project.toml. But Pkg requires it, otherwise it's not possible to add a compat entry for a weak dependency.

devmotion: Maybe we should just remove the test/Project.toml if possible (IIRC it helped with some dependency incompatibilities a while ago but maybe that's not a problem anymore).

Issue #287, opened 2023-08-25
Can't apply Bijectors.ordered to TDist() and MvTDist()


boogiewoogit:
These should both work, but don't:
```
julia> using Turing
julia> Bijectors.ordered(filldist(TDist(1),2))
ERROR: ArgumentError: ordered transform is currently only supported for unconstrained distributions.
```

```
julia> using Turing
julia> Bijectors.ordered(MvTDist(1,zeros(2),Matrix(I(2))))
ERROR: MethodError: no method matching bijector(::MvTDist)
```

In particular TDist is supported on the real line and therefore unrestricted in the sense of this package:

[I have reported](https://github.com/TuringLang/Turing.jl/issues/2064) this to Turing.jl first, where it was identified that TDist's return type has the wrong support:

```
julia> bijector(TDist(1))
Bijectors.TruncatedBijector{Float64, Float64}(-Inf, Inf)
```

Red-Portal:
Should have been addressed by #288 !

PR #288, opened 2023-09-02
Fix for #287

torfjelde:
This addresses #287, in addition to making transformations of product distributions faster for several univariate distributions.


test/bijectors/ordered.jl
```diff
@@ -17,20 +17,23 @@ using LinearAlgebra
...
-    d_ordered = ordered(d)
-    @test d_ordered isa Bijectors.TransformedDistribution
-    @test d_ordered.dist === d
-    @test d_ordered.transform isa OrderedBijector
-    y = randn(5)
-    x = inverse(bijector(d_ordered))(y)
-    @test issorted(x)
+    @testset "$d" for d in [
+        MvNormal(1:5, Diagonal(6:10)),
+        MvTDist(1, 1:5, Matrix(I(5))),
```
Red-Portal, line 22: I think you could use the following alias to fix the failing test:
```suggestion
        Distributions.IsoTDist(1.0, 5, 1.0)
```

torfjelde: Yeah that might do the trick, but I'd say that it's a bug that you can't do the above.

Basically, the constructors of MvTDist are specialized to work on Vector ratheer than AbstractVector, which seems overly restrictive.

So I'll raise an issue, and here I'll concretize the range to a vector I think.

PR #297, opened 2023-12-03
Add support for distributions with monotonically increasing `bijector`

torfjelde:
Related: https://github.com/TuringLang/Bijectors.jl/issues/220

### DIFF
--- a/src/bijectors/ordered.jl
+++ b/src/bijectors/ordered.jl
@@ -17,14 +17,21 @@ Return a `Distribution` whose support are ordered vectors, i.e., vectors with in
 This transformation is currently only supported for otherwise unconstrained distributions.
 """
 function ordered(d::ContinuousMultivariateDistribution)
-    if bijector(d) !== identity
+    # We're good if the map from unconstrained (in which we apply the ordered bijector)
+    # to constrained is monotonically increasing, i.e. order-preserving. In that case,
+    # we can form the ordered transformation as `binv ∘ OrderedBijector() ∘ b`.
+    # TODO: Add support for monotonically _decreasing_ transformations. This will be the
+    # the same as above, but the ordering will be reversed by `binv` so we need to handle this.
+    b = bijector(d)
+    binv = inverse(b)
+    if !is_monotonically_increasing(binv)
         throw(
             ArgumentError(
-                "ordered transform is currently only supported for unconstrained distributions.",
+                "ordered transform is not supported for $d.",
             ),
         )
     end
-    return transformed(d, OrderedBijector())
+    return transformed(d, binv ∘ OrderedBijector() ∘ b)
 end
 
 with_logabsdet_jacobian(b::OrderedBijector, x) = transform(b, x), logabsdetjac(b, x)
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -230,6 +230,23 @@ transform!(::typeof(identity), x, y) = copy!(y, x)
 logabsdetjac(::typeof(identity), x) = zero(eltype(x))
 logabsdetjac!(::typeof(identity), x, logjac) = logjac
 
+###################
+# Other utilities #
+###################
+"""
+    is_monotonically_increasing(f)
+
+Returns `true` if `f` is monotonically increasing.
+"""
+is_monotonically_increasing(f) = false
+is_monotonically_increasing(::typeof(identity)) = true
+function is_monotonically_increasing(cf::ComposedFunction)
+    return is_monotonically_increasing(cf.inner) && is_monotonically_increasing(cf.outer)
+end
+is_monotonically_increasing(::typeof(exp)) = true
+is_monotonically_increasing(ef::Elementwise) = is_monotonically_increasing(ef.x)
+
+
 ######################
 # Bijectors includes #
 ######################
--- a/test/bijectors/ordered.jl
+++ b/test/bijectors/ordered.jl
@@ -22,11 +22,13 @@ end
         MvTDist(1, collect(1.0:5), Matrix(I(5))),
         product_distribution(fill(Normal(), 5)),
         product_distribution(fill(TDist(1), 5)),
+        product_distribution(fill(LogNormal(), 5)),
+        product_distribution(fill(InverseGamma(2, 3), 5)),
     ]
         d_ordered = ordered(d)
         @test d_ordered isa Bijectors.TransformedDistribution
         @test d_ordered.dist === d
-        @test d_ordered.transform isa OrderedBijector
+        # @test d_ordered.transform isa OrderedBijector
         y = randn(5)
         x = inverse(bijector(d_ordered))(y)
         @test issorted(x)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.13.7"
+version = "0.13.8"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"


torfjelde:
@sethaxen maybe you want to have a look at this

src/interface.jl
```diff
@@ -230,6 +230,22 @@ transform!(::typeof(identity), x, y) = copy!(y, x)
...
+    is_monotonically_increasing(f)
+
+Returns `true` if `f` is monotonically increasing.
+"""
+is_monotonically_increasing(f) = false
+is_monotonically_increasing(::typeof(identity)) = true
+function is_monotonically_increasing(cf::ComposedFunction)
+    return is_monotonically_increasing(cf.inner) && is_monotonically_increasing(cf.outer)
+end
+is_monotonically_increasing(::typeof(exp)) = true
```
sethaxen, line 246:
```suggestion
is_monotonically_increasing(::typeof(exp)) = true
is_monotonically_increasing(::typeof(log)) = true
is_monotonically_increasing(binv::Inverse) = is_monotonically_increasing(inverse(b))
```

sethaxen: Since this is an interface function, would it be better to place these methods in the files where the corresponding bijectors are implemented? Also, I think we can mark this as true for `Logit`, `LeakyReLu`, `Scale` (when scale is positive), `Shift`, and `TruncatedBijector`.

sethaxen (changes requested):
The approach looks right! I suggest we add overloads for more bijectors. Is there a reason this PR doesn't also add `is_monotonically_decreasing`?

All univariate bijections are strictly monotonic. So we could define `is_monotonically_decreasing(b) = !ismonotonically_increasing(b)` if we documented that this function is only expected to give the correct answer when a univariate bijector is passed. But this could cause problems. Do we have any way to statically detect if a bijector is univariate?

The PR needs tests to cover each of the additions.

torfjelde:
> Is there a reason this PR doesn't also add is_monotonically_decreasing?
 
I did consider this, but AFAIK the only monotonically decreasing bijectors we have right now is `Scale` with negative coefficients which will require runtime checks and thus made me hesitant (was going to raise an issue about this).

But it's probably worth it, so I'll add that too :+1:

> Do we have any way to statically detect if a bijector is univariate?

Not at the moment, no.

> All univariate bijections are strictly monotonic. So we could define 

And because we don't, I'd prefer to make it all explicit so we end up with a method error instead of silently doing something strange.

src/interface.jl
torfjelde: Agreed:)

torfjelde:
Ah okay so now I remember another reason why I was holding back on `is_monotonically_decreasing`: AFAIK `Scale` is the only monotonically decreasing function we can have, but how do we implement `is_monotonically_decreasing` for `ComposedFunction`?

The condition

``` julia
is_monotonically_decreasing(f.inner) && is_monotonically_decreasing(f.outer)
```

won't be correct, e.g. `Scale(-1)` and `Scale(-1)` are both monotonically decreasing, but their composition is not.

sethaxen:
I don't understand the table, but I believe it amounts to first checking that all bijectors are (elementwise) univariate with `all(x -> is_monotonically_increasing(x) | is_monotonically_decreasing(x), bijectors)` and then checking that there are an odd number of decreasing bijectors with `mapreduce(is_monotonically_decreasing, xor, bijectors)`.

torfjelde:
My table is conveying the same idea, just on a per-composition-basis (since we're defining the method for `ComposedFunction`) :)

But I've now added support for monotonically decreasing functions too + tests:)

torfjelde:
I was trying to replicate a Markov Switching GARCH model and ran into wanting `ordered` for a positively-constrained distriibutions. Hence, I sort of need this PR :grimacing: Any chance we could get it through?

torfjelde:
@yebai @devmotion @sethaxen I believe this should be good to go

torfjelde:
I will merge this at the end of the day unless anyone else has any objections @devmotion @sethaxen :)

src/bijectors/ordered.jl
```diff
@@ -1,3 +1,11 @@
+struct SignFlip <: Bijector end
+
+with_logabsdet_jacobian(::SignFlip, x) = -x, zero(eltype(x))
+inverse(::SignFlip) = SignFlip()
+output_size(::SignFlip, dim) = dim
+is_monotonically_increasing(::SignFlip) = false
+is_monotonically_decreasing(::SignFlip) = false
```
sethaxen, line 7: Shouldn't this be `true`?

src/bijectors/ordered.jl
```diff
@@ -17,14 +25,22 @@ Return a `Distribution` whose support are ordered vectors, i.e., vectors with in
 This transformation is currently only supported for otherwise unconstrained distributions.
 """
 function ordered(d::ContinuousMultivariateDistribution)
-    if bijector(d) !== identity
-        throw(
-            ArgumentError(
-                "ordered transform is currently only supported for unconstrained distributions.",
-            ),
-        )
+    # We're good if the map from unconstrained (in which we apply the ordered bijector)
```
sethaxen, line 28: Would it be useful to add to the docstring a comment about when this is safe to use? e.g. in general, if `d` has variable parameters, then this is unsafe, because it implicitly renormalizes the distribution, but the log of the new normalization constant is never computed so is never included in the log density or gradient, though it would need to be when the parameters are variable.

torfjelde: When you say "variable parameters", do you mean that the size / length can vary?

src/bijectors/shift.jl
```diff
@@ -22,3 +22,5 @@ _logabsdetjac_shift(a, x) = zero(eltype(x))
 _logabsdetjac_shift_array_batch(a, x) = zeros(eltype(x), size(x, ndims(x)))
 
 with_logabsdet_jacobian(b::Shift, x) = transform(b, x), logabsdetjac(b, x)
+
+is_monotonically_increasing(::Shift) = true
```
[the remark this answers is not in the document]
torfjelde: Yep yep!

src/bijectors/ordered.jl
torfjelde, line 7:
```suggestion
is_monotonically_decreasing(::SignFlip) = true
```

test/bijectors/ordered.jl
```diff
@@ -22,11 +22,21 @@ end
...
+        product_distribution(fill(LogNormal(), 5)),
+        product_distribution(fill(InverseGamma(2, 3), 5)),
+        # negative supports
+        product_distribution(fill(-1 * InverseGamma(2, 3), 5)),
+        # bounded supports
+        product_distribution(fill(Uniform(1, 2), 5)),
+        product_distribution(fill(Beta(), 5)),
+        # different transformations
+        product_distribution(fill(transformed(InverseGamma(2, 3), Bijectors.Scale(3)), 5)),
+        product_distribution(fill(transformed(InverseGamma(2, 3), Bijectors.Scale(-3)), 5)),
```
torfjelde, line 35:
```suggestion
        # different transformations
        product_distribution(fill(transformed(InverseGamma(2, 3), Bijectors.Scale(3)), 5)),
        product_distribution(fill(transformed(InverseGamma(2, 3), Bijectors.Scale(-3)), 5)),
        product_distribution(fill(transformed(InverseGamma(2, 3), Bijectors.Shift(3)), 5)),
        product_distribution(fill(transformed(InverseGamma(2, 3), Bijectors.Shift(-3)), 5)),
```

src/bijectors/ordered.jl
sethaxen: No, I mean parameters that are being sampled, optimized in the model.

torfjelde: So you're saying
```julia
sigma ~ InverseGamma(2, 3)
x ~ ordered(MvNormal(zeros(2), sigma^2 * I))
```
will be incorrect?

torfjelde: > because it implicitly renormalizes the distribution, but the log of the new normalization constant is never computed so is never included in the log density or gradient,

Okay so I understand what you mean by this now, but I'm struggling a bit to understand why this is an issue here but not for other "typical" diffeomorphisms.

sethaxen: This technically would be an issue for other diffeomorphisms. E.g. if we used an Exp bijector for a normal random variable. The difference is that we generally define a bijector for a distribution already, so we reduce the chance this happens. E.g. in this case the user would construct truncated(Normal(m,s); lower=0), and then the normalization factor is computed. With ordered, the user is choosing to apply an additional ordered bijector, which then restricts the support.

FWIW, the docstring for ordered doesn't clarify whether it constructs the restriction of the wrapped distribution to the subset of the support consisting of ordered vectors (i.e. sample with rejection sampling) or whether it constructs the distribution whose random sampler consists of sampling from the wrapped distribution and then sorting. It should be the first.

sethaxen:
@torfjelde I haven't had a chance to check this directly myself, but does this do the right thing for products of heterogenous univariate distributions? It would be nice if a few cases were numerically checked via MC sampling. E.g. you could get exact MC draws with rejection sampling and then compare to MCMC draws and check expectations are similar using the MCSE.

torfjelde:
Yeah was thinking the same; will do that :+1:

src/bijectors/ordered.jl
torfjelde: > This technically would be an issue for other diffeomorphisms. E.g. if we used an Exp bijector for a normal random variable.

I still don't understand this part. The jacobian factor should make it so that everything is still normalized, no? So we're taking a normalized distribution and re-weighting it in a way to preserve normalization under the transformation.

> FWIW, the docstring for ordered doesn't clarify whether it constructs the restriction of the wrapped distribution to the subset of the support consisting of ordered vectors (i.e. sample with rejection sampling) or whether it constructs the distribution whose random sampler consists of sampling from the wrapped distribution and then sorting. It should be the first.

This I very much agree with :+1:

sethaxen: Well another way to put it is that the function is not in fact a diffeomorphisms, because it's injective, not bijective. The original distribution is defined on some support, but `ordered(...)` has a support that is a subset of the original distribution. So we're not just applying a bijective map and worrying about the change of measure. We're additionally restricting the support, which the Jacobian cannot capture. If we had a magic function that could compute normalization factors, then we would use that to compute the normalized logpdf after restricting the support, and then Ordered would indeed be a bijection.

sethaxen: Same deal. Because you're taking a density whose support is the whole real line and using a change of variables from the real line to the positive reals (not the whole real line). So you're changing the support if you use `Exp` for this distribution.

torfjelde: Nah, then I still don't get it. I'm with you that you're changing the domain, but $\exp: \mathbb{R} \to (0, \infty)$ is bijective, which are the domains we are interested in. Why does it matter that $\exp$ treated as a map from $\mathbb{R} \to \mathbb{R}$ is not bijective? o.O

Could you maybe write down exactly what you mean mathematically, or alternatively point me to a resource where I can read about this? I don't think I fully understand exactly what issue you're pointing to here, but really feel like this is something I should know so very keen on learning more.

torfjelde: Btw just implemented sampling from the following model:
``` julia
@model function demo_ordered(μ)
    k = length(μ)
    σ² ~ filldist(truncated(Normal(), lower=0), k)
    x ~ ordered(MvNormal(μ, Diagonal(σ²)))
    return (; σ², x)
end

k = 2
num_samples = 10_000

# Sample using NUTS.
μ = zeros(k)
σ² = ones(k)

model = fix(demo_ordered(μ); σ²)
chain = sample(model, NUTS(), num_samples)
xs_chain = permutedims(Array(chain))

# Rejection sampling.
d = MvNormal(μ, Diagonal(σ²))
xs_exact = mapreduce(hcat, 1:num_samples) do _
    xs = [rand(marginal(d, 1))]
    for i = 2:k
        x_prev = xs[end]
        while true
            x = rand(marginal(d, i))
            if x >= x_prev
                push!(xs, x)
                break
            end
        end
    end
    return xs
end

display(chain)

qts = [0.05, 0.25, 0.5, 0.75, 0.95]
qs_chain = mapslices(xs_chain; dims=2) do xs
    quantile(xs, qts)
end
qs_exact = mapslices(xs_exact; dims=2) do xs
    quantile(xs, qts)
end
```

and even that fails to produce consistent results :confused:

Are these two really supposed to be the same? 

Here's the output of the above:
```julia
┌ Info: Found initial step size
└   ϵ = 1.7000000000000002
Chains MCMC chain (10000×15×1 Array{Float64, 3}):

Iterations        = 1001:1:11000
Number of chains  = 1
Samples per chain = 10000
Wall duration     = 1.59 seconds
Compute duration  = 1.59 seconds
parameters        = x[1], x[2], x[3]
internals         = lp, n_steps, is_accept, acceptance_rate, log_density, hamiltonian_energy, hamiltonian_energy_error, max_hamiltonian_energy_error, tree_depth, numerical_error, step_size, nom_step_size

Summary Statistics
  parameters      mean       std      mcse     ess_bulk    ess_tail      rhat   ess_per_sec 
      Symbol   Float64   Float64   Float64      Float64     Float64   Float64       Float64 

        x[1]    0.0156    1.0073    0.0089   12830.2088   7634.0339    1.0003     8069.3137
        x[2]    1.6587    2.6114    0.0269   11976.8423   7526.6856    1.0000     7532.6052
        x[3]    3.2724    3.3218    0.0345   10898.1739   7911.5172    1.0001     6854.1974

Quantiles
  parameters      2.5%     25.0%     50.0%     75.0%     97.5% 
      Symbol   Float64   Float64   Float64   Float64   Float64 

        x[1]   -1.9646   -0.6662    0.0111    0.6789    1.9682
        x[2]   -1.2130    0.3388    1.2352    2.3719    7.3322
        x[3]   -0.3950    1.3836    2.6070    4.1849   11.4003

┌ Info: quantiles
│   qs_chain =
│    3×5 Matrix{Float64}:
│     -1.6457     -0.666194  0.0111229  0.678935  1.70693
│     -0.879387    0.338842  1.23522    2.37194   5.35761
│     -0.0182679   1.38357   2.60698    4.18492   8.67829
│   qs_exact =
│    3×5 Matrix{Float64}:
│     -1.6316    -0.683853  0.00858142  0.687207  1.64531
│     -0.506248   0.286727  0.888104    1.48044   2.36784
└      0.154898   0.905938  1.46605     2.03997   2.91713
```

As you can see, the quantiles are different.

src/bijectors/ordered.jl
```diff
@@ -17,14 +25,22 @@ Return a `Distribution` whose support are ordered vectors, i.e., vectors with in
...
+    if is_monotonically_decreasing(binv)
+        ordered_b = binv ∘ SignFlip() ∘ OrderedBijector() ∘ SignFlip() ∘ b
+    elseif is_monotonically_increasing(binv)
+        ordered_b = binv ∘ OrderedBijector() ∘ b
+    else
+        throw(ArgumentError("ordered transform is currently not supported for $d."))
     end
-    return transformed(d, OrderedBijector())
+
+    return transformed(d, ordered_b)
```
sethaxen, line 43: Hang on, is `transformed` even doing the right thing here? e.g. if `d` is an IID normal, then, `ordered_b` is an `OrderedBijector` that maps from unconstrained `y` to constrained `x`. So the result is an unconstrained distribution one samples from by first sampling from the IID normal and then constraining so the result is sorted. But this is not what we want. This is not the same thing as constraining `x` to be distributed according to `d` but restricted to the ordered subset of the support. Instead, what we want here is something that wraps `d` but assigns it to have the bijector `ordered_b`. Random generation is not possible, but log-density to within a normalization factor can be computed.

e.g. something like

```julia
using Distributions: Distribution, VariateForm, ValueSupport, ArrayLikeVariate

struct WithBijector{B,F<:VariateForm,S<:ValueSupport,D<:Distribution{F,S}} <: Distribution{F,S}
    dist::D
    bijector::B
end
with_bijector(d::Distribution, b) = WithBijector(d, b)
Bijectors.bijector(d::WithBijector) = d.bijector
Distributions.logpdf(d::WithBijector, x) = Distributions.logpdf(d.dist, x)
function Distributions.logpdf(d::WithBijector{<:Any, ArrayLikeVariate{N}}, x::AbstractArray{<:Real, M}) where {N, M}
    return Distributions.logpdf(d.dist, x)
end
Base.length(d::WithBijector) = length(d.dist)
```

If you then use `return with_bijector(d, inverse(ordered_b))` in this line, your example in https://github.com/TuringLang/Bijectors.jl/pull/297/files#r1597099187 seems to work.

torfjelde: > This is not the same thing as constraining x to be distributed according to d but restricted to the ordered subset of the support.

Ooookay so _now_ I'm with you! I originally thought that this was just _a_ transformation that moved us to an ordered distribution, but didn't necessary have any relation to the "base" distribution (until you're previous comment on the docstring clarifying what it was supposed to do), hence never considered this to be an issue.

But yeah this is a pretty significant bug then, and has been around for a while :grimacing:

sethaxen: Yeah, let's definitely get some tests in the CI that would have caught this bug.

torfjelde: Indeed. Working on a fix + tests as we speak.

sethaxen: Btw, something like `with_bijector` would be a convenient addition to the API for power users to manually select the bijector for a distribution. Applications include cases where the target distribution has high probability mass near a singularity of the default bijector z so one wants a custom bijector that moves the probability mass in the unconstrained space away from the singularity. Another application is programmatic benchmarking of alternative bijectors.

torfjelde: > something like with_bijector would be a convenient addition to the API 

Agreed! But I think this is such a special case it warrants its own distribution (because we _can_ implement `rand` using rejection sampling for several cases used in practice). But we should also add a more general `with_bijector` method :+1: 

But I tried implementing this, and I'm still seeing fairly drastic discrepancies in the quantiles :confused: Feeling a bit out of it today though so might just have messed up the impl or tests somewhere; let me know if you see anything obvious! I pushed the changes.

test/bijectors/ordered.jl
```diff
@@ -17,23 +37,80 @@ using LinearAlgebra
...
+            @test all(issorted, eachcol(xs_true))
+
+            qs_true = mapslices(xs_true; dims=2) do xs
+                quantile(xs, [0.25, 0.5, 0.75])
+            end
+            qs = mapslices(xs; dims=2) do xs
+                quantile(xs, [0.25, 0.5, 0.75])
+            end
+
+            @info "quantiles" qs_true qs
```
torfjelde, line 113: @sethaxen this is where I'm seeing a clear discrepancy still :confused:

src/bijectors/ordered.jl
sethaxen: > But I tried implementing this, and I'm still seeing fairly drastic discrepancies in the quantiles 😕 Feeling a bit out of it today though so might just have messed up the impl or tests somewhere; let me know if you see anything obvious! I pushed the changes.

Ah, really? The example you posted earlier looks pretty good to me:

```julia
using Turing
using Bijectors: ordered
using LinearAlgebra
using Random: Random
using Turing.DynamicPPL: fix
using PosteriorStats
using Random

@model function demo_ordered(μ)
    k = length(μ)
    σ² ~ filldist(truncated(Normal(), lower=0), k)
    x ~ ordered(MvNormal(μ, Diagonal(σ²)))
    return (; σ², x)
end

k = 2
num_samples = 1_000_000
num_chains = 8

# Sample using NUTS.
μ = zeros(k)
σ² = ones(k)

model = fix(demo_ordered(μ); σ²)

Random.seed!(0)
chain = sample(model, NUTS(), MCMCThreads(), num_samples ÷ num_chains, num_chains)
xs_chain = permutedims(Array(chain))

# Rejection sampling.
d = MvNormal(μ, Diagonal(σ²))
xs_exact = mapreduce(hcat, 1:num_samples) do _
    while true
        xs = rand(d)
        issorted(xs) && return xs
    end
end

qts = [0.05, 0.25, 0.5, 0.75, 0.95]
qt_names = map(q -> Symbol("q$(Int(100 * q))"), qts)
stats_with_mcses = (
    Tuple(qt_names) => Base.Fix2(quantile, qts),
    (Symbol("$(qn)_mcse") => (x -> mcse(x; kind=Base.Fix2(quantile, q))) for (q, qn) in zip(qts, qt_names))...,
)
```

Comparing the quantiles, they're generally within 3 MCSEs of each other. Not really sure what the cut-off should be, but given that the MCSEs are themselves estimated using a method that uses an asymptotic approximation, I'd guess this is good enough:
```julia
julia> PosteriorStats.summarize(reshape(xs_chain', :, 1, 2), stats_with_mcses...)
SummaryStats
        q5     q25     q50      q75    q95  q5_mcse  q25_mcse  q50_mcse  q75_mcse  q95_mcse 
 1  -1.961  -1.110  -0.546  -1.e-04  0.757   0.0042    0.0022    0.0017    0.0015    0.0021
 2  -0.761  -0.001   0.544   1.107   1.955   0.0020    0.0014    0.0013    0.0015    0.0027

julia> PosteriorStats.summarize(reshape(xs_exact', :, 1, 2), stats_with_mcses...)
SummaryStats
        q5     q25     q50      q75    q95  q5_mcse  q25_mcse  q50_mcse  q75_mcse  q95_mcse 
 1  -1.955  -1.109  -0.546  -0.0003  0.763   0.0021    0.0012    0.0010    0.0011    0.0017
 2  -0.761  -0.001   0.544   1.107   1.954   0.0017    0.0011    0.0010    0.0012    0.0017
```

src/bijectors/ordered.jl
```diff
@@ -88,3 +78,72 @@ end
...
+    for i in 2:length(x)
+        x_prev = x[i - 1]
+        d_i = marginal_dist(d.dist, i)
+        success = false
+        while !success
+            x[i] = rand(rng, d_i)
+            success |= x[i] >= x_prev
+        end
+    end
+    return x
```
sethaxen, line 148: This implementation seems to assume a `Product` structure. For `MvNormal` it would certainly do the wrong thing.

Why not just use this for the generic case:
```suggestion
    # Rejection sampling.
    while true
        x = rand(rng, d.dist)
        issorted(x) && return x
    end
```

And then specialize for `ProductDistribution`?

src/bijectors/ordered.jl
```diff
@@ -88,3 +78,72 @@ end
 
 logabsdetjac(b::OrderedBijector, x::AbstractVector) = sum(@view(x[2:end]))
 logabsdetjac(b::OrderedBijector, x::AbstractMatrix) = vec(sum(@view(x[2:end, :]); dims=1))
+
+# Need a custom distribution type to handle this properly.
+struct OrderedDistribution{B,D<:ContinuousMultivariateDistribution} <:
```
sethaxen, line 83: Maybe the types should be ordered the same as the fields:
```suggestion
struct OrderedDistribution{D<:ContinuousMultivariateDistribution,B} <:
```

src/bijectors/ordered.jl
```diff
@@ -88,3 +78,72 @@ end
...
+        ordered_b = binv ∘ OrderedBijector() ∘ b
+    else
+        throw(ArgumentError("ordered transform is currently not supported for $d."))
+    end
+
+    return OrderedDistribution(d, inverse(ordered_b))
+end
+
+bijector(d::OrderedDistribution) = d.bijector
+
```
sethaxen, line 116: You might also need this:
```suggestion

Base.eltype(::Type{<:OrderedDistribution{D}) where {D} = eltype(D)
Base.eltype(d::OrderedDistribution) = eltype(d.dist)
```

src/bijectors/ordered.jl
```diff
@@ -88,3 +78,72 @@ end
...
+function Distributions._logpdf(d::OrderedDistribution, x::AbstractVector{<:Real})
+    issorted(x) || return -eltype(d)(Inf)
+    return Distributions.logpdf(d.dist, x)
```
sethaxen, line 119: I would suggest this to avoid any heuristics for determining the return type (`eltype` is related to the eltype of random draws and is unrelated to the type of the return value of `logpdf`):
```suggestion
    lp = Distributions.logpdf(d.dist, x)
    issorted(x) && return lp
    return oftype(lp, -Inf)
```

src/bijectors/ordered.jl
```diff
@@ -88,3 +78,72 @@ end
...
+    binv = inverse(b)
+    if is_monotonically_decreasing(binv)
+        ordered_b = binv ∘ SignFlip() ∘ OrderedBijector() ∘ SignFlip() ∘ b
+    elseif is_monotonically_increasing(binv)
+        ordered_b = binv ∘ OrderedBijector() ∘ b
+    else
+        throw(ArgumentError("ordered transform is currently not supported for $d."))
+    end
+
+    return OrderedDistribution(d, inverse(ordered_b))
```
sethaxen, line 112: Maybe manually do the inversion?
```suggestion
        ordered_b = binv ∘ SignFlip() ∘ inverse(OrderedBijector()) ∘ SignFlip() ∘ b
    elseif is_monotonically_increasing(binv)
        ordered_b = binv ∘ inverse(OrderedBijector()) ∘ b
    else
        throw(ArgumentError("ordered transform is currently not supported for $d."))
    end

    return OrderedDistribution(d, ordered_b)
```

sethaxen (changes requested):

test/bijectors/ordered.jl
sethaxen: Ah, didn't see this. The issue looks to be that `rand!` is wrong for `OrderedDistribution`. Replacing it with the generic fallback defined below fixes the test.

src/bijectors/ordered.jl
torfjelde: My target was incorrect :man_facepalming: I implemented sequential rejection sampling, i.e. sample `x[1]` and then conditionally on this sample ordered vector using the marginals. This ofc leads to a different distribution.

Using the correct rejection sampling approach now:)

torfjelde: Further progress: it now works for most things but getting correctness issues for product dist with negative support (see tests).

torfjelde: Oh I just saw all your comments below :man_facepalming:

src/bijectors/ordered.jl
torfjelde: Isn't this missing also re-ordering the application of `b` and `binv`?

sethaxen: No, because in addition to inverting each individual bijector, one also must reverse the order of composition. These cancel out.

torfjelde: Ah ofc, I only thought about changing the order, not inverting them :facepalm:

torfjelde:
Damn. Seems like we missed something in #313

torfjelde:
Note that there doesn't seem to be anything incorrect with the impl, but it's failing because it's trying to compare elements which aren't part of the triangular part

torfjelde:
Pfft well that was painful. Added comments regarding what the issue is + fixed it by introducing a wrapper to avoid comparing `Matrix` values with potentially `undef` entries.

Would you have a quick look at some point @sethaxen ? :pray: Think we're there now after we've addressed the following:)

> BUT one final thing: what should we put as a warning regarding usage of ordered @sethaxen? I'm still a bit uncertain about exactly what you meant; I thought I understood wrt. restriction and not accounting for normalization constant, but then the example with changing normalization constant (variance parameter changing in a MvNormal) seemed to work, so now I'm confused again 🤷

sethaxen:
> Would you have a quick look at some point @sethaxen ? 🙏 Think we're there now after we've addressed the following:)

Cool will try to review this evening.
 
> BUT one final thing: what should we put as a warning regarding usage of ordered @sethaxen? I'm still a bit uncertain about exactly what you meant; I thought I understood wrt. restriction and not accounting for normalization constant, but then the example with changing normalization constant (variance parameter changing in a MvNormal) seemed to work, so now I'm confused again 🤷

Sounds weird, I'll check the test.

torfjelde:
> Sounds weird, I'll check the test.

I was talking about the example that we were discussing in one of the other comments; specifically https://github.com/TuringLang/Bijectors.jl/pull/297#discussion_r1597708555

sethaxen:
Ah, that's expected though. I assume you un-fixed the variance parameter and randomly sampled it within the rejection sampling inner loop? The issue here is that when the mean is the same for both components, then the variance actually has no impact on whether they are ordered. I think you should see a difference if you make the mean a reverse-ordered vector. The further the two mean components, the more pronounced the difference and the harder it is to rejection sample.

torfjelde:
the example I talk about is not related to rejection sampling; I'm referring to the example you ran with `NUTS`:)

sethaxen:
TBH I'm not certain if the above examples are even correct. The place I expect this to manifest is when conditioning. Which is implicitly what the rejection-sampling approach is doing (conditioning on `x[1] > x[2]`).

torfjelde:
> I'm confused which example you're referring to then. The one in the comment you linked to compares NUTS with rejection sampling, but it does so with fixed variance, so it would not manifest the issue I'm talking about.

Completely missed the fact that we were fixing the variance :facepalm: 

> Note that the rejection sampling approach makes sense. The quantiles of the two variances should be about the same, since to get an ordered draw with a well-separated reverse-ordered mean, one needs to increase the variance, but it doesn't matter which variance is increased.

Gotcha, gotcha; understand better now :+1: 

Soooo how do we summarize all this into a simple warning for the end-user? :eyes

torfjelde:
This PR is just waiting for the following:

> Soooo how do we summarize all this into a simple warning for the end-user? 

Think it's worth waiting until @sethaxen is back to let him have a final say before we merge :+1:

sethaxen:
> Soooo how do we summarize all this into a simple warning for the end-user? 👀

Maybe an admonition saying something like:

> The resulting ordered distribution is un-normalized. This is not a problem if used in a context where the normalizing factor is irrelevant, but if the value of the normalizing factor impacts the resulting computation, the results may be inaccurate. For example, if the distribution is used in sampling a posterior distribution with MCMC and the parameters of the ordered distribution are themselves sampled, then the normalizing factor would in general be needed for accurate sampling, and `ordered` should not be used. However, if the parameters are fixed, then since MCMC does not require distributions be normalized, `ordered` may be used without problems. A common case is where the distribution being ordered is a joint distribution of `n` identical univariate distributions. In this case the normalization factor works out to be the constant `1/n!`, and `ordered` can again be used without problems even if the parameters of the univariate distribution are sampled.

Not in love with it; feels too wordy.

src/bijectors/ordered.jl
```diff
@@ -9,24 +17,6 @@ A bijector mapping ordered vectors in ℝᵈ to unordered vectors in ℝᵈ.
 """
 struct OrderedBijector <: Bijector end
 
-"""
```
sethaxen, line 12: Was just looking at the docstring for `OrderedBijector`, and it states that it maps from ordered to unordered, but isn't the opposite the case?

```julia
julia> transform(Bijectors.OrderedBijector(), reverse(1.0:5))
5-element Vector{Float64}:
  5.0
 59.598150033144236
 79.68368695633191
 87.07274305526256
 89.7910248837216

julia> transform(inverse(Bijectors.OrderedBijector()), reverse(1.0:5))
ERROR: DomainError with -1.0:
```

torfjelde: Woah, true!

torfjelde:
>    Not in love with it; feels too wordy.

Added it, but did broke it up a bit + added a shorter warning to the initial parts of the docstring:) Thanks @sethaxen !

sethaxen (approved):
LGTM! Thanks @torfjelde for taking on this fix!

PR #302, opened 2024-04-15
Bugfix for `SimplexBijector` on `Matrix`

torfjelde:
Related: https://github.com/TuringLang/Turing.jl/issues/2190

### DIFF
--- a/src/bijectors/simplex.jl
+++ b/src/bijectors/simplex.jl
@@ -136,6 +136,12 @@ function logabsdetjac(b::SimplexBijector, x::AbstractVector{T}) where {T}
 
     return -lp
 end
+
+# Needed to avoid falling back to `with_logabsdet_jacobian` for matrix inputs.
+function logabsdetjac(b::SimplexBijector, x::AbstractMatrix{<:Real})
+    return sum(Base.Fix1(logabsdetjac, b), eachcol(x))
+end
+
 function simplex_logabsdetjac_gradient(x::AbstractVector)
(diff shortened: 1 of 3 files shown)


torfjelde:
@yebai @devmotion should be a quick merge

src/bijectors/simplex.jl
```diff
@@ -136,6 +136,12 @@ function logabsdetjac(b::SimplexBijector, x::AbstractVector{T}) where {T}
 
     return -lp
 end
+
+# Needed to avoid falling back to `with_logabsdet_jacobian` for matrix inputs.
```
devmotion, line 140: Which calls `logabsdetjac` again causing a stackoverflow error (had to check the code to realize this is the underlying problem).

devmotion (approved):
Looks good to me (the problem was not clear to me initially but I think I understood now).

The PR made me wonder whether it would be possible to improve performance of `with_logabsdet_jacobian` for `SimplexBijector` by not performing `transform` and `logabsdetjac` separately when both are requested. Doesn't block this PR and maybe would lead to more code duplications though.

torfjelde:
> The PR made me wonder whether it would be possible to improve performance of with_logabsdet_jacobian for SimplexBijector by not performing transform and logabsdetjac separately when both are requested. Doesn't block this PR and maybe would lead to more code duplications though.

Definitively! But yeah, probs for another PR (I'll make issue) :+1:

PR #305, opened 2024-04-19
Allow different types for upper and lower bound in `Logit`

torfjelde:
Currently the lower bound and upper bound are forced to be the same type, which removes the possibility of computing gradients wrt. just one of the parameter.

This PR just removes this restriction.

### DIFF
--- a/src/bijectors/logit.jl
+++ b/src/bijectors/logit.jl
@@ -1,9 +1,9 @@
 ######################
 # Logit and Logistic #
 ######################
-struct Logit{T} <: Bijector
-    a::T
-    b::T
+struct Logit{T1,T2} <: Bijector
+    a::T1
+    b::T2
 end
 
 Functors.@functor Logit


torfjelde:
Good point :thinking: But doesn't converting both into, say, vector of duals lead to perf implications since subsequent computations then also computes derivatives wrt. this?

devmotion:
Subsequent computations will be duals anyway if one of the fields are duals. So the only difference is that in one case the subsequent computation `f(a, b)` is started with both `a` and `b` being duals (but one of them having 0 partials) and in the other case one of them is a dual and the other one is non-dual but will be promoted to duals with 0 partials in every computation inside of `f` involving duals.

torfjelde:
Only every computation in which they're both related, no? For example, in the computation `(b - x) / (b - a)`, if only `a` is duals you'd unnecessarily also compute partials through `b - x` ?

But yeah, I do agree that it probably doesn't matter that much. Happy to accept a PR with proper promotion instead :+1:

Issue #301, opened 2024-03-08
Fixes to correlation bijectors


sethaxen:
The current bijectors for correlation matrices and their Cholesky factors when used in Turing are still quite brittle. A model just drawing from a uniform `LKJCholesky` regularly errors:

```julia
julia> using Turing

julia> @model function model(K, eta)
           x ~ LKJCholesky(K, eta, 'U')
       end;

julia> chns = sample(model(5, 1.0), NUTS(), 10_000)
┌ Info: Found initial step size
└   ϵ = 1.6
Sampling 100%|██████████████████████████████████████| Time: 0:00:04
ERROR: DomainError with -1.0798579785920026e-7:
sqrt was called with a negative real argument but will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
DomainError detected in the user `f` function. This occurs when the domain of a function is violated.
For example, `log(-1.0)` is undefined because `log` of a real number is defined to only output real
numbers, but `log` of a negative number is complex valued and therefore Julia throws a DomainError
by default. Cases to be aware of include:

* `log(x)`, `sqrt(x)`, `cbrt(x)`, etc. where `x<0`
* `x^y` for `x<0` floating point `y` (example: `(-1.0)^(1/2) == im`)

Within the context of SciML, this error can occur within the solver process even if the domain constraint
would not be violated in the solution due to adaptivity. For example, an ODE solver or optimization
routine may check a step at `new_u` which violates the domain constraint, and if violated reject the
step and use a smaller `dt`. However, the throwing of this error will have halted the solving process.

Thus the recommended fix is to replace this function with the equivalent ones from NaNMath.jl
(https://github.com/JuliaMath/NaNMath.jl) which returns a NaN instead of an error. The solver will then
effectively use the NaN within the error control routines to reject the out of bounds step. Additionally,
one could perform a domain transformation on the variables so that such an issue does not occur in the
definition of `f`.

For more information, check out the following FAQ page:
https://docs.sciml.ai/Optimization/stable/API/FAQ/#The-Solver-Seems-to-Violate-Constraints-During-the-Optimization,-Causing-DomainErrors,-What-Can-I-Do-About-That?

Stacktrace:
  [1] throw_complex_domainerror(f::Symbol, x::Float64)
    @ Base.Math ./math.jl:33
  [2] sqrt
    @ ./math.jl:686 [inlined]
  [3] sqrt
[...]

PR #313, opened 2024-05-27
Numerical improvements to correlation bijectors

sethaxen:
This PR implements the numerical suggestions in #301. It does not make any of the suggested renaming changes, which will be left for a future PR.


torfjelde (changes requested):
This will be really nice @sethaxen :)

I've added a few comments. It also seems as if the chain rule is somehow not type stable? :confused:

ext/BijectorsReverseDiffExt.jl
```diff
@@ -268,7 +268,6 @@ end
 @grad_from_chainrules _link_chol_lkj(x::TrackedMatrix)
 @grad_from_chainrules _link_chol_lkj_from_upper(x::TrackedMatrix)
 @grad_from_chainrules _link_chol_lkj_from_lower(x::TrackedMatrix)
-@grad_from_chainrules _inv_link_chol_lkj(x::TrackedVector)
```
torfjelde, line 271: Also, is this intentional?

src/chainrules.jl
```diff
@@ -267,55 +267,14 @@ function ChainRulesCore.rrule(::typeof(_link_chol_lkj_from_lower), W::AbstractMa
...
-
-                Δz = ΔW[i - 1, j] * tmp - Δtmp * tmp / sqrt(1 - z^2) * z
-                Δy[idx] = Δz / cosh(y[idx])^2
-                Δtmp = ΔW[i - 1, j] * z + Δtmp * sqrt(1 - z^2)
-            end
-        end
+    W_logJ, back = _inv_link_chol_lkj_rrule(y)
 
+    function pullback_inv_link_chol_lkj(ΔW_ΔlogJ)
+        Δy = back(ChainRulesCore.unthunk(ΔW_ΔlogJ))
```
torfjelde, line 273: Is this closure causing type stability issues? o.O Seems somewhat surprising but not impossible

torfjelde: Nah, just checked and it seems to be the `_inv_link_chol_lkj_rrule`

torfjelde: Uhmm interesting.

`idx` used as a variable inside the closure in `_inv_link_chol_lkj_rrule` is apparently making it more difficult to infer o.O See my comment on the pullback definition.

src/bijectors/corr.jl
```diff
@@ -333,47 +339,121 @@ _link_chol_lkj_from_lower(W::AbstractMatrix) = _link_chol_lkj_from_upper(transpo
...
+        @inbounds for j in K:-1:2
+            Δlog_remainder = W[j, j] * ΔW[j, j] + 2ΔlogJ
+            for i in (j - 1):-1:1
+                W_ΔW = W[i, j] * ΔW[i, j]
+                z = z_vec[idx]
+                Δy[idx] = (inv(z) - z) * W_ΔW - z * Δlog_remainder
+                idx -= 1
+                Δlog_remainder += ΔlogJ + W_ΔW
+            end
+        end
```
torfjelde, line 440: Replacing `idx` with `idx_local` fixes the type-stability issue o.O

```suggestion
    function pullback_inv_link_chol_lkj((ΔW, ΔlogJ))
        LinearAlgebra.require_one_based_indexing(ΔW)
        Δy = similar(y)

        idx_local = lastindex(y)
        @inbounds for j in K:-1:2
            Δlog_remainder = W[j, j] * ΔW[j, j] + 2ΔlogJ
            for i in (j - 1):-1:1
                W_ΔW = W[i, j] * ΔW[i, j]
                z = z_vec[idx_local]
                Δy[idx_local] = (inv(z) - z) * W_ΔW - z * Δlog_remainder
                idx_local -= 1
                Δlog_remainder += ΔlogJ + W_ΔW
            end
        end
```

torfjelde: In hindsight, I guess it's not that confusing. From https://docs.julialang.org/en/v1/manual/variables-and-scoping/#local-scope (emphasis mine)

> The next clear cut situation we'll consider is when there is already a local variable named x, in which case x = <value> **always assigns to this existing local x**. This is true whether the assignment occurs in the same local scope, an inner local scope in the same function body, or in the body of a function nested inside of another function, also known as a [closure](https://en.wikipedia.org/wiki/Closure_(computer_programming)).

torfjelde: Makes sense when you think about scenarios such as:
```julia
function f(x)
    function g(y)
        x = 1
        return y + x
    end

    return g, y -> y + x
end

g, h = f(3)
h(1) # => 4
g(1) # => 2
h(1) # => 2 (because `x = 1` when we call `g`)
```

src/bijectors/corr.jl
```diff
@@ -281,48 +291,44 @@ which is the above implementation.
...
         idx += 1
-        tmp = sqrt(1 - W[1, j]^2)
+        remainder_sq = 1 - W[1, j]^2
         for i in 2:(j - 1)
-            p = W[i, j] / tmp
-            tmp *= sqrt(1 - p^2)
-            z[idx] = atanh(p)
+            z = W[i, j] / sqrt(remainder_sq)
+            y[idx] = atanh(z)
+            remainder_sq -= W[i, j]^2
```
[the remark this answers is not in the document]
sethaxen: Yes, that's right. Here I've done 2 things: rename `z` to `y` for consistency with the other methods (I should also update the docstring of `_link_chol_lkj` accordingly) and replace the `tmp`/`p` with `remainder_sq`. The latter change is both more readable (`remainder_sq` is the square norm remaining for the unit vector) and also I believe accumulating via addition instead of multiplication will be more numerically stable. Mathematically they are the same.

src/bijectors/corr.jl
```diff
@@ -333,47 +339,121 @@ _link_chol_lkj_from_lower(W::AbstractMatrix) = _link_chol_lkj_from_upper(transpo
...
-        W[1, j] = 1
-        for i in 2:j
-            z = tanh(Y[i - 1, j])
-            tmp = W[i - 1, j]
-            W[i - 1, j] = z * tmp
-            W[i, j] = tmp * sqrt(1 - z^2)
+        log_remainder = zero(T)  # log of proportion of unit vector remaining
+        for i in 1:(j - 1)
+            z = tanh(Y[i, j])
+            idx += 1
```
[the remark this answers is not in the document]
sethaxen: Ah good catch, it's a carryover from copying this from the vector version
```suggestion
```

src/bijectors/corr.jl
```diff
@@ -333,47 +339,121 @@ _link_chol_lkj_from_lower(W::AbstractMatrix) = _link_chol_lkj_from_upper(transpo
 Inverse link function for cholesky factor.
 """
 function _inv_link_chol_lkj(Y::AbstractMatrix)
+    LinearAlgebra.require_one_based_indexing(Y)
     K = LinearAlgebra.checksquare(Y)
 
     W = similar(Y)
+    T = typeof(log(one(eltype(W))))
```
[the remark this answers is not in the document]
sethaxen: I guess not strictly necessary. Perhaps safer if someone passes some `W` with dimensional eltype? But probably the whole method doesn't work for such types, so we could use `float`.
```suggestion
    T = float(eltype(W))
```

src/bijectors/corr.jl
```diff
@@ -333,47 +339,121 @@ _link_chol_lkj_from_lower(W::AbstractMatrix) = _link_chol_lkj_from_upper(transpo
...
-    return W
+    return W, logJ
 end
 
 function _inv_link_chol_lkj(y::AbstractVector)
+    LinearAlgebra.require_one_based_indexing(y)
     K = _triu1_dim_from_length(length(y))
 
     W = similar(y, K, K)
+    T = typeof(log(one(eltype(W))))
```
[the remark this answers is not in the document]
sethaxen:
```suggestion
    T = float(eltype(W))
```

ext/BijectorsReverseDiffExt.jl
sethaxen: Yes, the problem is that neither `@grad` nor `@grad_from_chainrules` supports multi-output functions (https://github.com/JuliaDiff/ReverseDiff.jl/issues/221), so we cannot use this macro. At the same time, nothing in the function should not be AD-able by ReverseDiff, so I just removed the rule.

However, we have the same problem with Tracker. `Tracker.@grad` seems to not support multi-output functions, and I'm still working out how to AD through the primal (I have an idea for a fix).

sethaxen: If you have an idea how to get this working for ReverseDiff, let me know. It would be great to use the manual pullback.

sethaxen: Nope, my Tracker idea did not work.

sethaxen: Nevermind, it works!

src/bijectors/corr.jl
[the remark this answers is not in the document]
torfjelde: Ah, missed the `/ p` part; cheers!

ext/BijectorsReverseDiffExt.jl
torfjelde: Lovely:)

torfjelde: Actually, how did you achieve it?

torfjelde (approved):
Lovely stuff:) Happy with merging as soon as tests pass (which should be a quick merge with master after #314 )

ext/BijectorsReverseDiffExt.jl
sethaxen: The changes in this commit: https://github.com/TuringLang/Bijectors.jl/pull/313/commits/a2eac95e47b1ff235a8f5699e888f962cc241eaa . In Tracker, the cotangent of multi-output functions ends up being a `TrackedTuple`, which doesn't support iteration, so instead use indexing to split the tuple. And also make `pd_from_lower` and `pd_from_upper` use the same tricks as `cholesky_upper` and `cholesky_lower`.

sethaxen:
I don't think #314 was completed before merging. Its CI was still failing with a similar error.

torfjelde:
Yes, #314 was indeed not ready for a merge @yebai ; why was it merged?

ext/BijectorsReverseDiffExt.jl
torfjelde: Uuuuh I didn't know that _that_ was the reason why it was an issue! Dopey:)

torfjelde:
@sethaxen I just pushed the fix directly to this branch. Let's see if CI succeeds now:)

torfjelde:
Damn, even this doesn't work because `eachslice` only supports a single value as the `dims` arg :confused: 

Really sorry about this @sethaxen ; this bug was hidden behind an unrelated numerical issue that caused this particular test to never be run on 1.6. But the cause of this shouldn't have been merged.

I'll just disable those tests on this PR and then we'll have to fix it in a separate PR.

sethaxen:
Seems like it worked!

torfjelde:
Lovely! Feel free to hit the big green button:)

sethaxen:
Sadly, I am not an "authorized user," and the button is gray.

torfjelde:
Want me to do it then? Also happy to let give you authorization given your involvement in Bijectors.jl if you want to:)

sethaxen:
> Want me to do it then? Also happy to let give you authorization given your involvement in Bijectors.jl if you want to:)

Sure to both!

torfjelde:
Done:) Wonderful stuff @sethaxen ; thanks!

PR #314, opened 2024-06-03
Fix `ProductBijector` on Julia <1.9

torfjelde:
Ref: https://github.com/TuringLang/Bijectors.jl/pull/313#issuecomment-2145177048


torfjelde:
@sethaxen Tests are failing because of numerical issues with cholesky transforms, so I'm making the PR directly to your branch instead. Happy with this?

sethaxen:
Sure that's fine

PR #318, opened 2024-06-26
Test against Enzyme

mhauru:
Currently failing, using this to run CI and catch issues. Uses code from https://github.com/TuringLang/Bijectors.jl/pull/240.


mhauru:
CI currently fails with
```
julia: /buildworker/worker/package_linux64/build/src/llvm-late-gc-lowering.cpp:872: std::vector<int, std::allocator<int> > LateLowerGCFrame::NumberAllBase(State&, llvm::Value*): Assertion `Tracked.size() == BaseNumbers.size()' failed.
```

Running the tests locally I also see the following
```
  Expression: ≈(Enzyme.gradient(Enzyme.Reverse, f, x), finitediff, rtol = rtol, atol = atol)
  Enzyme compilation failed due to illegal type analysis.
  Current scope:
  ; Function Attrs: mustprogress willreturn
  define internal fastcc "enzyme_type"="{[-1]:Float@double}" double @preprocess_julia__solve__70_20557({ { [3 x double] }, [4 x double], { i64, i8 } } addrspace(11)* nocapture noundef nonnull readonly align 8 dereferenceable(72) "enzyme_type"="{[-1]:Pointer, [-1,0]:Float@double, [-1,8]:Float@double, [-1,16]:Float@double, [-1,24]:Float@double, [-1,32]:Float@double, [-1,40]:Float@double, [-1,48]:Float@double, [-1,56]:Integer, [-1,57]:Integer, [-1,58]:Integer, [-1,59]:Integer, [-1,60]:Integer, [-1,61]:Integer, [-1,62]:Integer, [-1,63]:Integer, [-1,64]:Integer}" "enzymejl_parmtype"="4511001616" "enzymejl_parmtype_ref"="1" %0) unnamed_addr #99 !dbg !10218 {
  [blahblahblah]
```

This is with Enzyme 0.12.19. @wsmoses, could you let me know if these are issues familiar to you, or if I should work to make a minimal reproduction of them?

wsmoses:
What is the Julia version of the first error. That looks like a Julia compiler bug that was subsequently fixed

wsmoses:
The latter issue is not known please open an issue with a mwe

mhauru:
The first one is on Julia 1.6, so the compiler bug it probably is then. I'll make CI run only on latest release and retry, and try to make an MWE of the latter.

mhauru:
Illegal type analysis error issue here: https://github.com/EnzymeAD/Enzyme.jl/issues/1573

mhauru:
The `Assertion failed: (vec.back() == currentBlock)` issue: https://github.com/EnzymeAD/Enzyme.jl/issues/1574

mhauru:
If I get rid of the `Pkg.add(Pkg.PackageSpec(; url="https://github.com/simsurace/Enzyme.jl.git", rev="fix-cholesky"))` and rather just get the latest Enzyme.jl release I run into https://github.com/EnzymeAD/Enzyme.jl/issues/1574, so will wait for that to be sorted first.

wsmoses:
@mhauru like I mentioned on the closed Enzyme issue, the illegal type analysis is thrown due to the presence of an unknown bithack in Roots.jl , rather than risk an incorrect answer: https://github.com/JuliaMath/Roots.jl/blob/4263c7683423af209af1879bd9cd223b0ba55acd/src/Bracketing/bisection.jl#L135

It looks like you all defined a custom rule for tapir/etc for find_alpha for it, which I presume is a reasonable place for a custom rule for it here

@willtebbutt @mhauru @yebai mind importing the same rule in Enzyme. Feel free to use Enzyme's chain rule import macro if you like: https://github.com/EnzymeAD/Enzyme.jl/blob/c4068fc96a9b574fd08593178611b153cb71ac5b/ext/EnzymeChainRulesCoreExt.jl#L126 If you test with both forward and reverse you'll need to import both forward and reverse rules

devmotion:
> It looks like you all defined a custom rule for tapir/etc for find_alpha for it, which I presume is a reasonable place for a custom rule for it here

Even if Enzyme would not throw an error, one really should define a custom rule for `find_alpha` (it's very simple) instead of differentiating through the root finding algorithm.

More generally, similar to the ForwardDiff support IMO it would be good to add support for Enzyme to Roots.jl such that the implicit function theorem is always exploited.

wsmoses:
Sure, but I definitely appreciate that Enzyme early errors rather than risking bad behavior here!

Since it looks like this repo already adds custom rules for this, can that be added here to start and after tests are validated possibly upstream to roots.jl

mhauru:
Thanks @wsmoses, I added an import of the existing ChainRule for `find_alpha`. This uncovered a new issue, filed here: https://github.com/EnzymeAD/Enzyme.jl/issues/1577

ext/BijectorsEnzymeExt.jl
```diff
@@ -0,0 +1,14 @@
...
+if isdefined(Base, :get_extension)
+    using Enzyme: @import_frule, @import_rrule
+    using Bijectors: find_alpha
+else
+    using ..Tapir: @import_frule, @import_rrule
+    using ..Bijectors: find_alpha
+end
+
+@import_rrule typeof(find_alpha) Float64 Float64 Float64
+@import_frule typeof(find_alpha) Float64 Float64 Float64
```
devmotion, line 12: Same comment as in https://github.com/TuringLang/Bijectors.jl/pull/319#discussion_r1657123613: I think this is too restrictive.

ext/BijectorsEnzymeExt.jl
```diff
@@ -0,0 +1,14 @@
+module BijectorsEnzymeExt
+
+if isdefined(Base, :get_extension)
+    using Enzyme: @import_frule, @import_rrule
+    using Bijectors: find_alpha
+else
+    using ..Tapir: @import_frule, @import_rrule
```
devmotion, line 7:
```suggestion
    using ..Enzyme: @import_frule, @import_rrule
```

ext/BijectorsEnzymeExt.jl
wsmoses: Yeah just make this real or whatever the original fn is defined on.

No restrictions on support from enzyme side

test/runtests.jl
```diff
@@ -29,6 +30,11 @@ using ChangesOfVariables: ChangesOfVariables
 using InverseFunctions: InverseFunctions
 using LazyArrays: LazyArrays
 
+# TODO(mhauru) Remove this once Enzyme is fixed?
```
wsmoses, line 33: This is the default and shouldn't be needed now

ext/BijectorsEnzymeExt.jl
mhauru: Done

ext/BijectorsEnzymeExt.jl
mhauru: Whooops... Done

test/runtests.jl
mhauru: Removed

test/ad/utils.jl
```diff
@@ -3,6 +3,7 @@ const AD = get(ENV, "AD", "All")
 
 function test_ad(f, x, broken=(); rtol=1e-6, atol=1e-6)
     finitediff = FiniteDifferences.grad(central_fdm(5, 1), f, x)[1]
+    et = eltype(finitediff)
```
devmotion, line 6: Why is this needed? Shouldn't Enzyme return the correct types automatically?

mhauru: In forward mode it returns tuples, and if the gradient is empty, the result is a `Tuple{}`. This resulted in comparing an empty `Float64[]` to an empty `Union{}[]`, which failed. See https://github.com/EnzymeAD/Enzyme.jl/issues/1584

devmotion: But the gradient should never be empty? Such a test would be quite useless, so I assume we don't run into this special case here? So maybe a simple `collect` (without specifying the element type) would be sufficient?

devmotion: I'd suggest not testing AD in the case `d = 1` - or just checking that the gradient is empty. We already handle this case in a special way in e.g. https://github.com/TuringLang/Bijectors.jl/blob/86544752e41a6a4aab0e1607fbf1e53a8c548a64/test/bijectors/corr.jl#L32-L33.

mhauru: It passes now though, and I don't really see a downside to testing it? Good to know for instance that nothing crashes even if you hit this corner case.

devmotion: No, I also would prefer to not remove the test completely (even though I think it's of very limited use)

> just checking that the gradient is empty

wsmoses:
since it looks like all of this passes, can this get merged? @devmotion @yebai @mhauru

test/ad/utils.jl
```diff
@@ -47,7 +47,7 @@ function test_ad(f, x, broken=(); rtol=1e-6, atol=1e-6)
         end
     end
 
-    if AD == "All" || AD == "Enzyme"
+    if (AD == "All" || AD == "Enzyme") && VERSION >= v"1.10"
         if :EnzymeReverse in broken
```
wsmoses, line 51: The Julia compiler bug on 1.6 only applies to [batched] fwd mode. It shouldn't be hit for reverse mode if you want to losen the tests

mhauru: Nice, changed that. Is `forward_broken = :EnzymeForward in broken || :Enzyme in broken || VERSION <= v"1.6"` correct, or did the compiler bug get fixed only later?

test/ad/utils.jl
devmotion: I meant something differently - removing the `eltype`/`collect` completely and only add a special case to this weird test of the `CorrBijector` since we already have special cases for `d = 1` there anyway.

mhauru:
@wsmoses, one of the tests started failing when I enabled reverse mode Enzyme checks for Julia 1.6. Could you please check the CI logs and let me know if this is a known issue with Julia 1.6 and what the appropriate version bound would be: https://github.com/TuringLang/Bijectors.jl/actions/runs/9871431348/job/27259337160?pr=318

test/ad/utils.jl
mhauru: Yeah, that's how I understood you, but that would require something like adding another argument to `test_ad` called `expect_empty` that would skip comparing to finitediff and instead check that the gradient has length 1 (for all AD backends) and setting that argument to `d == 1`, which to me seems more complicated, with a bunch of `if` statements, compared adding the one-liner enforcing `eltype`. I can do it if you prefer it, I just don't see the benefit.

wsmoses:
Hm yeah I haven't seen that one before. Open an issue and will investigate.

In interim, maybe we just enable for 1.10+ for now and update bounds once that's fixed

mhauru:
Reported the latest issue here: https://github.com/EnzymeAD/Enzyme.jl/issues/1629 It gets triggered on Julia 1.6 and 1.7, but not 1.8.

src/interface.jl
```diff
@@ -93,7 +93,7 @@ function transform(t::Transform, x)
     res = with_logabsdet_jacobian(t, x)
     if res isa ChangesOfVariables.NoLogAbsDetJacobian
         error(
-            "`transform` not implemented for $(typeof(f)); implement `transform` and/or `with_logabsdet_jacobian`.",
+            "`transform` not implemented for $(typeof(t)); implement `transform` and/or `with_logabsdet_jacobian`.",
```
mhauru, line 96: This change has nothing to do with this PR, I just spotted the typo while working on this PR and didn't feel like making a separate one character PR.

mhauru:
If we are happy with Enzyme only being tested on 1.10, this could be ready to merge. I'm the owner, so someone else needs to review.

wsmoses:
> or so Julia can handle it, instead of aborting/exiting? In my view, abouting a Julia session should only be used in unrecoverabl

Oh yeah totally agreed, PR's welcome!

mhauru:
Done: https://github.com/EnzymeAD/Enzyme.jl/issues/1634

PR #319, opened 2024-06-26
Tapir.jl Usage

willtebbutt:
#318 made me intrigued to see how well Tapir.jl fares on Bijectors.jl. It looks like I just need to make use of Bijectors.jl rule for `find_alpha` (a wrapper for a bisection routine) in order to make the tests pass, which is nice.

I'm a little surprised to be honest -- are we confident that the AD tests have decent coverage of the functionality in Bijectors.jl?


willtebbutt:
@yebai okay, this could in principle now be merged. Since Tapir.jl can't be installed on 1.6, I've had to install it conditionally, depending on whether the Julia version is 1.10 or greater.

ext/BijectorsTapirExt.jl
```diff
@@ -0,0 +1,13 @@
...
+
+if isdefined(Base, :get_extension)
+    using Tapir: @from_rrule, MinimalCtx, Tapir
+    using Bijectors: find_alpha
+else
+    using ..Tapir: @from_rrule, MinimalCtx, Tapir
+    using ..Bijectors: find_alpha
+end
+
+@from_rrule MinimalCtx Tuple{typeof(find_alpha),Float64,Float64,Float64}
```
devmotion, line 11: Does this imply that Tapir will error for e.g. `Float32` inputs?

willtebbutt: Yes -- Tapir.jl doesn't _really_ support `Float32` in general at the minute. This is because we've not set up testing infrastructure for it, and therefore dont want to make any promises about it.

In this particular instance, the rule won't be hit if `Float32`s are used, Tapir.jl will recurse into the definition, and eventually hit the bitcast that meant that this rule is necessary in the `Float64` case.

devmotion: I see. I wonder if it would be worth it (and possible) to import (or write, it's very simple) the rule in a more generic way such that Bijectors is already ready once Tapir supports more generic number types?

willtebbutt: I could definitely just make it applicable to `Union{Float32, Float64}` to future proof it. Is that the kind of thing you have in mind?

devmotion: Yes, something like that. The `rrule` is defined for `Real` but I guess `Float32` and `Float64` are the most relevant types; possibly `Float16` could be interesting as well. I assume Tapir might never be expected to work with more generic types such as `Real` or `AbstractFloat`?

devmotion: Hmm, from a quick look at the `find_alpha` occurrences in the code it seems that at least the last argument could easily be something more generic such as `Int` since it's directly taken from the user input. For the first argument it's unlikely but in principle also possible. The second argument should always be a floating point number (since its computation involves `log1pexp`).

willtebbutt: > Yes, something like that. The rrule is defined for Real but I guess Float32 and Float64 are the most relevant types; possibly Float16 could be interesting as well.

Good point -- I should just extend it to `Base.IEEEFloat`.

> I assume Tapir might never be expected to work with more generic types such as Real or AbstractFloat?

Short answer: a concrete subtype of `AbstractFloat` is unlikely to work. A concrete subtype of `Real` is unlikely to work if it's differentiable (e.g. if it's `Int`-like or `Bool`-like then it might work, because it's not a differentiable type, and we can largely exclude it from computation).

Slightly longer answer: in order to make a new concrete subtype of `AbstractFloat` work, Tapir.jl would either have to
1. differentiate through the primal code that Julia generates to run the function on this new type, or
2. specify a custom `tangent_type` and ensure there are rules for everything.

1 isn't likely to work because, much of the time, you'll wind up with quite a bit of `bitcast`-ing going on to implement things (i.e. so that you can do bit-twiddling stuff to make computations efficient), and Tapir.jl definitely can't differentiate through that kind of thing, and will (should) complain loudly if you try. This is to avoid silently giving the wrong answer.

2 requires work.

So the longer answer is effectively: it could be made to work, but would require effort.

> Hmm, from a quick look at the find_alpha occurrences in the code it seems that at least the last argument could easily be something more generic such as Int since it's directly taken from the user input. For the first argument it's unlikely but in principle also possible.

Okay -- I can definitely add support for this. It'll just be a few more lines of code in the extension, and some extra tests somewhere.

willtebbutt: Okay. Changes included.

@devmotion I've added support for `Integer` final arguments, and subtypes of `IEEEFloat` as inputs. I'll add support for Integer first arguments at some point in the future -- I need to extend the scope of Tapir's `ChainRules.rrule` conversion macro to support automatically dropped gradients w.r.t. isbits non-differentiable types (like integers), so this should become more straightforward in the future.

willtebbutt:
Okay. I think I'm basically happy with this. Once we get the sign off from @devmotion , if someone could please merge this (I don't have commit rights).

ext/BijectorsTapirExt.jl
```diff
@@ -0,0 +1,40 @@
+module BijectorsTapirExt
+
+if isdefined(Base, :get_extension)
+    using Tapir: @is_primitive, MinimalCtx, Tapir, CoDual, primal, tangent_type, @from_rrule
+    using Bijectors: find_alpha
+    using ChainRulesCore: rrule
```
devmotion, line 6: It's necessary (or at least recommended?) to not load any other packages apart from the weak dependency and the parent package: https://github.com/JuliaStats/LogExpFunctions.jl/pull/63#issue-1571562080
```suggestion
    using Bijectors: find_alpha, ChainRulesCore
```

ext/BijectorsTapirExt.jl
```diff
@@ -0,0 +1,40 @@
...
+else
+    using ..Tapir: @is_primitive, MinimalCtx, Tapir, primal, tangent_type, @from_rrule
+    using ..Bijectors: find_alpha, rrule
+    using ..ChainRulesCore: rrule
```
devmotion, line 10:
```suggestion
    using ..Bijectors: find_alpha, ChainRulesCore
```

ext/BijectorsTapirExt.jl
```diff
@@ -0,0 +1,40 @@
...
+
+function Tapir.rrule!!(
+    ::CoDual{typeof(find_alpha)}, x::CoDual{P}, y::CoDual{P}, z::CoDual{I}
+) where {P<:Base.IEEEFloat,I<:Integer}
+    # Require that the integer is non-differentiable.
+    if tangent_type(I) != Tapir.NoTangent
+        msg = "Integer argument has tangent type $(tangent_type(I)), should be NoTangent."
+        throw(ArgumentError(msg))
+    end
+    out, pb = rrule(find_alpha, primal(x), primal(y), primal(z))
```
devmotion, line 32:
```suggestion
    out, pb = ChainRulesCore.rrule(find_alpha, primal(x), primal(y), primal(z))
```

test/ad/chainrules.jl
```diff
@@ -27,6 +27,21 @@ end
...
     test_rrule(Bijectors.find_alpha, x, y, z)
 
+    if @isdefined Tapir
+        Tapir.TestUtils.test_rrule!!(
+            Xoshiro(123), Bijectors.find_alpha, x, y, z; is_primitive=true, perf_flag=:none
+        )
+        Tapir.TestUtils.test_rrule!!(
+            Xoshiro(123), Bijectors.find_alpha, x, y, 3; is_primitive=true, perf_flag=:none
+        )
+        #! format: off
```
devmotion, line 37: Is there any reason to disable the formatter here?

willtebbutt: I was trying to avoid it splitting all of the arguments onto different lines -- in this case it feels harder to read. I'm going to re-do this line to avoid any of these issues though. Will let you know when done.

willtebbutt: Okay. @devmotion I've just changed the way I've formatting this to avoid the issue.

PR #322, opened 2024-08-06
Fix for product of `Dirichlet`

torfjelde:
`logpdf_with_trans` currently doesn't handle stuff like products of `Dirichlet`

Long-term we want to remove `logpdf_with_trans` completely, but for now this is holding up https://github.com/TuringLang/DynamicPPL.jl/pull/575 so seems worth providing a hotfix.

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -135,6 +135,11 @@ invlink(d::Distribution, y) = inverse(bijector(d))(y)
 
 # To still allow `logpdf_with_trans` to work with "batches" in a similar way
 # as `logpdf` can.
+
+# Default catch-all so we can work with distributions by default and batch-support can be
+# added when needed.
+_logabsdetjac_dist(d, x) = logabsdetjac(bijector(d), x)
+
 _logabsdetjac_dist(d::UnivariateDistribution, x::Real) = logabsdetjac(bijector(d), x)
 function _logabsdetjac_dist(d::UnivariateDistribution, x::AbstractArray)
     return logabsdetjac.((bijector(d),), x)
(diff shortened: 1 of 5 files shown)


torfjelde:
@yebai @sunxd3 @mhauru

src/Bijectors.jl
```diff
@@ -135,6 +135,11 @@ invlink(d::Distribution, y) = inverse(bijector(d))(y)
 
 # To still allow `logpdf_with_trans` to work with "batches" in a similar way
 # as `logpdf` can.
+
+# Default catch-all so we can work with distributions by default and batch-support can be
+# added when needed.
+_logabsdetjac_dist(d, x) = logabsdetjac(bijector(d), x)
```
devmotion, line 141: It seems (based on https://github.com/TuringLang/Bijectors.jl/blob/7c8b533e4563f5a89b53dddc069e029e655ee370/src/Bijectors.jl#L167) we could be a bit more restrictive:
```suggestion
_logabsdetjac_dist(d::Distribution, x) = logabsdetjac(bijector(d), x)
```

devmotion (approved):

mhauru:
@devmotion has a fair comment, but the code looks good to me otherwise. I'm confused by the CI failure, the new test passes for me locally.

torfjelde:
@devmotion good shout 👍 

> but the code looks good to me otherwise. I'm confused by the CI failure, the new test passes for me locally.

Seems to be a julia v1.6 thing. Looking at the deps used, I notice DistributionsAD is on 0.6.45 vs. 0.6.55 (on Julia latest). Miiight be it, though uncertain why. Lookin' into it.

mhauru:
There's `mapslices` that has been taking multiple dimensions for longer than `eachslice`, I think switching to using that could work, since we call `map` on the results of `eachslice`. I'm happy to give this a go given @torfjelde, unless you're already in the middle of solving it.

torfjelde:
Go for it!

mhauru:
`mapslices` doesn't accept multiple arguments the way `map` does, so I ended up doing some reshaping instead: #323

PR #325, opened 2024-08-17
Implement logabsdetjac for Inverse{<:TruncatedBijector} for better numerical stability

acertain:
Example of previous badness: `logabsdetjac(inverse(bijector(Uniform(-1,1))), 80) = -Inf (is now -79.30685281944005)`

Formula stolen from stan, I didn't check its correctness.


src/bijectors/truncated.jl
```diff
@@ -68,6 +68,27 @@ end
 
 with_logabsdet_jacobian(b::TruncatedBijector, x) = transform(b, x), logabsdetjac(b, x)
 
+# from stan https://github.com/stan-dev/math/blob/develop/stan/math/prim/constraint/lub_constrain.hpp etc
```
devmotion, line 71: Is this actually necessary (or helpful)? It's simply the logarithm of the absolute value of the (arguably straightforward) derivative of `truncated_invlink` above.
```suggestion
```

src/bijectors/truncated.jl
```diff
@@ -68,6 +68,27 @@ end
...
+function truncated_inv_logabsdetjac(x, a, b)
```
devmotion, line 72: Maybe follow the convention above and use `y` for the argument of functions of the inverse bijector such as `truncated_invlink`?
 ```suggestion
function truncated_inv_logabsdetjac(y, a, b)
```

File: src/bijectors/truncated.jl [hunk 3]
```diff
@@ -68,6 +68,27 @@ end
...
+    lowerbounded, upperbounded = isfinite(a), isfinite(b)
+    if lowerbounded && upperbounded
+        return log(b - a) - abs(x) + 2.0 * LogExpFunctions.log1pexp(-abs(x))
```
devmotion, line 75: Generally, one should avoid `Float64` literals since they can cause undesired promotions (e.g. for `Float32` arguments):
```suggestion
        abs_y = abs(y)
        return log(b - a) - abs_y + 2 * LogExpFunctions.log1pexp(-abs_y)
```

src/bijectors/truncated.jl
```diff
@@ -68,6 +68,27 @@ end
...
+    elseif lowerbounded
+        return x
+    elseif upperbounded
+        return x
```
devmotion, line 79: Generally, this branch can lead to type instabilities since in contrast to the first branch it does not depend on the types of `a` and `b` (e.g., consider `y::Float32`, `a::Int`, `b::Int`: the first branch returns a `Float64` whereas the other branches return a `Float32`; another example is `y::Int`, `a::Int`, `b::Int`).
There are multiple approaches to fix it, e.g., one could add a
```julia
   y, a, b = map(float, promote(y, a, b))
```
add the top of the function body.

These two branches can be simplified to
```suggestion
    elseif lowerbounded || upperbounded
        return y
```

src/bijectors/truncated.jl
```diff
@@ -68,6 +68,27 @@ end
...
+    else
+        return zero(x)
```
devmotion, line 81:
```suggestion
        return zero(y)
```

src/bijectors/truncated.jl
```diff
@@ -68,6 +68,27 @@ end
...
+    end
+end
+
+function logabsdetjac(ib::Inverse{<:TruncatedBijector}, x)
+    a, b = ib.orig.lb, ib.orig.ub
+    return truncated_inv_logabsdetjac.(x, a, b)
```
devmotion, line 87:
```suggestion
function logabsdetjac(ib::Inverse{<:TruncatedBijector}, y)
    a, b = ib.orig.lb, ib.orig.ub
    return truncated_inv_logabsdetjac.(y, a, b)
```

src/bijectors/truncated.jl
```diff
@@ -68,6 +68,27 @@ end
...
+end
+
+with_logabsdet_jacobian(ib::Inverse{<:TruncatedBijector}, x) = transform(ib, x), logabsdetjac(ib, x)
```
devmotion, line 90:
```suggestion
with_logabsdet_jacobian(ib::Inverse{<:TruncatedBijector}, y) = transform(ib, y), logabsdetjac(ib, y)
```

acertain:
@devmotion thanks for the suggestions, I've implemented them

src/bijectors/truncated.jl
```diff
@@ -68,6 +68,25 @@ end
...
 
+function truncated_inv_logabsdetjac(y, a, b)
+    lowerbounded, upperbounded = isfinite(a), isfinite(b)
+    if lowerbounded && upperbounded
+        abs_y = abs(y)
+        return log(b - a) - abs_y + 2 * LogExpFunctions.log1pexp(-abs_y)
+    elseif lowerbounded || upperbounded
+        return convert(promote_type(typeof(y), typeof(a), typeof(b)), y)
+    else
+        return zero(y)
```
devmotion, line 79: This branch could still lead to a type instability. Instead of the `promote_type` etc. in the second branch it might be easier to just add a
```julia
y, a, b = promote(y, a, b)
```
at the top of the function body (outside of `if lowerbounded && ...`).

penelopeysm:
We dropped 1.6 already, and the remaining tests pass (known Enzyme crashes notwithstanding).

I'm wondering how we can implement a good test for this. We can't do this for example 

```julia
d = Uniform(-1, 1); b = bijector(d); y = 80; x = inverse(b)(y)

@test logpdf(d, inverse(b)(y)) + logabsdetjacinv(b, y) ≈ logpdf_with_trans(d, x, true)
```

Thanks to this PR the LHS evaluates to a finite value, but the comparison fails because because `x` evaluates to `1.0` and `logpdf_with_trans` returns infinite. (In theory, `x` should be 0.9999999....... but floats aren't precise enough to capture that.)

Maybe we just check isfinite() for the LHS terms?

devmotion:
Probably easiest to compare with `BigFloat`s. E.g., on the master branch the example above gives
```julia
julia> using Bijectors

julia> d = Uniform(big(-1.0), big(1.0));

julia> b = bijector(d);

julia> y = big(80.0);

julia> x = inverse(b)(y)
0.9999999999999999999999999999999999639029722430916965537574328529994522280756386

julia> logpdf_with_trans(d, x, true)
-80.000000000000000000000000000000000036097027594005675893754704116470597853965
```

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.15"
+version = "0.16"
```
devmotion, line 3: That's just a bugfix, isn't it?
```suggestion
version = "0.15.3"
```

penelopeysm: Hmm, pretty sure `truncated_inv_logabsdetjac` is new here

devmotion: Sure, but that's part of the fix for the numerical issues of the (already existing) inverse of the `TruncatedBijector`. So IMO it's a bugfix.

Even if you'd consider it a new feature (which I'd disagree with since it's not a new bijector) then according to (Julia) semver it should be released as 0.15.3 since it's a non-breaking change. 0.16 would be a breaking release.

devmotion (approved):

PR #337, opened 2024-10-28
Remove BijectorsEnzymeExt on 1.11.1

penelopeysm:
<details> <summary> Description of old PR, no longer relevant </summary>

See https://github.com/TuringLang/Bijectors.jl/pull/333 and [the comment added to BijectorsEnzymeExt](https://github.com/TuringLang/Bijectors.jl/pull/337/files#diff-1fc98439919307f9f1c1b212e22b402ea10bd34635e4389eb56173aa82448d6a) for context.

I can confirm that this _doesn't error_ in all contexts (but I can't confirm that it _has the desired effect_, because we can't actually meaningfully test Enzyme on 1.11):

 - Precompilation: doesn't crash
 - Runtime `using Enzyme; using Bijectors`: doesn't crash
 - Runtime `using Bijectors; using Enzyme`: doesn't crash

I tested this locally on Julia 1.11.1 and Enzyme@0.12.36. I've also tested on Julia 1.10.6.

In theory, the main downside of this is:

 - The inlined code is specific for a particular version of Enzyme, in this case, v0.12.36. This is maybe not a big deal if we don't expect any changes in [EnzymeChainRulesCoreExt](https://github.com/EnzymeAD/Enzyme.jl/blob/v0.12.36/ext/EnzymeChainRulesCoreExt.jl). I notice that between that version and now there has only been one change to this file, which is to add a `config` argument. https://github.com/EnzymeAD/Enzyme.jl/commit/ffcb7dd977fada76efa88c10d80f3b47d7bdc9a2#diff-ca235e21f5e92c1ccdcd841ca2665d4e7d9222f3e29e6e0fce9f601f74745175

Overall, though, I think this is a better fix than #333, because it _does_ address the root problem of us implicitly depending on the load order of the two extensions. It just introduces other problems (of maintainability) in the process, but hopefully the comment is clear enough and mitigates this.

</details>

Closes #332 by simply removing the code on 1.11.1+.

PR #338, opened 2024-10-29
Tapir -> Mooncake

penelopeysm:
See https://github.com/TuringLang/Bijectors.jl/pull/337#issuecomment-2443125795 for context.


mhauru (approved):
LGTM

willtebbutt:
Re the Mooncake failure -- I knew about this possibility previously, but this is the first time I've seen it in the wild. Sadly, I'm not 100% sure where to start on fixing this. See https://github.com/compintell/Mooncake.jl/issues/319 for more context.

penelopeysm:
I'll mark the Mooncake test as broken and then we can go from there. Thanks both!

test/ad/utils.jl
```diff
@@ -78,27 +78,39 @@ function test_ad(f, x, broken=(); rtol=1e-6, atol=1e-6)
...
-                )
-            )
+    if (AD == "All" || AD == "Mooncake") && VERSION >= v"1.10"
+        try
+            Mooncake.build_rrule(f, x)
+        catch exc
+            # TODO(penelopeysm):
+            # @test_throws AssertionError (expr...) doesn't work, unclear why
+            @test exc isa AssertionError
         end
```
penelopeysm, line 88: I don't know why `@test_throws` doesn't work here, it's completely inexplicable.

However, I also didn't want to spend another 2 hours debugging this, so I'm just reverting to catching the expression and testing it.

penelopeysm: BTW, we could simply exclude Mooncake from testing entirely, but I prefer having this as it will let us know when the bug above is fixed.

penelopeysm:
Remaining failing tests are due to Enzyme, I'm going to go ahead and merge first.

PR #342, opened 2024-11-01
Bump Julia compat entry to 1.10

penelopeysm:
Mooncake, and Enzyme 0.13, and Turing, only support Julia 1.10+ so I figure we may as well drop support for anything before it.

I tidied up the GHA workflows a bit too.

### DIFF
--- a/.github/workflows/AD.yml
+++ b/.github/workflows/AD.yml
@@ -28,8 +28,8 @@ jobs:
           - ReverseDiff
           - Zygote
     steps:
-      - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@v1
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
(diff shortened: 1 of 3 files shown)


torfjelde:
Happy with this! But can you bump the minor version since this will be breaking?

penelopeysm:
@torfjelde I'm sorry I missed this!

Do you explain how bumping Julia compatibility is breaking? The way I see it, for anybody on >=1.10, nothing changed. For anybody on <1.10, the new version effectively doesn't exist, so there is no breakage.

torfjelde:
That's a fair argument 👍 I think it's a bit of a gray area: I agree that it won't actually break anyone's code, however it arguably marks a point where we stop supporting pre-1.10 versions. I think one good argument for doing it is that it'll be easier to do backports of bugfixes if we end up feeling the need for that: we can just release new versions on the previous minor version.

Sensible or nah?

penelopeysm:
True that's a good argument:) Semver also doesn't mandate that a non-breaking release _has_ to be a patch, so I'm also quite happy to just consider bumping Julia compat to be a minor release based on convention. Thanks!

Issue #339, opened 2024-10-29
Re-introducing Enzyme support on Julia 1.11


penelopeysm:
This issue is intended to consolidate all the info about BijectorsEnzymeExt.

In the text that follows, to save the mental energy needed to parse long strings, `B` is `Bijectors`, `E` is `Enzyme`, and `C` is `ChainRulesCore`.

## What happened with Bijectors v0.13.18?

[BEExt uses the `@import_frule` and `@import_rrule` macros from Enzyme](https://github.com/TuringLang/Bijectors.jl/blob/v0.13.18/ext/BijectorsEnzymeExt.jl). While _these macros_ are defined in Enzyme itself, under the hood they call the `Enzyme._import_frule` and `Enzyme._import_rrule` functions, [which are defined inside `ECExt`](https://github.com/EnzymeAD/Enzyme.jl/blob/v0.12.36/ext/EnzymeChainRulesCoreExt.jl).

Thus, if `BEExt` is loaded before `ECExt`, then a `MethodError` will be raised because the appropriate method has not yet been defined.

Up until Julia 1.11.0, there has been no need to worry about this, because Julia somehow always ensured that `ECExt` was loaded before `BEExt`. There is no _logical reason_ for why one should be loaded before the other, but this behaviour has always been consistent. It is still true on Julia 1.10.6, as well.

However, Julia 1.11.1 (and specifically [this PR](https://github.com/JuliaLang/julia/pull/55589)) changed this and now, there's no guarantee that `ECExt` will be loaded first. **Indeed, it turns out that in Julia 1.11.1, `BEExt` is always loaded first, especially during precompilation.** This meant that environments containing `B` and `E` would fail to precompile, as described in #332.

Several (partial) solutions to this problem were explored.

## Solution 1: Disabling precompilation on 1.11.1+

See #333.

This workaround was suggested in [this comment on the corresponding Julia issue](https://github.com/JuliaLang/julia/issues/56204#issuecomment-2419553167). The result of this was that precompilation would succeed, and `using E; using B` at runtime would also succeed, but `using B; using E` at runtime would fail with the same error.

## Solution 2: Inlining the rule on 1.11.1+

See #337, specifically, [this commit](https://github.com/TuringLang/Bijectors.jl/pull/337/commits/6952360633936194f5dbe162bd1930178346f461).

What this does is to effectively remove the dependency on `ECExt`, by:

1. Inlining the output of

   ```julia
[...]

PR #346, opened 2024-11-06
Re-add expansion of `Enzyme.@import_{r,f}rule` to BijectorsEnzymeExt

penelopeysm:
Since Enzyme now works, but Julia hasn't been fixed, this re-enables the extension. See https://github.com/TuringLang/Bijectors.jl/issues/339#issuecomment-2460682359.

In the process, it aso re-enables tests that are no longer broken.


devmotion (commented):
IMO that's incomprehensible. What's the reason for not just implementing the forward and reverse-mode Enzyme rules? The function and the partial derivatives are very simple, and a direct implementation is likely faster and avoids unnecessary overhead. As also mentioned in the docstrings of `@import_rrule` and `@import_frule`, these macros most likely do not give you the most performant implementation (and they might lead to broken rules but I assume it's fine in this simple case here).

devmotion:
> it's only really intended as a stopgap measure until 1.11.2 is released, at which point we can get rid of all of this.

My point is that even when the Pkg issues are fixed, I think ideally you would use `@import_frule` neither `@import_rrule`.

penelopeysm:
I will pass that on :)

penelopeysm:
But maybe @willtebbutt or @mhauru one of you are better placed to do this (at some point in time)?

penelopeysm:
Is it possible to merge this first and keep track of the proper rule-writing in another issue?

Right now we have a missing rule and that will make things break.

devmotion:
> @devmotion, since you wrote the ChainRulesCore [rule](https://github.com/JuliaMath/Roots.jl/blob/master/ext/RootsChainRulesCoreExt.jl) for Roots.jl, would you be happy to do the same for Enzyme, or coach @mhauru to do it?

Well, one could do that. But that's again too generic for Bijectors IMO. The only interest here is one specific function, so only a rule for this specific function is needed. I assume generally performance of such a specific rule will be better than performance of generic Roots.jl rules.

I should have some time this evening, I'll open a draft PR and see how far I can get.

PR #351, opened 2024-11-19
Add docs for `link`, `invlink`, and `logpdf_with_trans`

penelopeysm:
The docs currently write (https://turinglang.org/Bijectors.jl/stable/):

> The 3 main functions implemented in this package are the link, invlink and logpdf_with_trans for a number of distributions.

However, none of these functions are documented. This PR fixes that.

### DIFF
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -79,10 +79,6 @@ export TransformDistribution,
     InvertibleBatchNorm,
     elementwise
 
-if VERSION < v"1.1"
-    using Compat: eachcol
-end
-
 if VERSION < v"1.9"
     using Compat: stack
 end
--- a/src/Bijectors.jl
+++ b/src/Bijectors.jl
@@ -126,7 +126,59 @@ end
 
 # Distributions
 
+"""
+    link(d::Distribution, x)
+
+Transforms the input `x` using the constrained-to-unconstrained bijector for
+distribution `d`.
+
+See also: [`invlink`](@ref).
+
+# Example
+
+```jldoctest
+julia> using Bijectors
+
+julia> d = LogNormal()   # support is (0, Inf)
+LogNormal{Float64}(μ=0.0, σ=1.0)
+
+julia> b = bijector(d)   # log function transforms to unconstrained space
+(::Base.Fix1{typeof(broadcast), typeof(log)}) (generic function with 1 method)
+
+julia> b(1.0)
+0.0
+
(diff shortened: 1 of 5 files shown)


src/Bijectors.jl
```diff
@@ -160,6 +208,35 @@ end
...
 _logabsdetjac_dist(d::LKJCholesky, x::AbstractVector) = logabsdetjac.((bijector(d),), x)
 
+"""
+    logpdf_with_trans(d::Distribution, x, transform::Bool)
+
+If `transform` is `false`, `logpdf_with_trans` calculates the log probability
+density function (logpdf) of distribution `d` at `x`.
+
+If `transform` is `true`, the logpdf is calculated after transforming `x` using
+the constrained-to-unconstrained bijector for distribution `d`. Specifically,
```
mhauru, line 218: I find the first sentence a bit difficult to understand correctly. One could be lead to think of something like `logpdf(d, link(d, x))`. Or if one doesn't think of that, then one might wonder why it makes a difference whether you transform or not. Is there some alternative phrasing for it? Maybe something about not accounting for the Jacobian term?

penelopeysm: I had an original draft that wrote 'transforming both `x` and `d` using the constrained-to-unconstrained bijector', then I took it out because I wasn't sure if 'transforming a distribution' was clear. Would that be better?

It is true that `logpdf_with_trans(d, x, true) = logpdf(transform(d), link(d, x))`.

mhauru: I think that's much clearer. Would you also be up for adding a sentence saying this is essentially just ignoring the Jacobian term? I think for anyone with background in this sort of stuff that could be illuminating.

penelopeysm: `transform=true` does use the Jacobian, whereas `transform=false` doesn't (because it doesn't change variables):

```julia
julia> using Bijectors

julia> logpdf_with_trans(LogNormal(), ℯ, true)
-1.4189385332046727

julia> logpdf_with_trans(LogNormal(), ℯ, false) - logabsdetjac(bijector(LogNormal()), ℯ)
-1.4189385332046727
```

I can definitely add a note to this effect, though!

penelopeysm: Should be better now :)

mhauru (approved):
Thanks Penny!

sunxd3 (approved):
LGTM, thanks!

PR #353, opened 2024-11-28
Bump minor version to 0.15

mhauru:
Raises minimum Julia version to v1.10 and bumps EnzymeCore compat to latest, with the manual `find_alpha` rule.

~Could also bump the minor, rather than patch, version, because of the changing Julia compat. This shouldn't cause breakage for anyone who manages to install it, so in that sense it's not breaking, but bumping Julia versions could still be considered a "major" thing. Do we have a convention for this in Julia-land or Turing-land?~ EDIT: See later comments, decided to go with a minor

### DIFF
--- a/.github/workflows/AD.yml
+++ b/.github/workflows/AD.yml
@@ -41,7 +41,7 @@ jobs:
       - uses: julia-actions/julia-buildpkg@v1
 
       - uses: julia-actions/julia-runtest@v1
-        continue-on-error: ${{ matrix.AD == 'Enzyme' }}
+        continue-on-error: ${{ matrix.AD == 'Enzyme' && matrix.version == '1' }}
         env:
           GROUP: AD
           AD: ${{ matrix.AD }}
--- a/.github/workflows/AD.yml
+++ b/.github/workflows/AD.yml
@@ -15,12 +15,13 @@ concurrency:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.AD == 'Enzyme' }}
+
     strategy:
       fail-fast: false
       matrix:
         version:
           - 'min'
-          - 'lts'
           - '1'
         os:
           - ubuntu-latest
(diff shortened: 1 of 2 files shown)


penelopeysm (approved):
Compat was already bumped in 0.14.0 so bumping patch on this is fine

penelopeysm:
@yebai The new rule fixes the Interface tests, so the only failing ones are AD/Enzyme. There isn't any requirement that any CI test passes for PRs (https://github.com/TuringLang/Bijectors.jl/settings/branches), so we can just ignore the AD/Enzyme ones when making PRs?

We could hide the red cross (e.g. by not running the tests, or by using `continue-on-error: true`) but I would rather not do this as it's easy to forget to undo

penelopeysm:
That's a fair way of guarding against errors-on-import spoiling the non-Enzyme jobs, but it doesn't help us with errors-on-use. As long as Enzyme is tested in some CI job, there will be a red cross, so doing this won't affect the CI status of new PRs unless we either disable testing or explicitly hide the red cross with `continue-on-error`.

penelopeysm:
This should give us back the green tick. It doesn't fix the possibility of error-on-import, but that's not currently a problem (as the extension stuff was fixed) so that can be for another time imo

mhauru:
> Compat was already bumped in 0.14.0 so bumping patch on this is fine

1.4.2 seems to still say v1.6: https://github.com/TuringLang/Bijectors.jl/blob/v0.14.2/Project.toml#L73

Thanks for the help here @penelopeysm! @yebai, are you happy to leave guarding against the error-on-import stuff for later?

penelopeysm:
> still say v1.6: https://github.com/TuringLang/Bijectors.jl/blob/v0.14.2/Project.toml#L73

Wut, I guess this got lost somewhere in the bunch of PRs.

Anyway we decided that a minor bump is more appropriate for Julia compat bumps:

https://github.com/TuringLang/Bijectors.jl/pull/342

So maybe this should be 0.15.0. I don't know what happened to that PR and why it got lost.

mhauru:
Now bumping to 0.15.

PR #355, opened 2024-11-30
Fix infinite recursion in CorrBijector/VecCorrBijector

### LINKED ISSUE #320: StackOverflow on calling inverse VecCorrBijector with a multidimensional array
```
using Bijectors: VecCorrBijector, inverse
inverse(VecCorrBijector())([[1.0]])
```
causes a stack overflow. It may be a non-sense call, but it should error in a more helpful, manageable way.

penelopeysm:
Closes #320 

With this PR:

```julia
julia> using Bijectors: CorrBijector, VecCorrBijector, inverse

julia> v = [1.0]; m = [[1.0]]
1-element Vector{Vector{Float64}}:
 [1.0]

julia> CorrBijector()(v)
ERROR: MethodError: no method matching cholesky_upper(::Vector{Float64})

julia> inverse(CorrBijector())(m)
ERROR: MethodError: no method matching AbstractFloat(::Type{Vector{Float64}})

julia> inverse(VecCorrBijector())(m)
ERROR: MethodError: no method matching AbstractFloat(::Type{Vector{Float64}})
```

Previously, all of these would fail with StackOverflowError due to infinite recursion between 

https://github.com/TuringLang/Bijectors.jl/blob/55501ae8e6fbb414b743fdf10be34f3275616464/src/interface.jl#L92-L93

and

https://github.com/TuringLang/Bijectors.jl/blob/55501ae8e6fbb414b743fdf10be34f3275616464/src/interface.jl#L213-L216

or in the case of CorrBijector

https://github.com/TuringLang/Bijectors.jl/blob/55501ae8e6fbb414b743fdf10be34f3275616464/src/bijectors/corr.jl#L66

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.15.1"
+version = "0.15.2"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
--- a/src/bijectors/corr.jl
+++ b/src/bijectors/corr.jl
@@ -65,13 +65,13 @@ struct CorrBijector <: Bijector end
 
 with_logabsdet_jacobian(b::CorrBijector, x) = transform(b, x), logabsdetjac(b, x)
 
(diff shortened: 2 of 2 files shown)


torfjelde (approved):
Lovely stuff @penelopeysm :)

Issue #279, opened 2023-07-26
Domain Error for VecCholeskyBijector bijector when calling logabsdetjac


paschermayr:
Hi there, 

Thanks for all the hard work and for updating all the transforms! 

I noticed I get a domain error when evaluating 'logabsdetjac' after mapping the parameter in the unconstrained dimension back via the inverse bijector of the LKJ Cholesky transform (MWE below). 

Would it be possible to return a NaN or -Inf instead of throwing that error? I use this bijector in a larger code base and this sometimes stops a loop, I assume this would be similar during Turing MCMC iterations?

```
using Bijectors		#v0.10.35
using Distributions #v0.25.98
using ForwardDiff 	#v0.10.35
using ReverseDiff 	#v1.15.0

θ_unconstrained = [
	-1.9887091960524537,
	-13.499454444466279,
	-0.39328331954134665,
	-4.426097270849902,
	13.101175413857023,
	7.66647404712346,
	9.249285786544894,
	4.714877413573335,
	6.233118490809442,
	22.28264809311481
]
n = 5
nparam = binomial(n, 2)
d = LKJCholesky(n, 10)
b = Bijectors.bijector(d)
b_inv = inverse(b)

θ = b_inv(θ_unconstrained)
Bijectors.logabsdetjac(b, θ) #ERROR: DomainError with -1.2425014227268605e-5:

```

torfjelde:
> Would it be possible to return a NaN or -Inf instead of throwing that error? 

I think this is a bit too drastic to make _default_ behavior, but I wouldn't be opposed to supporting some sort of "mode" for this though, though uncertain exactly where to put.

> I use this bijector in a larger code base and this sometimes stops a loop, I assume this would be similar during Turing MCMC iterations?

If you use the bijector directly, as in the above snippet, then it should be fairly easy to just check if you hit a `DomainError` yourself, and return `-Inf` in that scenario, no? E.g.

```julia
try
    res = Bijectors.logabsdetjac(b, θ)
catch e
    if e isa DomainError
        res = -Inf
    else
        rethrow(e)
    end
end
```

If you're running into this in a Turing model, then we probably need to implement a evaluation-mode or something as I mentioned above to deal with this.

paschermayr:
> If you use the bijector directly, as in the above snippet, then it should be fairly easy to just check if you hit a DomainError yourself, and return -Inf in that scenario, no? E.g.

[...]

PR #356, opened 2024-11-30
Attempt to improve Cholesky numerical stability

penelopeysm:
The following replaces `log1p(-z^2) / 2` where `z = tanh(y[idx])` with `log(2 / (exp(y[idx]) + exp(-y[idx])))` (which is the same mathematical expression) in `_inv_link_chol_lkj`:

$$\begin{align}
z &= \frac{e^y - e^{-y}}{e^y + e^{-y}} \\
\frac{1}{2}\log(1 - z^2) &= \frac{1}{2}\log\left[1 - \left(\frac{e^y - e^{-y}}{e^y + e^{-y}}\right)^2\right] \\
&= \frac{1}{2}\log\left[1 - \left(\frac{e^{2y} - 2 + e^{-2y}}{e^{2y} + 2 + e^{-2y}}\right)\right] \\
&= \frac{1}{2}\log\left[\frac{4}{e^{2y} + 2 + e^{-2y}}\right] \\
&= \frac{1}{2}\log\left[\left(\frac{2}{e^{y} + e^{-y}}\right)^2\right] \\
&= \log\left[\frac{2}{e^{y} + e^{-y}}\right] \\
\end{align}$$

## Accuracy 1

First, to make sure there aren't any regressions, we'll:

1. Sample from a Cholesky distribution
2. Transform it with the existing bijector
3. Un-transform it with the function that is changed
4. Calculate the max absolute error introduced by the roundtrip transformation
5. Plot the error with the new implementation vs the error with the old implementation

<details><summary>Code to generate plot</summary>

```julia
using Bijectors
using LinearAlgebra
using Distributions
using Random
using Plots

_triu1_dim_from_length(d) = (1 + isqrt(1 + 8d)) ÷ 2

# This is the implementation on master
function _inv_link_chol_lkj_old(y::AbstractVector)
    LinearAlgebra.require_one_based_indexing(y)
    W = similar(y, K, K)
    T = float(eltype(W))
    logJ = zero(T)
    idx = 1
    @inbounds for j in 1:K
        log_remainder = zero(T)  # log of proportion of unit vector remaining
        for i in 1:(j - 1)
            z = tanh(y[idx])
            W[i, j] = z * exp(log_remainder)
            log_remainder += log1p(-z^2) / 2
            logJ += log_remainder
            idx += 1
        end
        logJ += log_remainder
        W[j, j] = exp(log_remainder)
        for i in (j + 1):K
            W[i, j] = 0
        end
    end
    return W, logJ
end

# This is this PR
function _inv_link_chol_lkj_new(y::AbstractVector)
    LinearAlgebra.require_one_based_indexing(y)
    W = similar(y, K, K)
    T = float(eltype(W))
    logJ = zero(T)
    idx = 1
    @inbounds for j in 1:K
        log_remainder = zero(T)  # log of proportion of unit vector remaining
        for i in 1:(j - 1)
            z = tanh(y[idx])
            W[i, j] = z * exp(log_remainder)
            # log_remainder += log(2 * exp(y[idx]) / (1 + exp(2 * y[idx])))
            log_remainder += log(2 / (exp(-y[idx]) + exp(y[idx])))
            logJ += log_remainder
            idx += 1
        end
        logJ += log_remainder
        W[j, j] = exp(log_remainder)
        for i in (j + 1):K
            W[i, j] = 0
        end
    end
    return W, logJ
end

function test_inverse_bijector(f_old, f_new)
    dist = LKJCholesky(5, 1.0, 'U')
    b = Bijectors.VecCorrBijector()
    Random.seed!(468)
    samples = map(1:500) do _
        x = rand(dist)
        y = b(x)
        x_true = Matrix{Float64}(x.U) # Convert to full matrix
        x_old = f_old(y)[1]
        x_new = f_new(y)[1]
        # Return the maximum absolute error between the original sample
        # and sample after roundtrip transformation
        (maximum(abs.(x_true - x_old)), maximum(abs.(x_true - x_new)))
    end
    log_mae_old = log10.([x[1] for x in samples])
    log_mae_new = log10.([x[2] for x in samples])
    scatter(log_mae_old, log_mae_new, label="")
    lim_min = floor(min(minimum(log_mae_old), minimum(log_mae_new)))
    lim_max = ceil(max(maximum(log_mae_old), maximum(log_mae_new)))
    plot!(lim_min:lim_max, lim_min:lim_max, label="y=x", color=:black)
    xlabel!("log10(maximum abs error old)")
    ylabel!("log10(maximum abs error new)")
end

test_inverse_bijector(_inv_link_chol_lkj_old, _inv_link_chol_lkj_new)
savefig("bijector_typical.png")
```

</details>

bijector_typical

There isn't really much between the two implementations, sometimes the old one is better, sometimes the new one is better. In any case, the differences are very small so I think the new implementation can be said to at least break even.

## Accuracy 2

However, when sampling in the unconstrained space, there's no guarantee that the transformed value will resemble anything obtained via a forward transformation. This leads to issues like #279.

To test out the numerical performance of invlinking random transformed samples, we can:

1. Generate a random transformed sample.
2. Invlink it with the old method, but using arbitrary precision floats. This is our ground truth.
3. Invlink it with the old method, with Float64 precision
4. Invlink it with the new method, with Float64 precision
5. Compare the errors as above

<details><summary>Code to generate plot</summary>

```julia
function test_inverse_bijector_unconstrained(f_old, f_new)
    dist = LKJCholesky(5, 1.0, 'U')
    Random.seed!(468)
    samples = map(1:500) do _
        y = rand(dist.d * (dist.d - 1) ÷ 2) * 10
        x_true = f_old(Vector{BigFloat}(y))[1]
        x_old = f_old(y)[1]
        x_new = f_new(y)[1]
        # Return the maximum absolute error between the original sample
        # and sample after roundtrip transformation
        (maximum(abs.(x_true - x_old)), maximum(abs.(x_true - x_new)))
    end
    log_mae_old = log10.([x[1] for x in samples])
    log_mae_new = log10.([x[2] for x in samples])
    scatter(log_mae_old, log_mae_new, label="")
    lim_min = floor(min(minimum(log_mae_old), minimum(log_mae_new)))
    lim_max = ceil(max(maximum(log_mae_old), maximum(log_mae_new)))
    plot!(lim_min:lim_max, lim_min:lim_max, label="y=x", color=:black)
    xlabel!("log10(maximum abs error old)")
    ylabel!("log10(maximum abs error new)")
end
test_inverse_bijector_unconstrained(_inv_link_chol_lkj_old, _inv_link_chol_lkj_new)
savefig("bijector_unconstrained.png")
```

</details>

bijector_unconstrained

As can be seen, the new method leads to much smaller errors (consistently around the magnitude of `eps() ~ 1e-16`) whereas the old method often has errors that are several orders of magnitude larger.


src/bijectors/corr.jl
```diff
@@ -348,13 +348,12 @@ function _inv_link_chol_lkj(Y::AbstractMatrix)
...
     logJ = zero(T)
 
-    idx = 1
     @inbounds for j in 1:K
         log_remainder = zero(T)  # log of proportion of unit vector remaining
         for i in 1:(j - 1)
             z = tanh(Y[i, j])
             W[i, j] = z * exp(log_remainder)
-            log_remainder += log1p(-z^2) / 2
+            log_remainder += log(2 / (exp(Y[i, j]) + exp(-Y[i, j])))
```
devmotion, line 356: AFAICT you can just use
```suggestion
            log_remainder -= logcosh(Y[i, j])
```
(assuming LogExpFunctions is loaded)

src/bijectors/corr.jl
```diff
@@ -380,10 +379,10 @@ function _inv_link_chol_lkj(y::AbstractVector)
         log_remainder = zero(T)  # log of proportion of unit vector remaining
         for i in 1:(j - 1)
             z = tanh(y[idx])
-            idx += 1
             W[i, j] = z * exp(log_remainder)
-            log_remainder += log1p(-z^2) / 2
+            log_remainder += log(2 / (exp(y[idx]) + exp(-y[idx])))
```
devmotion, line 383:
```suggestion
            log_remainder -= logcosh(y[idx])
```

src/bijectors/corr.jl
```diff
@@ -497,7 +496,7 @@ function _logabsdetjac_inv_chol(y::AbstractVector)
         tmp = zero(result)
         for _ in 1:(j - 1)
             z = tanh(y[idx])
-            logz = log(1 - z^2)
+            logz = 2 * log(2 / (exp(y[idx]) + exp(-y[idx])))
```
devmotion, line 499:
```suggestion
            logz = -logcosh(y[idx])
```

devmotion (commented):
AFAICT it's even simpler since $$\log(1 - \tanh^2(x))/2 = \log(\mathrm{sech}^2(x)) / 2 = \log(\mathrm{sech}(x)) = \log(1 / \cosh(x)) = -\log(\cosh(x))$$, and `LogExpFunctions.logcosh` is supposed to provide a numerically stable and efficient implementation of $$\log(\cosh(\cdot))$$ (if there are problems they should be considered bugs).

penelopeysm:
👀 I'll give that a spin

penelopeysm:
Looking under the hood logcosh is implemented the same way as above, but the single function call is great 👍

src/bijectors/corr.jl
```diff
@@ -460,13 +460,8 @@ function _logabsdetjac_inv_corr(Y::AbstractMatrix)
...
     result = float(zero(eltype(Y)))
-    for j in 2:K, i in 1:(j - 1)
-        @inbounds abs_y_i_j = abs(Y[i, j])
-        result +=
-            (K - i + 1) * (
-                IrrationalConstants.logtwo -
-                (abs_y_i_j + LogExpFunctions.log1pexp(-2 * abs_y_i_j))
-            )
+    @inbounds for j in 2:K, i in 1:(j - 1)
+        result += (K - i + 1) * (-LogExpFunctions.logcosh(Y[i, j]))
```
devmotion: Slightly simpler maybe
```suggestion
        result -= (K - i + 1) * LogExpFunctions.logcosh(Y[i, j])
```
or
```suggestion
        result += (i - K - 1) * LogExpFunctions.logcosh(Y[i, j])
```

src/bijectors/corr.jl
```diff
@@ -495,8 +490,7 @@ function _logabsdetjac_inv_chol(y::AbstractVector)
     @inbounds for j in 2:K
         tmp = zero(result)
         for _ in 1:(j - 1)
-            z = tanh(y[idx])
-            logz = 2 * log(2 / (exp(y[idx]) + exp(-y[idx])))
+            logz = -2 * LogExpFunctions.logcosh(y[idx])
```
devmotion, line 488: The name `logz` is a bit meaningless now I guess 🙂

src/bijectors/corr.jl
penelopeysm: I changed that one just 10 seconds ago 😄

penelopeysm:
# Forward transforms

Digging into the forward transform bit, the upper triangular matrix `W` which it acts on is supposed to have unit vectors i.e. `sum(W[:, j] .^ 2)` should be 1 for each `j`:

```julia
julia> s = rand(LKJCholesky(5, 1.0, 'U')).U
5×5 UpperTriangular{Float64, Matrix{Float64}}:
 1.0  0.345448  -0.478      0.455158   0.385151
  ⋅   0.938438  -0.331921  -0.305083  -0.0469749
  ⋅    ⋅         0.813231  -0.397178   0.831726
  ⋅    ⋅          ⋅         0.73621    0.0298828
  ⋅    ⋅          ⋅          ⋅         0.395968

julia> [sum(s[:, i] .^ 2) for i in 1:5]
5-element Vector{Float64}:
 1.0
 1.0
 1.0
 1.0
 1.0000000000000002
```

In the forward transform code, `remainder_sq` is initialised at one and then the squares of each element going down the column are successively subtracted, so `remainder_sq` is really a sum of squares of elements not yet seen.

https://github.com/TuringLang/Bijectors.jl/blob/f52a9c52ede1f43155239447601387eb1dafe394/src/bijectors/corr.jl#L321-L331

This implementation, instead of subtracting successive squares, just declares remainder_sq to be that sum:

```julia
    @inbounds for j in 2:K
        y[idx] = atanh(W[1, j])
        idx += 1
        remainder_sq = 1 - W[1, j]^2
        for i in 2:(j - 1)
            remainder_sq = sum(W[i:end, j].^2)
            z = W[i, j] / sqrt(remainder_sq)
            y[idx] = atanh(z)
            idx += 1
        end
end
```

Now, because `W[i, j] ^ 2` is part of that sum, `z` can now no longer be larger than 1, and atanh doesn't throw a DomainError.

## Implementation

<details><summary>Setup code</summary>

```julia
using Bijectors
using LinearAlgebra
using Distributions
using Random
using LogExpFunctions

# Using the invlink definition from this PR
_triu1_dim_from_length(d) = (1 + isqrt(1 + 8d)) ÷ 2
function _inv_link_chol_lkj_new(y::AbstractVector)
    LinearAlgebra.require_one_based_indexing(y)
    W = similar(y, K, K)
    T = float(eltype(W))
    logJ = zero(T)
    idx = 1
    @inbounds for j in 1:K
        log_remainder = zero(T)  # log of proportion of unit vector remaining
        for i in 1:(j - 1)
            z = tanh(y[idx])
            W[i, j] = z * exp(log_remainder)
            log_remainder -= LogExpFunctions.logcosh(y[idx])
            logJ += log_remainder
            idx += 1
        end
        logJ += log_remainder
        W[j, j] = exp(log_remainder)
        for i in (j + 1):K
            W[i, j] = 0
        end
    end
    return W, logJ
end

# Existing link implementation
function _link_chol_lkj_from_upper_old(W::AbstractMatrix)
    K = LinearAlgebra.checksquare(W)
    N = ((K - 1) * K) ÷ 2   # {K \choose 2} free parameters
    y = similar(W, N)
    idx = 1
    @inbounds for j in 2:K
        y[idx] = atanh(W[1, j])
        idx += 1
        remainder_sq = 1 - W[1, j]^2
        for i in 2:(j - 1)
            z = W[i, j] / sqrt(remainder_sq)
            y[idx] = atanh(z)
            remainder_sq -= W[i, j]^2
            idx += 1
        end
    end
    return y
end

# New proposal
function _link_chol_lkj_from_upper_new(W::AbstractMatrix)
    K = LinearAlgebra.checksquare(W)
    N = ((K - 1) * K) ÷ 2   # {K \choose 2} free parameters
    y = similar(W, N)
    idx = 1
    @inbounds for j in 2:K
        y[idx] = atanh(W[1, j])
        idx += 1
        for i in 2:(j - 1)
            remainder = norm(W[i:end, j].^2)
            z = W[i, j] / remainder
            y[idx] = atanh(z)
            idx += 1
        end
    end
    return y
end

function plot_maes(samples)
    log_mae_old = log10.([sample[1] for sample in samples])
    log_mae_new = log10.([sample[2] for sample in samples])
    scatter(log_mae_old, log_mae_new, label="")
    lim_min = floor(min(minimum(log_mae_old), minimum(log_mae_new)))
    lim_max = ceil(max(maximum(log_mae_old), maximum(log_mae_new)))
    plot!(lim_min:lim_max, lim_min:lim_max, label="y=x", color=:black)
    xlabel!("log10(maximum abs error old)")
    ylabel!("log10(maximum abs error new)")
end

function test_forward_bijector(f_old, f_new)
    dist = LKJCholesky(5, 1.0, 'U')
    Random.seed!(468)
    samples = map(1:500) do _
        x = rand(dist)
        x_again_old = _inv_link_chol_lkj_new(f_old(x.U))[1]
        x_again_new = _inv_link_chol_lkj_new(f_new(x.U))[1]
        # Return the maximum absolute error between the original sample
        # and sample after roundtrip transformation
        (maximum(abs.(x.U - x_again_old)), maximum(abs.(x.U - x_again_new)))
    end
    return samples
end
```

</details>

```julia
julia> y = rand(Random.Xoshiro(468), 10) * 16;

julia> x = _inv_link_chol_lkj_new(y)[1];

julia> y_old = _link_chol_lkj_from_upper_old(x)
ERROR: DomainError with 1.000207932997037:
atanh(x) is only defined for |x| ≤ 1.
Stacktrace:
 [1] atanh_domain_error(x::Float64)
   @ Base.Math ./special/hyperbolic.jl:240
 [2] atanh
   @ ./special/hyperbolic.jl:256 [inlined]
 [3] _link_chol_lkj_from_upper_old(W::Matrix{Float64})
   @ Main ./REPL[29]:12
 [4] top-level scope
   @ REPL[78]:1

julia> y_new = _link_chol_lkj_from_upper_new(x)
10-element Vector{Float64}:
  1.7139942709891685
  4.050190371709019
 12.606351374271206
  8.239542965781226
  7.897855159032619
  6.885928358454504
  7.201266901997009
  4.588778566499247
  5.507106236959028
 11.582258189742753
```

So, this implementation will at least solve the DomainErrors. The numerical accuracy on roundtrip transformation is actually marginally better than the existing implementation:

```julia
julia> plot_maes(test_forward_bijector(_link_chol_lkj_from_upper_old, _link_chol_lkj_from_upper_new))
```

bijector_forward_typical

Remaining concerns with this are (1) performance (urk!):

```julia
julia> using Chairmarks

julia> @be (rand(LKJCholesky(5, 1.0, 'U'))) _link_chol_lkj_from_upper_old(_.U)
Benchmark: 2915 samples with 285 evaluations
 min    94.007 ns (2 allocs: 144 bytes)
 median 102.193 ns (2 allocs: 144 bytes)
 mean   111.010 ns (2 allocs: 144 bytes, 0.25% gc time)
 max    8.571 μs (2 allocs: 144 bytes, 97.63% gc time)

julia> @be (rand(LKJCholesky(5, 1.0, 'U'))) _link_chol_lkj_from_upper_new(_.U)
Benchmark: 2632 samples with 90 evaluations
 min    319.444 ns (14 allocs: 672 bytes)
 median 335.189 ns (14 allocs: 672 bytes)
 mean   387.088 ns (14 allocs: 672 bytes, 0.15% gc time)
 max    40.176 μs (14 allocs: 672 bytes, 98.47% gc time)
```

And accuracy on pathological samples:

```julia
julia> y = rand(Random.Xoshiro(468), 10) * 16;

julia> x = _inv_link_chol_lkj_new(y)[1];

julia> y_new = _link_chol_lkj_from_upper_new(x)
10-element Vector{Float64}:
  1.7139942709891685
  4.050190371709019
 12.606351374271206
  8.239542965781226
  7.897855159032619
  6.885928358454504
  7.201266901997009
  4.588778566499247
  5.507106236959028
 11.582258189742753

julia> maximum(abs.(y - y_new))
1.2023473718869582e-6
```

penelopeysm:
Sorry I dropped the ball on this, it probably slipped off my radar around Christmas. @devmotion Would you be willing to review again? I'll try to look into the forward transform PR (#357) separately.

joelkandiah (commented):
This looks good to me but it may be worth someone checking if this same replacement could or should also be used on line 417 in the rrule.

devmotion (commented):
Can you update the initial comment to (more clearly) reflect the current status and benchmarks of the PR? IMO it's a bit difficult to tell whether the comments and plots refer to the current version of the PR or some previous iteration.

joelkandiah (approved):

penelopeysm:
@devmotion I've just updated them! I also added another commit to calculate tanh and logcosh outside of the loop, which means that it's now also a tiny bit faster than before.

src/bijectors/corr.jl
```diff
@@ -375,15 +374,18 @@ function _inv_link_chol_lkj(y::AbstractVector)
     T = float(eltype(W))
     logJ = zero(T)
 
+    z_vec = tanh.(y)
+    lc_vec = LogExpFunctions.logcosh.(y)
```
devmotion, line 378: This doesn't save any floating point operations, does it? Every element seems to be accessed exactly once in the nested loops below. Is it faster due to SIMD optimizations, or is there another reason to move it outside of the loop?

In general, I think you'd want to avoid broadcasting here. The operation is a simple `map` (so not actually broadcasting dimensions etc.) which tends to cause less stress on the compiler than `broadcast` and often leads to more optimized code.
```suggestion
    z_vec = map(tanh, y)
    lc_vec = map(LogExpFunctions.logcosh, y)
```

src/bijectors/corr.jl
```diff
@@ -405,17 +407,18 @@ function _inv_link_chol_lkj_rrule(y::AbstractVector)
     logJ = zero(T)
 
     z_vec = tanh.(y)
+    lc_vec = LogExpFunctions.logcosh.(y)
```
devmotion, line 410: Same here, why move this outside of the loop? Also, I'd suggest
```suggestion
    z_vec = map(tanh, y)
    lc_vec = map(LogExpFunctions.logcosh, y)
```

src/bijectors/corr.jl
penelopeysm: It's the same number of operations, but a bit faster probably due to lower level optimisations (it wasn't immediately obvious to me why). Will change to `map`!

yebai:
Many thanks @penelopeysm, @joelkandiah and @devmotion!

Issue #365, opened 2025-02-19
`Bijectors.with_logabsdet_jacobian` on `Bijectors.inverse(Bijectors.PDVecBijector())` with `ReverseDiff` errors


sunxd3:
MWE

```julia
using Bijectors, ReverseDiff

f(X::AbstractVector{<:Real}) = Bijectors.with_logabsdet_jacobian(
        Bijectors.inverse(Bijectors.PDVecBijector()),
        X,
    )[2]
f(rand(3))
ReverseDiff.gradient(f, rand(3))
```
generates

<details>
  <summary>stacktrace</summary>

```julia
ERROR: MethodError: no method matching iterate(::Nothing)
The function `iterate` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  iterate(::SimpleTraits.GenerateTypeVars{:upcase})
   @ SimpleTraits ~/.julia/packages/SimpleTraits/l1ZsK/src/SimpleTraits.jl:445
  iterate(::SimpleTraits.GenerateTypeVars{:lcase})
   @ SimpleTraits ~/.julia/packages/SimpleTraits/l1ZsK/src/SimpleTraits.jl:446
  iterate(::SimpleTraits.GenerateTypeVars{:upcase}, ::Any)
   @ SimpleTraits ~/.julia/packages/SimpleTraits/l1ZsK/src/SimpleTraits.jl:445
  ...

Stacktrace:
  [1] indexed_iterate(I::Nothing, i::Int64)
    @ Base ./tuple.jl:162
  [2] (::BijectorsReverseDiffExt.var"##transpose_eager#373#163")(X_tracked::ReverseDiff.TrackedArray{…})
    @ BijectorsReverseDiffExt /generic/path/BijectorsReverseDiffExt.jl:264
  [3] track(::typeof(Bijectors.transpose_eager), X_tracked::ReverseDiff.TrackedArray{…})
    @ BijectorsReverseDiffExt /generic/path/macros.jl:195
  [4] transpose_eager(X::ReverseDiff.TrackedArray{Float64, Float64, 2, Matrix{Float64}, Matrix{Float64}})
    @ BijectorsReverseDiffExt /generic/path/BijectorsReverseDiffExt.jl:261
  [5] transform(::Inverse{Bijectors.PDVecBijector}, y::ReverseDiff.TrackedArray{Float64, Float64, 1, Vector{…}, Vector{…}})
    @ Bijectors /generic/path/pd.jl:41
[...]

PR #366, opened 2025-02-20
Add `ChainRules` to `Bijectors` weak dependencies

sunxd3:
Fix https://github.com/TuringLang/Bijectors.jl/issues/365

Bijectors ~indirectly~ directly depends on `ChainRules.jl` on version 0.14, but this ~indirect~ dependency is dropped in version 0.15.
~I am not certain what changed, and don't have a clever way to find this out.~ https://github.com/TuringLang/Bijectors.jl/pull/350

There are some code in `ReverseDiffExt`:
https://github.com/TuringLang/Bijectors.jl/blob/df3695cd5e4650e7fc6c819774f445878d8b4371/ext/BijectorsReverseDiffExt.jl#L228-L229
https://github.com/TuringLang/Bijectors.jl/blob/df3695cd5e4650e7fc6c819774f445878d8b4371/ext/BijectorsReverseDiffExt.jl#L246-L247
https://github.com/TuringLang/Bijectors.jl/blob/df3695cd5e4650e7fc6c819774f445878d8b4371/ext/BijectorsReverseDiffExt.jl#L264

The third is easy to fix (see [David's comment](https://github.com/TuringLang/Bijectors.jl/issues/365#issuecomment-2667914537). The first two, however, don't look so trivial([`cholesky`](https://github.com/JuliaDiff/ChainRules.jl/blob/02e21bacd8cb0aec3e80678ec2d0e1e1a964e9e3/src/rulesets/LinearAlgebra/factorization.jl#L458-L544) [`Hermitian`](https://github.com/JuliaDiff/ChainRules.jl/blob/02e21bacd8cb0aec3e80678ec2d0e1e1a964e9e3/src/rulesets/LinearAlgebra/symmetric.jl#L4C1-L93C1)).

Adding `ChainRules` as a direct dependency of `Bijectors` is an obvious solution, not sure if this is proper.


Project.toml
```diff
@@ -42,6 +43,7 @@ BijectorsZygoteExt = "Zygote"
 
 [compat]
 ArgCheck = "1, 2"
+ChainRules = "1.72.2"
```
sunxd3, line 46: not sure what versions are good here

penelopeysm:
The ChainRules dep was removed as part of https://github.com/TuringLang/Bijectors.jl/pull/350 I guess maybe we didn't realise that we needed it for ReverseDiffExt too.
Personally I'm happy to just re-add it as a dep.

> don't have a clever way to find this out

`git bisect` 😄

Project.toml
penelopeysm: before #350 the version bound was just 

```toml
ChainRules = "1"
```

so we could go back to that.

sunxd3:
> git bisect

yep, yep

I am thinking there might be a Julia package that lays out all the indirect dependencies. `Pkg.dependencies` can do this somewhat.

> re-add it as a dep

`DistributionsAD` has `ChainRules` as a dependency, so the tests env has `ChainRules` in it.

sunxd3:
The test errors look to be not related to the PR.

penelopeysm:
Are you happy to wait for me to fix that error first before merging this? It shouldn't take too long

sunxd3:
not at all, thanks!

Project.toml
```diff
@@ -1,9 +1,10 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.15.4"
+version = "0.15.5"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
+ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
```
yebai, line 7: If we only need `ChainRules` in the `ReverseDiffBijectors` extension, maybe a weak dep is sufficient?

penelopeysm: I thought about that too - the reason why I feel a bit less inclined to do that is because the user still has to import ChainRules manually before the extension code will load, and if they just add Bijectors and ReverseDiff the extension won't trigger - and it's hard for a user to find that out without looking at the Project.toml.

Agree that in principle a weakdep would be better though.

sunxd3: I had the exact thought as Penny's

penelopeysm: I don't mind that — just that we might also need to check if any of the other AD extensions also need ChainRules to work

sunxd3: It's good to double check, I think the only other `rrule` definition I found is https://github.com/TuringLang/Bijectors.jl/blob/ed0f74417c132f455c8ae778b15b5273aa9a6dac/ext/BijectorsMooncakeExt.jl#L26
But it seems that this is not defined in `ChainRules`. @willtebbutt

devmotion: The rule is defined in Bijectors: https://github.com/TuringLang/Bijectors.jl/blob/dda1b497776fd60ca224f8ed09122901de3d4080/src/chainrules.jl#L2-L6 It's so simple that in principle Mooncake could also implement it directly AFAICT (as done for ReverseDiff and Enzyme).

ext/BijectorsReverseDiffExt.jl
```diff
@@ -223,47 +223,24 @@ end
...
+        "Please load ChainRules to use ReverseDiff with this bijector."))
 end
 
 transpose_eager(X::TrackedMatrix) = track(transpose_eager, X)
 @grad function transpose_eager(X_tracked::TrackedMatrix)
     X = value(X_tracked)
-    y, y_pullback = ChainRulesCore.rrule(permutedims, X, (2, 1))
-    transpose_eager_pullback(Δ) = (y_pullback(Δ)[2],)
-    return y, transpose_eager_pullback
+    transpose_eager_pullback(Δ) = (transpose_eager(Δ),)
```
sunxd3, line 250: I used David's suggestion here and removed the `ChainRules` dependency.

ext/BijectorsReverseDiffExt.jl
```diff
@@ -223,47 +223,32 @@ end
...
-
-    return lower_triangular(parent(C.L)), cholesky_lower_pullback
+@grad function cholesky_lower(X::TrackedMatrix)
+    throw(
+        ErrorException(
+            "The gradient function for `cholesky_lower` is not available in BijectorsReverseDiffExt. " *
+            "It is implemented in BijectorsReverseDiffChainRulesExt and requires ChainRules.jl. " *
+            "Please load ChainRules to use ReverseDiff with this bijector.",
+        ),
+    )
```
sunxd3, line 233: I am not sure if it is correct to error in a `grad` macro decorated function.

Project.toml
```diff
@@ -36,12 +37,13 @@ BijectorsEnzymeCoreExt = "EnzymeCore"
 BijectorsForwardDiffExt = "ForwardDiff"
 BijectorsLazyArraysExt = "LazyArrays"
 BijectorsMooncakeExt = "Mooncake"
-BijectorsReverseDiffExt = "ReverseDiff"
+BijectorsReverseDiffChainRulesExt = "ReverseDiff"
```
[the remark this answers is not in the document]
sunxd3: Thanks for the suggestions, I tried to do what you described, let me know where to improve!

ext/BijectorsReverseDiffChainRulesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+
+import Bijectors: lower_triangular, upper_triangular, cholesky_lower, cholesky_upper
+
+using Bijectors.LinearAlgebra
+using Bijectors.Distributions: LocationScale
+
+@grad function cholesky_lower(X_tracked::TrackedMatrix)
+    X = value(X_tracked)
+    H, hermitian_pullback = ChainRulesCore.rrule(Hermitian, X, :L)
+    C, cholesky_pullback = ChainRulesCore.rrule(cholesky, H, Val(false))
```
devmotion, line 15:
```suggestion
    H, hermitian_pullback = ChainRules.rrule(Hermitian, X, :L)
    C, cholesky_pullback = ChainRules.rrule(cholesky, H, Val(false))
```

ext/BijectorsReverseDiffChainRulesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+    function cholesky_lower_pullback(ΔL)
+        ΔC = ChainRulesCore.Tangent{typeof(C)}(; factors=(C.uplo === :L ? ΔL : ΔL'))
```
devmotion, line 17:
```suggestion
        ΔC = ChainRules.Tangent{typeof(C)}(; factors=(C.uplo === :L ? ΔL : ΔL'))
```

ext/BijectorsReverseDiffChainRulesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+        # part zeroed out).
+        return (Δx,)
+    end
+
+    return lower_triangular(parent(C.L)), cholesky_lower_pullback
+end
+
+@grad function cholesky_upper(X_tracked::TrackedMatrix)
+    X = value(X_tracked)
+    H, hermitian_pullback = ChainRulesCore.rrule(Hermitian, X, :U)
```
devmotion, line 31:
```suggestion
    H, hermitian_pullback = ChainRules.rrule(Hermitian, X, :U)
```

ext/BijectorsReverseDiffChainRulesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+    C, cholesky_pullback = ChainRulesCore.rrule(cholesky, H, Val(false))
```
devmotion, line 32:
```suggestion
    C, cholesky_pullback = ChainRules.rrule(cholesky, H, Val(false))
```

ext/BijectorsReverseDiffChainRulesExt.jl
```diff
@@ -0,0 +1,46 @@
...
+    function cholesky_upper_pullback(ΔU)
+        ΔC = ChainRulesCore.Tangent{typeof(C)}(; factors=(C.uplo === :U ? ΔU : ΔU'))
```
devmotion, line 34:
```suggestion
        ΔC = ChainRules.Tangent{typeof(C)}(; factors=(C.uplo === :U ? ΔU : ΔU'))
```

ext/BijectorsReverseDiffExt.jl
devmotion: No, this is not correct since you would try to overwrite `track(cholesky_lower, X::TrackedMatrix)` in the other extension which leads to warnings: https://github.com/TuringLang/Bijectors.jl/actions/runs/13481480358/job/37667244593?pr=366#step:5:473

I would suggest adding an error hint in the vanilla ReverseDiff extension: https://docs.julialang.org/en/v1/base/base/#Base.Experimental.register_error_hint

That is, I'd add something like

```julia
if isdefined(Base.Experimental, :register_error_hint)
    function __init__()
        Base.Experimental.register_error_hint(MethodError) do io, exc, argtypes, kwargs
            if exc.f === ReverseDiff.track && length(argtypes) == 2 && (argtypes[1] === typeof(cholesky_lower) || argtypes[1] === typeof(cholesky_upper))
                 print(io, lazy"\nThe gradient function for ")
                 if argtypes[1] === typeof(cholesky_lower)
                     print(io, "`cholesky_lower`")
                 else
                     print(io, "`cholesky_upper`")
                 end
                 print(io, " requires ChainRules.jl. ")
                 print(io, "Please load ChainRules to use ReverseDiff with this bijector.")
            end
        end
    end
end
```

to this file here.

sunxd3:
Curiously, the error seems to go away by simply removing the definition of gradient functions for `cholesky_upper` and `cholesky_lower` (without loading `BijectorsReverseDiffChainRulesExt`). I'll investigate a bit later.

ext/BijectorsReverseDiffExt.jl
```diff
@@ -222,48 +222,33 @@ end
 @grad_from_chainrules _link_chol_lkj_from_upper(x::TrackedMatrix)
 @grad_from_chainrules _link_chol_lkj_from_lower(x::TrackedMatrix)
 
-cholesky_lower(X::TrackedMatrix) = track(cholesky_lower, X)
```
devmotion, line 225: You have to keep this line in this file, to ensure that `TrackedMatrix` does not hit the `AbstractMatrix` fallback and that the error hint works as intended.

ext/BijectorsReverseDiffExt.jl
```diff
@@ -222,48 +222,33 @@ end
...
+                print(io, " requires ChainRules.jl. ")
+                print(io, "Please load ChainRules to use ReverseDiff with this bijector.")
+            end
+        end
     end
-
-    return lower_triangular(parent(C.L)), cholesky_lower_pullback
-end
-
-cholesky_upper(X::TrackedMatrix) = track(cholesky_upper, X)
```
devmotion, line 243: Same here.

test/Project.toml
```diff
@@ -1,6 +1,7 @@
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 AdvancedHMC = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
+ChainRules = "082447d4-558c-5d27-93f4-14fc19e9eca2"
```
devmotion, line 4: Since ChainRules is a dependency of Zygote (and possibly other AD backends) it will always be loaded anyway. So it's not necessary to add it explicitly here, but it might be "less magical".

sunxd3:
Aha, that makes sense. I moved the two methods back.

I made a slight adjustment where I removed the use of `lazy_str` because it leads to  error message like
```julia
julia> ReverseDiff.gradient(f, rand(3))
ERROR: MethodError: no method matching track(::typeof(Bijectors.cholesky_lower), ::ReverseDiff.TrackedArray{Float64, Float64, 2, Matrix{Float64}, Matrix{Float64}})
The function `track` exists, but no method is defined for this combination of argument types.\nThe gradient function for `cholesky_lower` requires ChainRules.jl. Please load ChainRules to use ReverseDiff with this bijector.

[...]
```

instead of 

```julia
ERROR: MethodError: no method matching track(::typeof(Bijectors.cholesky_lower), ::ReverseDiff.TrackedArray{Float64, Float64, 2, Matrix{Float64}, Matrix{Float64}})
The function `track` exists, but no method is defined for this combination of argument types.
The gradient function for `cholesky_lower` requires ChainRules.jl. Please load ChainRules to use ReverseDiff with this bijector.
[...]
```

sunxd3:
The two failed tests are with Enzyme and look unrelated.

yebai (approved):

sunxd3:
I am not authorized to merge. So feel free to merge if some of you see fit!

Issue #367, opened 2025-02-21
1.10.0 CI failing


penelopeysm:
And it's a nasty one, a compiler error:

```
Internal error: encountered unexpected error in runtime:
MethodError(f=Core.Compiler.widenconst, args=(Base.ComposedFunction{typeof(Base.identity), Base.Fix1{Type{Base.MappingRF{F, T} where T where F}, typeof(LinearAlgebra.norm)}}(outer=typeof(Base.identity)(), inner=Base.Fix1{Type{Base.MappingRF{F, T} where T where F}, typeof(LinearAlgebra.norm)}(f=Base.MappingRF{F, T} where T where F, x=typeof(LinearAlgebra.norm)())),), world=0x00000000000015a8)
jl_method_error_bare at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:2208
jl_method_error at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:2226
jl_lookup_generic_ at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:3057 [inlined]
ijl_apply_generic at /cache/build/builder-amdci4-6/julialang/julia-release-1-dot-10/src/gf.c:3072
sroa_pass! at ./compiler/ssair/passes.jl:1048
run_passes at ./compiler/optimize.jl:489
run_passes at ./compiler/optimize.jl:504 [inlined]
optimize at ./compiler/optimize.jl:453 [inlined]
_typeinf at ./compiler/typeinfer.jl:272
typeinf at ./compiler/typeinfer.jl:216
typeinf_edge at ./compiler/typeinfer.jl:930
abstract_call_method at ./compiler/abstractinterpretation.jl:629
abstract_call_gf_by_type at ./compiler/abstractinterpretation.jl:95
abstract_call_known at ./compiler/abstractinterpretation.jl:2083
abstract_call at ./compiler/abstractinterpretation.jl:2165
abstract_call at ./compiler/abstractinterpretation.jl:2158
abstract_call at ./compiler/abstractinterpretation.jl:2350
abstract_eval_call at ./compiler/abstractinterpretation.jl:2366
abstract_eval_statement_expr at ./compiler/abstractinterpretation.jl:2376
abstract_eval_statement at ./compiler/abstractinterpretation.jl:2620
abstract_eval_basic_statement at ./compiler/abstractinterpretation.jl:2909
[...]

PR #368, opened 2025-02-21
Bump Julia compat to 1.10.8

penelopeysm:
Attempt at fixing https://github.com/TuringLang/Bijectors.jl/issues/367


penelopeysm:
Lovely, 1.10.8 works, so now we just have to bisect to find which patch version to stick in our compat file

penelopeysm:
1.10.4 fails...

penelopeysm:
1.10.6 fails...

penelopeysm:
Will continue bisecting tomorrow

penelopeysm:
OK. It's caused by Mooncake, but not _just_ Mooncake, because DynamicPPL tests still pass with new Mooncake + 1.10.0. So it's really some interaction between Bijectors and Mooncake that causes this failure.

In this case, I think the right move is to just bump the Julia compat entry in Bijectors to 1.10.8.

penelopeysm:
It's Mooncake 0.4.82 that broke it https://github.com/compintell/Mooncake.jl/pull/460

sunxd3:
I remember seeing [this PR](https://github.com/compintell/Mooncake.jl/pull/429). (Zero certain it's related)

penelopeysm:
Quite possibly!

test/ad/utils.jl
```diff
@@ -87,44 +87,26 @@ function test_ad(f, x, broken=(); rtol=1e-6, atol=1e-6)
...
+                    rtol=rtol,
+                    atol=atol,
+                )
+            )
         end
-        # TODO: The above @test_throws happens because of 
-        # https://github.com/compintell/Mooncake.jl/issues/319. If that test
-        # fails, it probably means that the issue was fixed, in which case
-        # we can remove that block and uncomment the following instead.
-
```
penelopeysm, line 107: this was fixed in recent Mooncake versions

Project.toml
```diff
@@ -55,14 +55,14 @@ IrrationalConstants = "0.1, 0.2"
...
-Mooncake = "0.4.19"
+Mooncake = "0.4.95"
 Reexport = "0.2, 1"
 ReverseDiff = "1"
 Roots = "1.3.15, 2"
 Statistics = "1"
 Tracker = "0.2"
 Zygote = "0.6.63, 0.7"
-julia = "1.10"
+julia = "1.10.8"
```
penelopeysm, line 65: I thought about just skipping the failing MooncakeInterpreter tests and leaving the compat entry alone, but I don't really feel comfortable claiming that Bijectors works with Mooncake on 1.10.0 but then not testing. I think it's just safer to bump the bound on the patch version. It's hard because we can't be more granular in specifying compatibility ranges (e.g. we can't express "Bijectors on its own works with 1.10.0, but with Mooncake it only works on 1.10.8") but IMO this is the least imperfect outcome

sunxd3: I am a bit uneasy about this -- I totally get your point. But being practical, I suspect a lot of people have not updated to the newer Julia versions.

sunxd3: Just to confirm, we can use Mooncake with, say, Julia 10.1 and only hit the compiler error in rear cases?

sunxd3: I am not sure what to do, thoughts @yebai @willtebbutt regarding raising Julia version to 1.10.8?

yebai (approved):
Happy with the changes!

PR #378, opened 2025-03-28
add ForwardDiff@1

penelopeysm:
Hooray, all tests passing (except known Enzyme failures!)

Closes https://github.com/TuringLang/Bijectors.jl/pull/377
Closes https://github.com/TuringLang/Bijectors.jl/pull/376


penelopeysm:
Blocked by DistributionsAD - I've just made a new release so we should be able to rerun this soon

penelopeysm:
It still doesn't resolve to FD=1 since that would downgrade a bunch of other packages it seems. I'm trying to slowly track them down, but it might be a while.

penelopeysm:
ForwardDiff test fails at `test/interface.jl:189`, I bisected to https://github.com/JuliaDiff/ForwardDiff.jl/pull/695, the relevance of which I don't immediately see...

reproduce:

```julia
using Distributions, Bijectors, ForwardDiff, LinearAlgebra, Test

dist = Dirichlet([1000 * one(Float64), eps(Float64)])
b = Bijectors.SimplexBijector()
x = if any(rand(dist) .> 0.9999)
    [0.0, 1.0][sortperm(rand(dist))]
else
    rand(dist)
end
y = b(x)
@test logabsdet(ForwardDiff.jacobian(b, x)[:, 1:(end - 1)])[1] ≈
    logabsdetjac(b, x)

ForwardDiff.jacobian(inverse(b), y)[1:(end - 1), :]

@test logabsdet(ForwardDiff.jacobian(inverse(b), y)[1:(end - 1), :])[1] ≈
    logabsdetjac(inverse(b), y)
```

penelopeysm:
Failing tests in `test/transform.jl` reported https://github.com/JuliaDiff/ForwardDiff.jl/issues/738

sunxd3:
On buildkite: short while ago, I asked the JuliaGPU folks to add this. We don't currently have GPU testing set up. And I think this is why the buildkite is failing. (It doesn't look great to have a failed test though, I'll fix it soon.)

test/interface.jl
```diff
@@ -145,12 +145,10 @@ end
...
     vector_dists = [
         Dirichlet(2, 3),
-        Dirichlet([1000 * one(Float64), eps(Float64)]),
-        Dirichlet([eps(Float64), 1000 * one(Float64)]),
+        Dirichlet([10.0, 0.1]),
+        Dirichlet([0.1, 10.0]),
         MvNormal(randn(10), Diagonal(exp.(randn(10)))),
         MvLogNormal(MvNormal(randn(10), Diagonal(exp.(randn(10))))),
-        Dirichlet([1000 * one(Float64), eps(Float64)]),
-        Dirichlet([eps(Float64), 1000 * one(Float64)]),
```
sunxd3, line 153: Were these lines just simply repeated?

test/transform.jl
```diff
@@ -240,14 +218,14 @@ end
...
-        inds = [
-            LinearIndices(size(x))[I] for I in CartesianIndices(size(x)) if
-            (uplo === :L && I[2] < I[1]) || (uplo === :U && I[2] > I[1])
-        ]
         J = ForwardDiff.jacobian(z -> link(dist, Cholesky(z, x.uplo, x.info)), x.UL)
+        # Remove columns of Jacobian that are all zero (i.e. those
+        # corresponding to entries above the diagonal for uplo = :U, or below
+        # the diagonal for uplo = :L). This slightly unscientific approach
+        # based on filter() is needed to handle both ForwardDiff 0.10 and 1 as
+        # the exact indices will differ for the two versions.
```
sunxd3, line 227: Does it worth it to add a reference for discussion at https://github.com/JuliaDiff/ForwardDiff.jl/issues/738?

sunxd3 (approved):
Looks great!

test/interface.jl
penelopeysm: Indeed, I don't know why they were there twice, I can only suppose they were accidentally copy pasted

test/interface.jl
```diff
@@ -145,12 +145,10 @@ end
 @testset "Multivariate" begin
     vector_dists = [
         Dirichlet(2, 3),
-        Dirichlet([1000 * one(Float64), eps(Float64)]),
-        Dirichlet([eps(Float64), 1000 * one(Float64)]),
+        Dirichlet([10.0, 0.1]),
```
yebai, line 148: I am not sure these new tests could replace the current ones. Tests like 

```julia
Dirichlet([1000 * one(Float64), eps(Float64)]),
        Dirichlet([eps(Float64), 1000 * one(Float64)]),
```

are aimed at the numerical stability of very extrate examples of Dirichlet distributions, i.e. one axis has a very tiny probability mass in average.

penelopeysm: It's actually the sample that's the problem. For the sample `x = [1.0, 0.0]`, the transformed variable is `y = [36.0436]` which is outside of the range for which Float64 is numerically stable.

The issue comes from these lines:

https://github.com/TuringLang/Bijectors.jl/blob/d8d781b5e124c47b1adb23b1480a1f923ef18267/src/bijectors/simplex.jl#L89-L90

As `y[1]` tends to `+Inf`, `z` tends to 1, and the expression `(z - ϵ) / (one(T) - 2ϵ)` tends towards `1.0000000000000002`. If that expression is greater than 1, then it gets `_clamp`ed to 1, and the derivative is set to 0.

The difference between FD 0.10 and FD 1.0 is that the new version sets the derivative to 0 if the expression is greater than, _or equal to_, 1. And that in turn means that there is a larger range of `y[1]` for which the derivative gets clamped. Unfortunately, Float64 36.0436 falls into that category (35.8 would have been fine, or alternatively, BigFloat is ok up until around 175).

As far as I can tell the fact that it used to work with FD 0.10 was a happy accident haha

penelopeysm: I am still not fully sure how to resolve this though, which is why I haven't really come back to this PR. Obviously changing the sample fixes the tests (and the easiest way to change the sample was to change the distribution from which it was drawn), but I can't tell if there's a workaround in the code that makes it work again for `(z - ϵ) / (one(T) - 2ϵ) == 1.0`.

penelopeysm: Pinging @devmotion for your thoughts too :)

yebai (approved):

yebai:
@penelopeysm, please feel free to merge.

test/interface.jl
penelopeysm: > But I'm not sure what exactly would be broken when they would be removed, maybe would be interesting to see.

Finally came back to this; I've been holding off on merging because I haven't had time to properly look into this.

Removing eps actually makes the original tests pass!... But for all the wrong reasons:

https://github.com/TuringLang/Bijectors.jl/blob/fbaf783c1e84540903d62179775936fd5169392a/src/bijectors/simplex.jl#L52-L62

For the sample `x = [1.0, 0.0]`, `z = 1.0` and `LogExpFunctions.logit(z) = Inf`, which means that the bijector itself becomes numerically unstable.

Right now it's just the ForwardDiff jacobian used to verify the manual logabsdetjac that returns `±Inf`; our implementation of `link`, `invlink`, `logpdf_with_trans`, etc. are still stable against extreme values, so this test failure shouldn't have negative effects on upstream packages.

I tried 2*eps and unfortunately that made ForwardDiff even less happy. So I think I'm inclined to merge (and @yebai really wants this PR in).

PR #381, opened 2025-04-14
Replace mentions of stack with Stacked in docs

### LINKED ISSUE #380: Error of Example in Documentation of Stacked
Bijectors 0.15.6 Documentation states the following example:

```
  using Bijectors
  b1 = Logit(0.0, 1.0)
  b2 = identity
  b = stack(b1, b2)
  b([0.0, 1.0]) == [b1(0.0), 1.0]  # => true
```

First, `Logit` is not exported and it needs to be qualified.
```
  b1 = Bijectors.Logit(0.0, 1.0)
```

For me (Julia 11.4 on Linux) also the stack fails:
```
ERROR: MethodError: no method matching length(::typeof(identity))
The function `length` exists, but no method is defined for this combination of argument types.
```

I encountered this issue when trying to create an Identity-Bijector that I can pass as an `Bijectors.Transform` object. What is the best way to create such an object?

mhauru:
Closes #380


penelopeysm (approved):
I don't think #382 and this PR conflict, and since CI passed on this already, I'll just merge without waiting for CI.

PR #405, opened 2025-08-13
Trying to organise the tests, part 1

penelopeysm:
In the Bijectors test suite there are two kinds of AD tests:

1. Check that the manual implementation of `Bijectors.logabsdetjac` returns the same numerical result as an AD backend (I think we should ideally like to always use FiniteDifferences, but in some cases we use ForwardDiff). This usually involves a manual call to `ForwardDiff.jacobian` or similar.

2. Check that AD backends can differentiate through Bijectors code. This uses the `test_ad` function.

Generally these fall under the 'interface' / 'AD' test suite respectively.

But there are a couple of stragglers where the `test_ad` function is being called in the interface tests. This PR gets rid of the exceptions.

This PR also streamlines CI a bit:

- Removes macOS, I don't think it gives us any extra information over Ubuntu. It also makes it quite hard for other packages to get macOS runners for CI.
- Don't test `min` and `lts` separately (our min compat is 1.10.8 and lts is 1.10.10 so it's really no different).

### DIFF
--- a/.github/workflows/AD.yml
+++ b/.github/workflows/AD.yml
@@ -23,7 +23,6 @@ jobs:
           - '1'
         os:
           - ubuntu-latest
-          - macOS-latest
         AD:
           - Enzyme
           - ForwardDiff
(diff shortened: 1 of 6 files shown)


penelopeysm:
actually i decided to just fix even more things in #406 so this PR probably doesn't need to be separately reviewed, sorry

PR #423, opened 2025-11-15
Enable Enzyme tests that are fixed now

penelopeysm:


penelopeysm:
CI failure will be fixed by https://github.com/EnzymeAD/Enzyme.jl/pull/2790. Tested locally with that patch and it works

penelopeysm:
if anyone asks why this PR actually did nothing, it's because i enabled them in a previous PR already and didn't realise it lol. so this is just inflating my github PR statistics

PR #427, opened 2025-11-28
add a walkthrough on how to define bijectors

penelopeysm:
This is just a docs change, no source code changes. It summarises a bunch of things I've learnt working on this, and especially things that might not be obvious to somebody who doesn't have a very strong mathematical background (one general gripe I have with Turing docs is that it's not super beginner friendly, it very much assumes you know what Bayesian inference is and why you want to do it).

Preview @ https://turinglang.org/Bijectors.jl/previews/PR427/defining_examples/

I checked the maths about five times and I am pretty sure I eradicated all sign errors 😅

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+# 0.15.14
+
+Added a new documentation page describing how to implement custom bijectors (this release contains no code changes; it only exists to make sure that the 'stable' docs are updated).
+
 # 0.15.13
 
 Exports extra functionality that should probably have been exported, namely `ordered`, `isinvertible`, and `columnwise`, from Bijectors.jl.
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Bijectors"
 uuid = "76274a88-744f-5084-9051-94815aaf08c4"
-version = "0.15.13"
+version = "0.15.14"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"


docs/src/defining_examples.md
```diff
@@ -0,0 +1,279 @@
...
+This means that the log-absolute determinant of the Jacobian is always `0`.
+
+!!! note
+    
+    In our case, the permutation is always even, since we are performing a cyclic shift. However, since we only need the log-*absolute* determinant, we don't even need to worry about knowing this.
+
+We can now implement `with_logabsdet_jacobian`.
+
+```@example cyclic
+function Bijectors.with_logabsdet_jacobian(
```
devmotion: Bijectors doesn't actually own `with_logabsdet_jacobian`, and additionally it might be nice to show that you don't have to depend on Bijectors if you only want to define such a simple bijector:


```suggestion
function ChangesOfVariables.with_logabsdet_jacobian(
```

docs/src/defining_examples.md
```diff
@@ -0,0 +1,279 @@
...
+logabsdetjac(b, x)
+```
+
+We can also define the inverse bijector.
+A default definition for `inverse(b)` already exists: it would return `Bijectors.Inverse(b)`.
+But, if we used this default definition, we would have to also define `with_logabsdet_jacobian(::Inverse{CircShift}, y)`.
+We can save ourselves this hassle by overloading the method:
+
+```@example cyclic
+inverse(b::CircShift) = CircShift(-b.shift)
```
devmotion: AFAICT you'd have to qualify this method, eg

```suggestion
InverseFunctions.inverse(b::CircShift) = CircShift(-b.shift)
```

docs/src/defining_examples.md
```diff
@@ -0,0 +1,288 @@
...
+0 & 1 & 0
+\end{bmatrix}.$$
+
+In general, the Jacobian of such a transformation is a [permutation matrix](https://en.wikipedia.org/wiki/Permutation_matrix).
+The determinant of a permutation matrix is either `1` or `-1`, depending on whether the permutation is even or odd.
+This means that the log-absolute determinant of the Jacobian is always `0`.
+
+!!! note
+    
+    In our case, the permutation is always even, since we are performing a cyclic shift. However, since we only need the log-*absolute* determinant, we don't even need to worry about knowing this.
```
sunxd3: on
> permutation is always even

is this because we are using a length-3 vector?

penelopeysm: Yes, indeed, I think when I wrote this I firmly had length-3 in mind, but it will depend! Made the phrasing more general now. Thanks

PR #429, opened 2025-12-25
Vector bijectors

### LINKED ISSUE #237: Bijector for MatrixNormal
Hi! I'm trying to sample the following Turing Model that uses a `MatrixNormal` distribution:

```{julia}
using Random, Turing, Bijectors
Random.seed!(123)

#Estimate a MatrixNormal as simulated here
U  = rand(LKJ(2, 0.5))
V  = rand(LKJ(2, 0.5))
Uₐ = rand(LKJ(2, 0.5))
Vₐ = rand(LKJ(2, 0.5))
Asample = rand(MatrixNormal(zeros(Float64, 2, 2), U, V))

#Create the model
@model function estimateA(A, U, V, Uₐ, Vₐ)
    mu ~ MatrixNormal(zeros(Float64,size(A,1), size(A,2)), Uₐ, Vₐ)
    A  ~ MatrixNormal(mu, U, V)
end

#Estimate!
model  = estimateA(Asample, U, V, Uₐ, Vₐ);
chains = sample(model, NUTS(), 100);
```

however I get the following error: 

```
ERROR: MethodError: no method matching bijector(::MatrixNormal{Float64, Matrix{Float64}, PDMats.PDMat{Float64, Matrix{Float64}}, PDMats.PDMat{Float64, Matrix{Float64}}})
Closest candidates are:
  bijector(::Union{Kolmogorov, BetaPrime, Chi, Chisq, Erlang, Exponential, FDist, Frechet, Gamma, InverseGamma, InverseGaussian, LogNormal, NoncentralChisq, NoncentralF, Rayleigh, Weibull}) at ~/.julia/packages/Bijectors/vUc4m/src/transformed_distribution.jl:58
  bijector(::Union{Arcsine, Beta, Biweight, Cosine, Epanechnikov, NoncentralBeta}) at ~/.julia/packages/Bijectors/vUc4m/src/transformed_distribution.jl:69
  bijector(::Union{Levy, Pareto}) at ~/.julia/packages/Bijectors/vUc4m/src/transformed_distribution.jl:72
```

The same error happens with the following code: 
```{julia}
dist = MatrixNormal(zeros(2,2), rand(LKJ(2, 0.5)), rand(LKJ(2, 0.5)))
b     = bijector(dist)
```

I'm relatively new to `Turing` so maybe my diagnosis is not correct but it seems to me that the Bijectors package is lacking a definition for the `MatrixNormal`.

### LINKED ISSUE #258: Adding bijectors for OrderStatistic and JointOrderStatistics
https://github.com/JuliaStats/Distributions.jl/pull/1668 added distributions for [sample order statistics](https://juliastats.org/Distributions.jl/stable/order_statistics/)" `OrderStatistic` and `JointOrderStatistics`.

`OrderStatistic` has the same constraints as the distribution it wraps and should get the same bijector.

```julia
bijector(d::OrderStatistic) = bijector(d.dist)
```

The support of `JointOrderStatistics` is the sorted vectors whose elements are in the support of the wrapped distribution. An example bijector could be:

```julia
function bijector(d::JointOrderStatistics)
    dist = d.dist
    binner = bijector(dist)
    bunsort = inverse(OrderedBijector())
    # assume all upper bounded distributions have monotonically decreasing bijectors.
    scale = (binner isa Union{typeof(identity),Truncated} || islowerbounded(dist)) ? 1 : -1
    return bunsort ∘ Scale(scale) ∘ binner
end
```

To set `scale` correctly, we need to know if the bijector being used is monotonically increasing or decreasing. While we can hardcode a check for all bijectors defined in this package, to support user-implemented bijectors, we would seem to need an `isdecreasing(::Bijector)` API function.

Note that with a correct bijector defined for `JointOrderStatistics`, in a PPL like Turing, a user could use `JointOrderStatistics(Exponential(), 100)` to infer a sorted vector of parameters from the same distribution, instead of `Bijectors.ordered(filldist(Exponential(), 100))`, which as noted in https://github.com/TuringLang/Bijectors.jl/issues/220#issuecomment-1409115410 does the wrong thing (and now raises an error).

### LINKED ISSUE #398: inconsistent invlink types with 0-dim arrays / floats
Ran into this while working on an upstream PR: https://github.com/TuringLang/DynamicPPL.jl/pull/967

```julia
julia> using Bijectors: bijector, inverse, output_size

julia> for dist in (LogNormal(), Normal())
           b = bijector(dist)
           sz = output_size(b, size(dist))
           y = randn(sz)
           b_inv = inverse(b)
           x = b_inv(y)
           @show typeof(x)
       end
typeof(x) = Float64
typeof(x) = Array{Float64, 0}
```

The workaround for this is straightforward, but I feel like it shouldn't be DynamicPPL's responsibility to work around it?

I guess are a couple of possible ways out, though none strike me as obviously correct:

1. Make output_size return `(1,)` if `size(dist) == ()`. Would make life easier on DynamicPPL side as I wouldn't have to deal with any 0-dim arrays. However, I don't think this is technically correct.
2. Keep the behaviour of `output_size` but change the behaviour for LogNormal to match that for Normal, i.e. both cases should return 0-dim arrays. This would be more principled, but I'm frankly unsure if this is possible because Julia is weird with 0-dim arrays anyway.

```julia
julia> x = randn(())
0-dimensional Array{Float64, 0}:
1.427397389133524

julia> exp.(x)  # this is the source of the problem isn't it
4.167837804112459
```

penelopeysm:
This copies the interface over from https://github.com/penelopeysm/VectorBijectors.jl, as discussed in previous meetings, and extends it to support a wider range of distributions than Bijectors itself does.

See https://turinglang.org/Bijectors.jl/previews/PR429/vector/.

The aim of this is to present a coherent and performant API that DynamicPPL and Turing can rely on. The current Bijectors interface is scattered and unnecessarily wasteful sometimes (see e.g. https://github.com/TuringLang/DynamicPPL.jl/issues/1142).

Closes #237
Closes #258
Closes #398

It also adds a megaton of tests, including full tests with all the AD backends we care about, including both Enzyme/Mooncake forward and reverse modes. The VectorBijectors test suite takes quite a long time to run (45–60 mins on CI), so I've made it its own test group. All the old tests are under the 'Classic' test group.

## Progress

- [x] Univariates
  - [x] LocationScale
  - [x] Truncated
  - [x] Censored
- [x] Multivariates
- [x] ReshapedDistribution
- [x] Matrix
   - [x] MatrixNormal, MatrixTDist
   - [x] InverseWishart,Wishart generate positive (semi)definite matrices, and there is a PDBijector.
   - [x] LKJ uses VecCorrBijector (although this one is not perfect, see recent issues)
   - Note that there are quite a few other matrix distributions (https://juliastats.org/Distributions.jl/stable/matrix/#Distributions), but we could conceivably ignore them for now — Bijectors' original implementation does not cover them, so skipping them for now would not be a regression.
- [x] LKJCholesky
- [x] Product distributions
  - [x] tuples
  - [x] arrays
  - [x] namedtuples
- [x] Mixture models. Note that we can probably only handle univariate mixtures sensibly and I think those should already subtype UnivariateDistribution so should end up using the truncated bijector --- if my instincts are correct, then that should already work out of the box and so the only thing we need to do is to add a test case.
- [x] Order statistics and joint order statistics ~~(should be trivial to implement by forwarding to the base distribution)~~ It was quite tricky...

### DIFF
--- a/src/vector/product/fill.jl
+++ b/src/vector/product/fill.jl
@@ -64,7 +64,7 @@ function (t::ProductVecTransform{<:Elementwise{F,Dims{M}},Nothing,Dims{N}})(
         vec(trf.(x))
     else
         dims = ntuple(i -> i + N, Val(M))
-        vec(stack(map(trf, eachslice(x; dims=dims))))
+        vec(stack(trf, eachslice(x; dims=dims)))
     end
 end
 
@@ -98,7 +98,7 @@ function with_logabsdet_jacobian(
     else
         dims = ntuple(i -> i + N, Val(M))
         lj = WithLogabsdetjac(trf, _fzero(T))
-        y = vec(stack(map(lj, eachslice(x; dims=dims))))
+        y = vec(stack(lj, eachslice(x; dims=dims)))
         y, lj.logjac
     end
 end
@@ -114,7 +114,7 @@ function (t::ProductVecInvTransform{<:Elementwise{F,Dims{M}},Nothing,Dims{N}})(
         # dim 1 is the input for the inverse transform
         reshaped_y = reshape(y, :, t.transforms.size...)
         dims = ntuple(i -> i + 1, Val(M))
-        stack(map(t.transforms.value, eachslice(reshaped_y; dims=dims)))
+        stack(t.transforms.value, eachslice(reshaped_y; dims=dims))
     end
 end
 
@@ -136,7 +136,7 @@ function with_logabsdet_jacobian(
         reshaped_y = reshape(y, :, t.transforms.size...)
         dims = ntuple(i -> i + 1, Val(M))
         lj = WithLogabsdetjac(trf, _fzero(T))
-        x = stack(map(lj, eachslice(reshaped_y; dims=dims)))
+        x = stack(lj, eachslice(reshaped_y; dims=dims))
         x, lj.logjac
     end
 end


penelopeysm:
@sunxd3 I'm still working on adding the actual distributions themselves, but could I get you to look at the tests and tell me if they seem sensible / if there's anything else I can add to them? So basically just the `src/vector/test_utils.jl` file, and also `test/` (the interesting bit is `test_utils`, the tests are very boring, they just iterate over a ton of different distributions).

There's docs for the API at https://turinglang.org/Bijectors.jl/previews/PR429/vector/, if any of that is unclear do point out too. I suspect I might need to explain `linked_optic_vec` better.

Personally the test suite makes me feel quite confident that it's impossible to mess up the implementation of the bijectors, so I think the test suite is probably the most important part to get a review on.

src/vector/univariate/positive.jl
```diff
@@ -0,0 +1,83 @@
+# Bijectors for continuous univariate distributions which have support over the positive (or
+# non-negative) real numbers.
+
+"""
+    Exp(bound, sign) <: ScalarToScalarBijector
+
+Callable struct, defined such that `(e::Exp)(y) = ((e.sign * exp(y)) + e.bound)`. The sign
+is determined by the `negative` field.
```
sunxd3:
```suggestion
is determined by the `sign` field.
```

src/vector/univariate/positive.jl
```diff
@@ -0,0 +1,83 @@
...
+    D.Rayleigh,
+    D.Rician,
+    D.StudentizedRange,
+    D.Weibull,
+}
+VectorBijectors.from_linked_vec(d::POSITIVE_UNIVARIATES) = OnlyWrap(Exp(minimum(d), 1))
+VectorBijectors.to_linked_vec(d::POSITIVE_UNIVARIATES) = VectWrap(Log(minimum(d), 1))
+
+function VectorBijectors.from_linked_vec(
+    d::D.LocationScale{<:Any,<:Any,<:POSITIVE_UNIVARIATES}
```
sunxd3: should we use `AffineDistribution` consistently?

src/vector/product/product.jl
```diff
@@ -0,0 +1,456 @@
...
+end
+
+@generated function with_logabsdet_jacobian(
+    t::ProductVecTransform{<:NTuple{P,Any},<:NTuple{P,Any},<:NTuple{N,Int}},
+    x::AbstractArray{T},
+) where {P,N,T}
+    # P = number of distributions in the product distribution
+    # N = dimension of each distribution
+    exprs = []
+    push!(exprs, :(total_length = sum(length, t.ranges)))
```
sunxd3: do you think we should check dimension match for these transformations, right now, I think it will only error when `x` is shorter than expected but not longer than expected. I don't think this is critical, but being more defensive could be worthwhile?

src/vector/univariate/positive.jl
penelopeysm: Not sure, hmm... Distributions.jl defines `const LocationScale{T,S,D} = AffineDistribution{T,S,D}` so probably better to use the base type rather than the alias. I'll change it

penelopeysm: Done

PR #443, opened 2026-03-03
optimisations for vector bijectors

penelopeysm:
Mostly stealing an idea from the original Bijectors implementation, this PR optimises for the case when all the elements of a product distribution have the same bijector (technically, only when it is statically inferrable that they have the same bijector -- otherwise we would lose type stability).

Pushing this somewhere as I want to be able to depend on this in other PRs, but not merging into #429 yet since that's under review.

Some benchmarks, where "previously" = on #429:

```julia
using Bijectors, Chairmarks, Test

# base = Beta(2, 2)
base = Dirichlet(ones(2))

d = product_distribution(fill(base, 10))
x = rand(d)
tlv = @inferred Bijectors.VectorBijectors.to_linked_vec(d)
flv = @inferred Bijectors.VectorBijectors.from_linked_vec(d)
y = tlv(x) 
@test flv(y) ≈ x

# Beta
@b tlv(x)                          # 75 ns, 2 allocs  (previously 278 ns, 22 allocs)
@b with_logabsdet_jacobian(tlv, x) # 149 ns, 3 allocs (previously 342 ns, 23 allocs)
@b flv(y)                          # 59 ns, 2 allocs  (previously 731 ns, 32 allocs)
@b with_logabsdet_jacobian(flv, y) # 270 ns, 3 allocs (previously 903 ns, 33 allocs)

# Dirichlet
@b tlv(x)                          # 304 ns, 25 allocs   (previously 312 ns, 22 allocs)
@b with_logabsdet_jacobian(tlv, x) # 420 ns, 27 allocs   (previously 435 ns, 23 allocs)
@b flv(y)                          # 424 ns, 25 allocs   (previously 2.30 µs, 52 allocs)
@b with_logabsdet_jacobian(flv, y) # 752 ns, 27 allocs (previously 2.51 µs, 53 allocs)
```

### DIFF
--- a/src/vector/product/fill.jl
+++ b/src/vector/product/fill.jl
@@ -74,7 +74,7 @@ mutable struct WithLogabsdetjac{T,R}
     trf::T
     logjac::R
 end
-function (w::WithLogabsdetjac{T,R})(x::AbstractArray{N}) where {T,R,N}
+function (w::WithLogabsdetjac)(x::AbstractArray)
     y, lj = with_logabsdet_jacobian(w.trf, x)
     # mutate the logjac in place; we'll pick it up later
     w.logjac += lj


penelopeysm:
And wow, there's a Mooncake compilation error! A rarity indeed.

PR #453, opened 2026-03-11
Add custom ordering to OrderedBijector

a1ix2:
I work with pharmacokinetic models where you often have local but not global identifiability and you need to impose some ordering constraint on subsets of elements of a multivariate rv but not other A strictly increasing ordering allows you to partially deal with that but at the cost of introducing a lot of reordering gymnastic within the model specification itself.

This PR introduces a type parameter so that now OrderedBijector{OT<:AbstractOrdering} that allows to impose any fix ordering, including an incomplete one, e.g. for some rv X ~ MvNormal(I(D)), you could have x[3] < x[4] < x[1] while all other indices are left unconstrained. The type parameter OT can Ascending, Descending, or FixedOrder{ordertuple}. In the previous example this would lead to OrderedBijector{FixedOrdering{(3,4,1)}} and OrderedDistribution{D, B, FixedOrdering{(3,4,1)}}. The default is Ascending. I've also modified chainrules.jl to account for these changes. 

It passes all tests, but looking for feedback to bring everything up to the project's standards and style and for someone to double-check the math.


src/bijectors/ordered.jl
```diff
@@ -15,13 +26,22 @@ A bijector mapping unordered vectors in ℝᵈ to ordered vectors in ℝᵈ.
...
   - Note that this transformation and its inverse are the _opposite_ of in this reference.
 """
-struct OrderedBijector <: Bijector end
+struct OrderedBijector{OT<:AbstractOrdering} <: Bijector
+    OrderedBijector() = new{Ascending}()
+    OrderedBijector(::OT) where {OT <: AbstractOrdering}= new{OT}()
+    OrderedBijector{OT}() where {OT <: AbstractOrdering} = new{OT}()
+    OrderedBijector(ordertuple::Tuple{Int, Int, Vararg{Int}}) = OrderedBijector{FixedOrder{ordertuple}}()
+    function OrderedBijector{FixedOrder{ordertuple}}() where {ordertuple}
+        @assert (ordertuple isa Tuple{Int, Int, Vararg{Int}}) && all(ordertuple .> 0) && allunique(ordertuple)
```
penelopeysm: Given that these checks are already inside the inner constructor of FixedOrder, they don't need to be repeated here.

src/bijectors/ordered.jl
```diff
@@ -48,20 +68,79 @@ function _transform_ordered(y::AbstractMatrix)
...
 
-transform(ib::Inverse{OrderedBijector}, x::AbstractVecOrMat) = _transform_inverse_ordered(x)
-function _transform_inverse_ordered(x::AbstractVector)
+function _transform_ordered(y::AbstractVector, ::Type{Descending})
+    x = similar(y)
+    @assert !isempty(y)
+
+    N = length(x)
+    @inbounds x[N] = y[N]
+    @inbounds for i in N-1:-1:1
```
penelopeysm: I learnt fairly recently that peppering `@inbounds` everywhere is not really helpful: https://github.com/JuliaStats/Distributions.jl/pull/2005

So far I don't think I've ran into a case where they actually helped, so I think it would probably be better to remove them, unless there's a demonstrable case where it improves performance.

src/bijectors/ordered.jl
penelopeysm: The docstring of `ordered` needs to be updated to describe the improvements in this PR.

src/bijectors/ordered.jl
```diff
@@ -6,6 +6,17 @@ output_size(::SignFlip, dim) = dim
 is_monotonically_increasing(::SignFlip) = false
 is_monotonically_decreasing(::SignFlip) = true
 
+abstract type AbstractOrdering end
+struct Ascending <: AbstractOrdering end
+struct Descending <: AbstractOrdering end
+struct FixedOrder{ordertuple} <: AbstractOrdering
```
penelopeysm: Does the tuple itself need to be a type parameter? Looking at the code below, I'm not sure if any of it really needs the tuple to be known at the type level for type stability, because the ordered stuff just turn a vector (or matrix) into a vector (or matrix). If we don't need it in the type, then it would probably be better to keep it as an ordinary field:

```julia
struct FixedOrder{T} <: AbstractOrdering
    ordertuple::T
end
```

as otherwise we might end up creating multiple method specialisations for different values of `ordertuple` even when it's not needed.

penelopeysm (changes requested):
Hi @a1ix2, thanks for the PR, and sorry that nobody responded until now! Please feel free to ping me if there are PRs that need to be reviewed.

I do like the direction of this PR and it is implemented cleanly. There are a handful of things I'd change, which are in the comments below.

One obvious missing piece is that the new functionality is not tested. The AD rules are tested, but I think it would be valuable to add tests to `test/bijector/ordered.jl` which exercise the functions you've implemented, and check that the invariants you are expecting hold true. For example, if you sample from `ordered(MvNormal(...), Descending())` are the resulting samples reverse-sorted, and so on. I don't think you need to go to the extreme of doing the MCMC sampling in there, but some common-sense tests would go a long way.

PR #469, opened 2026-05-13
Move all Enzyme tests into a separate integration test environment

### LINKED ISSUE #347: Move all Enzyme related tests into a seprate CI
Let's move all Enzyme related tests in 

- [Interface test](https://github.com/TuringLang/Bijectors.jl/actions/runs/11704820479/job/32598116729)
- [ Enzyme test ](https://github.com/TuringLang/Bijectors.jl/actions/runs/11704820490/job/32598119219)

into a separate test script and run that on a dedicated CI task (allowed to fail for PRs )

yebai:
Enzyme tests are currently broken on main and take the rest of the suite down with them. This PR moves all Enzyme coverage into test/integration/enzyme/ (its own Project.toml, run by the existing Enzyme.yml workflow) so Enzyme failures stay scoped to that job.

Fix https://github.com/TuringLang/Bijectors.jl/issues/347

Testing against Enzyme is a waste of TuringLang researchers' time until Enzyme is sufficiently robust. The same applies to Turing, Bijectors, DynamicPPL and AdvancedVI. These tests were added based on past goodwill but have limited value because they are constantly broken for entirely normal Julia syntax and library usage.


wsmoses:
The full bijectors.jl suite has remained passing on Enzyme CI. If there is an error it is your responsibility to open issues, given everything we run passes

test/integration/enzyme/main.jl
```diff
@@ -1,105 +1,226 @@
...
-    @test isapprox(gradient, ref_gradient; rtol=rtol, atol=atol)
-end
-
-BACKENDS = [
+include(joinpath(@__DIR__, "..", "..", "test_resources.jl"))
+
+# Enzyme adtype flavours. The flags differ across tests because closure shapes and
+# Enzyme's activity inference do not permit a single configuration:
+#
+# * `function_annotation=Const` — the closure itself captures a bijector that is
```
wsmoses: This comment is wrong.

@yebai per your comment on other turing repos on whether someone is allowed to review, am I allowed to review this?

wsmoses (changes requested):

test/integration/enzyme/main.jl
yebai: I revised the comment a bit; it is quite unfortunate that we need different enzymes for different tests. 

Suggestions on whether a universal Enzyme config to pass all tests are welcome.

PR #460, opened 2026-04-23
Use AbstractPPL AD interface in tests.

yebai:
Replacing https://github.com/TuringLang/Bijectors.jl/pull/458

TODO 

- [x] uncomment `min` / LTS tests. 
- [x] remove AbstractPPL pinned sources.

PR #470, opened 2026-05-14
Move AD backends to standalone integration test suites

yebai:
No changes to `src`, only tests restructuring. Supersede https://github.com/TuringLang/Bijectors.jl/pull/469

- Per-backend integration suites. ReverseDiff, Mooncake, and Enzyme now live under test/integration_tests/{reversediff,mooncake,enzyme}/, each with its own Project.toml, main.jl, and CI         
workflow. ReverseDiff and Mooncake block PRs; Enzyme reports failures for visibility but is continue-on-error. The main suite covers ForwardDiff inline (it's the AD reference for both
run_ad_case and the vector AD loops).                                                                                                                                                             
- Drop DifferentiationInterface from Bijectors deps. DI is now test-only. The DI-dependent helpers (linked_optic_vec AD check, AD-vs-analytical logjac, test_ad) move from                      
src/vector/test_utils.jl into shared test resources; VectorBijectors.test_all keeps only the structural (DI-free) checks.                                                                         
- Drop DistributionsAD from test deps. test/distributionsad.jl and the TuringWishart/filldist blocks that depended on it are removed; the package's [weakdeps] entry is untouched.
- Unified test-case machinery. New test/test_resources.jl provides ADTestCase/VectorTestCase with per-source tag::Symbols and the aggregate entry points generate_ad_testcases() /                
generate_vector_testcases(). Each integration main.jl reduces to one adtype list + one is_broken(c) predicate + two short loops. Enzyme collapses to a single runtime_activity + Const adtype     
across all cases; product-bijector Enzyme failures are skipped by a structural predicate, not by index.                                                                                           
- Test reorganisation. test/ad/* folds into the integration mains (rule-correctness moves to test/bijectors/chainrules.jl). test/interface.jl is split into per-feature files under               
test/bijectors/ (stacked.jl, simplex.jl, equality.jl, scale.jl). Renames: test/integration → test/integration_tests, test/norm_flows.jl → test/normalising_flows.jl, test/transform.jl →          
test/legacy_interface.jl. JuliaFormatter v1.0 applied across the board.


yebai:
@sunxd3, a quick look should be enough -- there is no functionality change or loss of tests here (verified), it is only restructuring tests.

wsmoses:
why are you considering Enzyme differently here?

sunxd3:
Two points worth noting:
1. https://github.com/TuringLang/Bijectors.jl/blob/bc58fcddd8cc98f54d98c436efc74068de98d6e3/docs/src/vector.md?plain=1#L110-L122 needs some update
2. should we even keep DistributionsAD extension if we remove the tests?

yebai:
Thanks @sunxd3, both good catches.

(2) DistributionsAD extension — dropped in 24ed323. Verified it's safe:

DistributionsAD.jl is archived (last release March 2025); Turing v0.43 explicitly deprecated it.
None of the current TuringLang packages (Turing/DynamicPPL/AdvancedVI/AdvancedHMC/AdvancedMH) depend on it, and arraydist/filldist no longer return the Turing*/FillVector* types the extension dispatches on.

~~(1) docs/src/vector.md—leaving for a follow-up PR since it's a doc cleanup independent of the test restructuring here.~~

EDIT: I removed the changes to `src/vector/test_utils.jl`, now only changing the default adtype to ForwardDiff. ReverseDiff, Enzyme, and Mooncake are still tested in their own integration test environments.

sunxd3:
thanks Hong.

couple of follow-ups:
1. it's worth updating https://github.com/TuringLang/Bijectors.jl/blob/bc58fcddd8cc98f54d98c436efc74068de98d6e3/docs/src/vector.md?plain=1#L120 now that only ForwardDiff is the only default in this PR
2. lkj's test on AD is narrow on main https://github.com/TuringLang/Bijectors.jl/blob/ad208818fd2ae0cbd8a32eccd905e5e23c0ab522/test/vector/matrix.jl#L19-L22 I think this PR might have added tests of lkj on Enzyme and Forwardiff

yebai:
@sunxd3 all comments addressed - this should be ready for a final look. See the commit message for what has changed.

src/vector/test_utils.jl
```diff
@@ -652,3 +637,48 @@ function test_ad(d::D.Distribution, adtypes::Vector{<:DI.AbstractADType}, atol,
...
+                    ) &&
+                    isapprox(
+                        DI.gradient(ladj_rev, adtype, yvec),
+                        ref_grad_ladj_rev;
+                        atol=atol,
+                        rtol=rtol,
+                    )
+                )
+            end
+        end
```
sunxd3, line 682: maybe split these to give more visibility to failure

Project.toml
```diff
@@ -47,9 +45,8 @@ ArgCheck = "1, 2"
 ChainRules = "1"
 ChainRulesCore = "0.10.11, 1"
 ChangesOfVariables = "0.1"
-DifferentiationInterface = "0.7.14"
+DifferentiationInterface = "0.7.7"
```
sunxd3: is there a reason for lowering this bound?

sunxd3 (commented):
Couple of items worth attention.

An inconsistency between PR description and code diff is DI dependency (DI is still in package Project.toml, line 10)

src/vector/test_utils.jl
yebai: It's not easy to do that at the moment due to VectorBijectors's current testing setup -- we prefer to minimise changes to the `src` folder in this PR.

Project.toml
yebai: DI will be removed in #460 -- this PR cleans up some tests which I noticed while working on #460.

yebai:
@sunxd3 all addressed.

PR #480, opened 2026-06-18
Support Mooncake 0.6 forward mode in find_alpha rules

yebai:
Mooncake 0.6 reworks batched forward AD around `Lifted`/`NDual` and routes `prepare_derivative_cache` (used by `AutoMooncakeForward`) through `build_frule`. Add `frule!!(::Lifted, ...)` rules for `find_alpha` (float and integer third argument), gated behind `pkgversion(Mooncake) >= v"0.6"`, keeping the existing `Mooncake.Dual` rules for 0.5.

Under 0.6, these `Lifted` rules intercept `find_alpha` as a primitive, so the batched pass never enters its body. Under 0.5 the batched (Nfwd) pass instead propagates `NDual`s straight through `find_alpha`, tripping on `ceil(::Int, ::NDual)` inside the root-finder; the `find_alpha(::NDual, ...)` short-circuit methods fix that and are therefore gated to `< v"0.6"` (verified redundant on 0.6, required on 0.5).

This unbreaks the `AutoMooncakeForward` PlanarLayer-inverse integration case, so drop its `@test_broken` marker. Bump Mooncake compat to allow 0.6.

Note: Mooncake 0.6 is not yet released. This PR fix Mooncake 0.6 compatibility in a non-breaking way.


ext/BijectorsMooncakeExt.jl
```diff
@@ -16,20 +22,61 @@ function _find_alpha_partials(wt_y::P, wt_u_hat::P, b::P) where {P<:Base.IEEEFlo
...
+        y::Lifted{P,N,Mooncake.NDual{P,N}},
+        z::Lifted{P,N,Mooncake.NDual{P,N}},
+    ) where {N,P<:Base.IEEEFloat}
+        α, ∂y, ∂u, ∂b = _find_alpha_partials(primal(x), primal(y), primal(z))
+        dα = ntuple(Val(N)) do lane
+            ∂y * Mooncake.tangent(x).partials[lane] +
+            ∂u * Mooncake.tangent(y).partials[lane] +
+            ∂b * Mooncake.tangent(z).partials[lane]
+        end
+        return Lifted{P,N}(α, Mooncake.NDual{P,N}(α, dα))
```
shravanngoswamii, line 65: In the 0.6-direction Mooncake branches (`step-8-ndual-routing`, `expt/primal-mode`), `Lifted`
is a plain Union alias, not a parametric struct:

```julia
const Lifted = Union{_LiftedBase, MemoryRef{<:NDual}, MemoryRef{<:Complex{<:NDual}}}
```

Project.toml
```diff
@@ -59,7 +59,7 @@ IrrationalConstants = "0.1, 0.2"
 LazyArrays = "2"
 LogExpFunctions = "0.3.3"
 MappedArrays = "0.2.2, 0.3, 0.4"
-Mooncake = "0.4.95, 0.5"
+Mooncake = "0.4.95, 0.5, 0.6"
```
shravanngoswamii: I guess 0.4 will break now, so lets make 0.5 as minimum supported version.

shravanngoswamii:
> Note: Mooncake 0.6 is not yet released. This PR fix Mooncake 0.6 compatibility in a non-breaking way.

Should we wait here until Mooncake 0.6 release?

ext/BijectorsMooncakeExt.jl
yebai: This fix refers to a new version in the branch https://github.com/chalk-lab/Mooncake.jl/pull/1215. 

The `step-8-ndual-routing`, `expt/primal-mode` branches have been superseded.

Project.toml
yebai: Nice catch!

yebai:
> Should we wait here until Mooncake 0.6 release?

Probably not, as this branch is fully compitable with 0.5, and the upcoming 0.6 is unlikely to change `Lifted`.

PR #482, opened 2026-07-08
Use A42 instead of ITP in find_alpha

yebai:
Roots 3.0.2 made `find_zero` return `NaN` (→ `ConvergenceFailed`) when a bracketing method stops in a `:not_converged` state instead of returning the best endpoint. `ITP` hits this by stalling short of the unreachable `eps^3` x-tolerance, breaking `find_alpha`.

`A42` is superlinear like ITP but narrows to adjacent floats and converges via the exact-x path, working on both old and new Roots. The analytic AD rules remain unchanged (they depend only on the returned value being a root).


yebai:
@devmotion, can you help review this PR?

test/interface.jl
```diff
@@ -6,6 +6,11 @@ using ForwardDiff
 using Bijectors
 Random.seed!(123)
 
+# Distributions 0.25.129's bounded kernel-density samplers (Biweight, Cosine, …) can
```
yebai, line 9: This is a separate, unrelated test-fragility fix.

src/vector/test_utils.jl
```diff
@@ -26,13 +26,20 @@ function _name(d::D.JointOrderStatistics)
     return "joint order statistic $(_name(d.dist)) with length $(length(d))"
 end
 
+# Some Distributions kernel-density samplers (Biweight, Cosine, …) return NaN in recent
```
yebai: Also, a separate, unrelated test-fragility fix.

yebai:
I'm merging this as it (1) is fairly low risk, (2) it currently breaks other TuringLang tests.

devmotion:
Probably can be reverted, this was fixed upstream in Roots.

Issue #370, opened 2025-03-11
Batched support for `NormalizingFlows.jl` and `Bijectors.jl`


yebai:
Introduce a `batched mode` to `Bijectors.jl` and `NormalizingFlows.jl`, which are built on top of `Bijectors.jl`.

Put simply, we want to enable users to provide multiple inputs to the model simultaneously by “stacking” the parameters into a higher-dimensional array.

The implementation can take various forms, as a team of developers who care about both performance and user experience, we are open to different approaches and discussions. One possible approach is to develop a mechanism that signals the code to process the given input as a batch rather than as individual entries. 

A preliminary implementation can be found [here](https://github.com/torfjelde/Batching.jl).

(copy and pasted from [here](https://github.com/JuliaLang/www.julialang.org/blob/1811a778c292da509365860d662f4c053a6ecfc8/jsoc/gsoc/turing.md#batched-support-for-normalizingflowsjl-and-bijectorsjl))

yebai:
`AcceleratedKernels.map!` could be an interesting alternative to batching.

PR #485, opened 2026-07-18
Batched, GPU-capable rational quadratic spline

shravanngoswamii:
Adds a batched, GPU-capable rational quadratic spline that works with any AD backend. Goal is for NormalizingFlows nsf to use it instead of MonotonicSplines (TuringLang/NormalizingFlows.jl#63).

Built as plain array operations (reduction bin-search, differentiable gather, boolean-mask branch selection) so it runs on Array and CuArray from one code path and every backend differentiates it directly. No hand-written derivative rules; the ReverseDiff extension only strips tracking from the integer bin location and the root-sign mask.

An independent adversarial review checked every formula against the NSF paper (Durkan et al. 2019) and against nflows and distrax numerically (agreement at Float64 rounding level), and its findings are addressed here:

- minimum bin fraction and minimum derivative (1e-3, as in nflows and distrax), so extreme logits cannot underflow a bin or slope to zero and produce NaN
- boundary knots pinned exactly to -B and B
- the inverse root uses the cancellation-free form for either sign of b, with the sqrt argument kept away from zero so its gradient stays finite
- gradients with respect to the inputs only (a fixed trained flow) now work under ReverseDiff
- tests pin the raw parameter layout, probe extreme logits, and run out-of-range inputs through every reverse-mode suite

Verified locally: 1152 value tests over Float32/Float64, ForwardDiff, Zygote, ReverseDiff, Mooncake, and Enzyme all agreeing with the ForwardDiff reference, including out-of-range inputs. GPU suite runs on the buildkite cuda queue.

Open design questions for review: whether zero raw parameters should give the identity map (distrax offsets the softplus for this, nflows makes it optional, this PR matches the paper and the nflows default), and whether the top-knot pinning plus floors warrant a mention in the docstring.


shravanngoswamii:
@yebai @sunxd3 first sprint was implemented with Claude. I read the implementation against https://arxiv.org/pdf/1906.04032 -- forward map, log-Jacobian, and the quadratic solve in the inverse all match section 3 and appendix A. I also went through the rational quadratic spline code in nflows and distrax and matched their numerical safeguards: a minimum bin width and minimum derivative of `1e-3`, boundary knots pinned exactly to `-B` and `B`, and the stable form of the quadratic root for either sign of b. With matching settings the outputs agree with both of them to about `1e-14` in `Float64`, including inputs in the identity tails.

Happy to walk through any part of it.

yebai (approved):
Good work @shravanngoswamii!

sunxd3:
The logic looks good for what it does and this is a great effort!

I have some concerns about the interface. Bijectors.jl doesn't have a clear batch semantics (appearantly https://github.com/TuringLang/Bijectors.jl/pull/214/ removed some batching support existed before that, there are also discussion at https://github.com/TuringLang/Bijectors.jl/issues/370).

A potential issue with this PR is that it introduced a pretty strong batching semantic (by this I mean which dimension of input correspond to batch dimension and feature). The dimension of logjac looks good otherwise.

Given this, do you think it makes more sense to move the implementation to https://github.com/TuringLang/NormalizingFlows.jl/pull/63? (I know we have talked about this before, sorry for flipping on it.)

And if people think it's worth it to have it in Bijectors.jl, then we probably should have a little bit more doc on the design choices.

shravanngoswamii:
Agree, Batching in Bijectors needs more thought, I wasn't aware of the background work on Bijectors. NormalizingFlows already uses that (D, N) convention in its coupling layers, and NeuralSplineCoupling only calls the array functions (`rqs_params_from_raw`, `rqs_forward`, `rqs_inverse`), never the `BatchedRQS` type, so the move is fine and the type introduced here can go.

We will need a small ReverseDiff extension for the bin-search overload, I will do that and close this PR.

shravanngoswamii:
Agreed. Batching deserves a package-wide design in Bijectors rather than one bijector settling it ahead of #370: a matrix input is ambiguous between a batch of vectors and a single matrix-variate value, and this PR quietly picks the first reading.

NormalizingFlows already uses that (D, N) convention in its coupling layers, and NeuralSplineCoupling only calls the array functions, never the BatchedRQS type, so the type is gone and the rest moved as is.

Moved to TuringLang/NormalizingFlows.jl#63, closing this.

shravanngoswamii:
Moved to TuringLang/NormalizingFlows.jl#63, closing this.

